PR #103, opened 2022-10-03
Fix aria issues and allow inputs to be full width.

micheal-parks:
This PR also:
* Updates dependencies
* Fixed some minor display issues with elements

More information in comments.


.eslintrc.js
```diff
@@ -128,6 +128,7 @@ module.exports = {
     ],
 
     // Unicorn rules
+    'unicorn/switch-case-braces': 'off',
```
micheal-parks, line 131: New rule. I don't like it. If anyone else does LMK.

src/elements/badge.svelte
```diff
@@ -15,7 +15,7 @@ addStyles();
 
 </script>
 
-<small class={cx('block rounded-full px-3 py-0.5 text-xs', {
+<small class={cx('inline-block rounded-full px-3 py-0.5 text-xs', {
```
micheal-parks, line 18: Badge should be inline block since its width shouldn't expand past the text.

src/elements/breadcrumbs.svelte
```diff
@@ -1,4 +1,4 @@
-<svelte:options immutable={true} tag='v-breadcrumbs' />
+<svelte:options immutable tag='v-breadcrumbs' />
```
micheal-parks, line 1: Cleaned up these immutable attributes.

src/elements/breadcrumbs.svelte
```diff
@@ -14,7 +14,7 @@ $: parsedCrumbs = crumbs.split(',').map((str) => str.trim());
 
 </script>
 
-<div class="flex gap-3 px-4 border border-black rounded-full">
+<div class="inline-flex gap-3 px-4 border border-black rounded-full">
```
micheal-parks, line 17: This again should not be wider than text.

mrloureed (approved):
Awesome LGTM!

.eslintrc.js
DTCurrie: I wouldn't mind maybe using it with `['error', 'avoid']`:

> Only allow braces when there are variable declaration or function declaration which requires a scope.

PR #106, opened 2022-10-10
add input message and success state

sbal13:
Adds ability to add message to input. Increases the styling for varying states (input outline and bg colors) and adds a new success state.

### DIFF
--- a/src/elements/input/input.svelte
+++ b/src/elements/input/input.svelte
@@ -24,7 +24,8 @@ export let step = '1';
 export let name = '';
 export let labelposition: LabelPosition = 'top';
 export let tooltip = '';
-export let state: 'info' | 'warn' | 'error' | '' = 'info';
+export let state: 'info' | 'warn' | 'error'| 'success' | '' = 'info';
+export let message: '';
 
 let root: HTMLElement;
 let input: HTMLInputElement;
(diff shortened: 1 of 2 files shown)


micheal-parks:
@sbal13 Can you add some screenshots of the proposed changes?

micheal-parks:
A few concerns:
* This is a bit noisy (border color, background, and text color changes as well as introducing new text and sometimes a tooltip) and I think that there are some potential layout issues with changing border width dynamically: this could cause some inputs to be misaligned with others that don't have errors / warnings / etc. Thoughts on being a bit more conservative at the start and just changing border color and adding a colored message? Another additional option would be to introduce icons like done here: https://boweryfarming.github.io/femo/?path=/docs/elements-input--default-text

* I understand the use case for error, but what is the use case for warning and success? We already have info and warning tooltips that we display alongside input labels to alert users of things that aren't pressing like errors, and a success state usually transitions the user away from the input.

sbal13:
@micheal-parks 

Yeah I can pare it down to just border color and colored text. 
I didn't have a use-case for success - normally used to indicate that a validation is passing or something like that. Just threw it in there because in other places like table there is a success variant, so was just trying to be consistent.
I thought about using Warn to indicate to a user that they have a malformed filter and that while the request will still run, that that filter will not be applied. Could always be an error state, but there was already a warn state in this component so decided to run with.

sbal13:
@micheal-parks basically i just cut out everything except the error state

PR #112, opened 2022-10-14
Radio select checkmark icon

mrloureed:
_To be deployed after https://github.com/viamrobotics/prime/pull/109_

This adds checkmark icon to selected radios

### DIFF
--- a/src/elements/radio.svelte
+++ b/src/elements/radio.svelte
@@ -30,7 +30,7 @@ const handleClick = (value: string) => {
 
 <!-- svelte-ignore a11y-label-has-associated-control -->
 <label>
-  <div class={cx('flex items-center gap-1.5', {
+  <div class={cx('flex', {
     'pb-1': labelposition === 'top',
   })}>
     {#if label}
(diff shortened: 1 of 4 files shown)


micheal-parks:
Could you add a screenshot of the design?

src/elements/radio.svelte
```diff
@@ -30,7 +30,7 @@ const handleClick = (value: string) => {
 
 <!-- svelte-ignore a11y-label-has-associated-control -->
 <label>
-  <div class={cx('flex items-center gap-1.5', {
+  <div class={cx('flex', {
```
micheal-parks, line 33: Why'd we remove these styles?

src/elements/radio.svelte
```diff
@@ -52,16 +52,30 @@ const handleClick = (value: string) => {
...
-        'bg-black text-white': option === selected,
-      })}
-      on:click={() => handleClick(option)}
-    >
-      {option}
+  <div class="flex">
+    {#each parsedOptions as option}
+      <button
+        bind:this={button}
+        class={cx('capitalize border-y border-l last:border-r border-black px-3 py-2 text-xs', {
```
micheal-parks, line 59: Why did we increase padding here? we're generally trying to keep the UI for app tight / smaller

mrloureed: Good point! How's this?

image

src/elements/radio.svelte
mrloureed: Not sure. Good catch. Fixed.

src/elements/select/select.svelte
```diff
@@ -57,6 +57,9 @@ let open = false;
 let navigationIndex = -1;
```
micheal-parks, line 57: Can we revert the select changes here? It might end up conflicting with new fixes in your other PR.

PR #114, opened 2022-10-24
Add title to button

DTCurrie:
Adds the `title` prop to the button.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.0.71",
+  "version": "0.0.72",
   "lockfileVersion": 2,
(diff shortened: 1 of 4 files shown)


mrloureed (approved):
LGTM

micheal-parks (approved):
Is this to make buttons optionally have more accessible / less ambiguous supplementary text when hovering / for screenreaders?

DTCurrie:
> Is this to make buttons optionally have more accessible / less ambiguous supplementary text when hovering / for screenreaders?

Pretty much, there are places we do this in the `app` previously with just `buttons`, I want to replace them with `v-buttons` and not lose any accessibility. Plus we use them a lot for e2e test element selection.

PR #118, opened 2022-11-04
App 887 responsiveness

mrloureed:
This makes PRIME more responsive. See examples [here](https://viam.atlassian.net/browse/APP-887).


micheal-parks:
@mrloureed Don't forget to bump version!

PR #137, opened 2022-11-29
[DATA-800] make v-pill component

alexis-wei:


src/elements/pill.svelte
```diff
@@ -0,0 +1,35 @@
...
+addStyles();
+
+const handleRemove = () => {
+  dispatch('remove', { value });
+};
+
+</script>
+
+<!-- svelte-ignore a11y-click-events-have-key-events -->
+<div class='flex cursor-pointer items-center max-w-fit gap-1 rounded-xl bg-[#C4C4C4] py-0.5 px-2 text-[10px] hover:bg-gray-300'>
```
sbal13, line 25: think cursor-pointer should be on the v-icon only since it's the only clickable element

src/elements/pill.svelte
```diff
@@ -0,0 +1,35 @@
...
+const dispatch = dispatcher();
+addStyles();
+
+const handleRemove = () => {
+  dispatch('remove', { value });
+};
+
+</script>
+
+<!-- svelte-ignore a11y-click-events-have-key-events -->
```
sbal13, line 24: nit: we're not super focused on accessibility just yet so maybe not a big deal right now, but for accessibility reasons, a user should be able to tab to the x on this pill and press enter to delete this pill. I think this is copy and pasted, but something to consider for the future

sbal13 (changes requested):
Looks good! Just one thing about where the cursor-pointer class is put

src/elements/pill.svelte
alexis-wei: moved!

src/stories/pill.stories.mdx
```diff
@@ -0,0 +1,54 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs'
+
+<Meta
+  title='Elements/Pill'
```
sbal13, line 4: You can get the remove event to show up in the interactions tab in the canvas view by adding this to Meta

```
  parameters={{
    actions: {
      handles: ['remove'],
    },
  }}
  ```
  
  Useful for users who are trying to see how the details will be surfaced to them

File: src/stories/select.stories.mdx [hunk 4]
```diff
@@ -25,6 +25,14 @@ import gpioOptions from './assets/gpio-options'
       description: 'The placeholder value',
       table: { defaultValue: { summary: '' } }
     },
+    showpill: { 
+      description: 'when in multi-select variant, have the option to show pills in the select representing selected values',
+      table: { defaultValue: { summary: 'false' } }
```
sbal13, line 30: nit: If you want to make this controllable you can add `control: { type: 'boolean' }` here. You'd also have to add something like this to each Story too

image

Can reference the `match` and `open` properties in `dropdown.strories.mdx` if you want to make this change, but
it's not super necessary - I think having the example with pills you have already is probably enoug

sbal13 (changes requested):
Sorry just the one additional thing I didn't catch last time about adding the parameter for the interaction panel. The other bit about the control isn't really necessary, more of an FYI if you want to do it

sbal13 (approved):
🪨 ⭐

PR #138, opened 2022-11-29
prevents v-button from being invoked when disabled

sbal13:
Adds a click handler to the v-button that prevents the click handler attached at point of implementation from being invoked. `stopImmediatePropagation` stops event handlers of the same type added after from being invoked - can then choose to dispatch (proceeding with the propagation) if the input is not disabled. Also adds a new dispatcher that copies the dispatch event, allowing the event to pass through as the correct type.

### DIFF
--- a/src/lib/dispatch.ts
+++ b/src/lib/dispatch.ts
@@ -10,4 +10,3 @@ export const dispatcher = () => {
     detail,
   }));
 };
-
--- a/src/lib/dispatch.ts
+++ b/src/lib/dispatch.ts
@@ -14,7 +14,6 @@ export const dispatcher = () => {
 
 export const dispatcherWithEventPassThrough = () => {
   const element = get_current_component() as HTMLElement;
-  // eslint-disable-next-line no-prototype-builtins
   return <EventType extends Event>(name: string, event: EventType) => {
     const copyEvent = new Proxy(event, 
       { get: (target: object, prop: keyof object) => prop === 'composed' || prop === 'bubbles' ? true : target[prop] }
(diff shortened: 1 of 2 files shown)


src/lib/dispatch.ts
```diff
@@ -10,3 +10,16 @@ export const dispatcher = () => {
     detail,
   }));
 };
+
+
+export const dispatcherWithEventPassThrough = () => {
+  const element = get_current_component() as HTMLElement;
+  return <EventType extends Event>(name: string, event: EventType) => {
+    const copyEvent = new Proxy(event,
```
micheal-parks, line 18: I'm slightly worried about overall performance of adding proxy objects for every button in the app, so if we can eventually find a solution that doesn't require proxies I think I'd want to switch to it. Maybe adding an event listener whenever the button becomes disabled that cancels the event and stops propagation would work 🤔. But I think this is great for now! Good stuff.

micheal-parks (approved):

src/lib/dispatch.ts
sbal13: I can probably add a flag that makes this functionality optional - didn't realize Proxies would have that much of a performance hit

micheal-parks: yeah they're in the class of avoid unless necessary for me. They're apparently quite hard to optimize when compiling :( https://gomakethings.com/proxies-vs.-setter-functions-in-state-based-ui-libraries/ and the concern here is cumulative since we add one for every button.

mrloureed (approved):
LGTM

micheal-parks (approved):
🎉

PR #143, opened 2022-12-06
[DATA-804] add reduce sort to select, split out multi-select

alexis-wei:
Functionalities included in this PR:
- added `dosearch` prop, if false, then sort will not be done within component
- added `reducesort` prop, if true, will only show matching results instead of priority listing
- split up v-select to a separate v-multiselect which has functionalities that only apply to multi select. The select component will be refactored in the future to only have single select features
- add `heading` prop, if has value, then will show a heading within the dropdown

heading preview:
Screen Shot 2022-12-06 at 2 49 25 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.0.88",
+  "version": "0.0.89",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@viamrobotics/prime",
-      "version": "0.0.88",
+      "version": "0.0.89",
       "license": "Apache-2.0",
       "devDependencies": {
         "@babel/core": "^7.19.6",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.0.89",
+  "version": "0.0.90",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@viamrobotics/prime",
-      "version": "0.0.89",
+      "version": "0.0.90",
       "license": "Apache-2.0",
       "devDependencies": {
         "@babel/core": "^7.19.6",
--- a/playground/src/App.vue
+++ b/playground/src/App.vue
@@ -99,6 +99,7 @@ setTimeout(() => {
       withbutton="true"
       buttontext="hello"
       buttonicon="x"
+      heading="heading title"
     />
 
     <v-collapse class="p-3" title='Hello world?'>
--- a/playground/src/App.vue
+++ b/playground/src/App.vue
@@ -91,12 +91,16 @@ setTimeout(() => {
 
     <Monaco />
 
-    <v-select 
-      options="ab, bc, cd, dz, e24gsdg, f dskfm 3??"
-      value="a"
-      variant="multiple"
+    <v-multiselect 
+      options="ab, bc, cd, dz, e24gsdg, f dskfm 3??, abs, abd, ac, ad, adb"
+      value="ab, cd"
+      reducesort="true"
       dosearch="true"
+      withbutton="true"
+      buttontext="hello"
+      buttonicon="x"
     />
+
     <v-collapse class="p-3" title='Hello world?'>
       Hello world
     </v-collapse>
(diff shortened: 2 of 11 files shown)


src/elements/select/multiselect.svelte
```diff
@@ -0,0 +1,425 @@
...
+    dispatch('input', { value, values: value.split(','), added: target });
+  } else {
+    dispatch('input', { value, values: value.split(','), removed: target });
+  }
+};
+
+const handleClearAll = () => {
+  value = '';
+  optionsContainer.scrollTop = 0;
+  dispatch('input', { value, values: value.split(',') });
```
sbal13, line 210: can we also dispatch another event here in addition to this called `clear-all-click`? since we don't return which values have been removed like in the other `input` events, the regular `input` event doesn't have enough info for me to write a generic `handleInput` function that accounts for clearing all values. if i attach a separate listener for a clear all event, I can handle that specific scenario independently of input changing

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.0.88",
+  "version": "0.0.89",
```
sbal13, line 3: probably have to bump the version one more time, I think someone just pushed a new version at `0.0.89`

src/elements/select/multiselect.svelte
alexis-wei: sure! will add

playground/src/App.vue
```diff
@@ -91,6 +91,17 @@ setTimeout(() => {
 
     <Monaco />
 
+    <v-multiselect 
+      options="ab, bc, cd, dz, e24gsdg, f dskfm 3??, abs, abd, ac, ad, adb"
+      value="ab, cd"
+      reducesort="true"
+      dosearch="true"
```
micheal-parks, line 98:
```suggestion
      search="true"
```

src/elements/select/select-button.svelte
```diff
@@ -0,0 +1,32 @@
...
+
+export let buttontext = 'ENTER';
+export let buttonicon = '';
+
+const dispatch = dispatcher();
+
+addStyles();
+
+const handleButtonClick = () => {
+  dispatch('button-click');
```
micheal-parks, line 18: We don't need this dispatcher or `button-click` event, since the consumer of the component can just do:
```
<v-select-button on:click={handleClick} />
```

src/lib/sort.ts
```diff
@@ -29,10 +30,19 @@ export const searchSort = (data: string[], searchTerm: string) => {
...
   const finalResults: string[] = [];
 
-  for (const key of Object.keys(results)) {
-    const sorted = (results[key] || []);
-    
-    finalResults.push(...sorted);
+
+  if (reduce) {
+    for (const key of Object.keys(results)) {
+      if (Number(key) !== -1) {
```
micheal-parks, line 36:
```suggestion
      if (Number.parseInt(key, 10) !== -1) {

**Comment by micheal-parks** [comment]
+1 to steven's suggestion, we should probably remove all multiselect capabilities from the current `<v-select>` element. If we want to do that now it's completely fine since we'd want to replace all current usage of `<v-select multiple>` with `<multiselect>` asap anyway.

**Comment by micheal-parks** [comment]
After reading through your new features, I think we should consolidate the `reducesort"`and `dosearch` attribs into one "sort" attrib, like this:
```jsx
type SortOptions = 'default' | 'reduce' | 'off'

<v-multiselect
  sort="option as SortOptions"
/>
```
This moves the behavior to act more like a state machine. The multiselect can either be in default sort mode, reduce sort mode, or no sort mode. This both makes the API and likely our design a little cleaner.

playground/src/App.vue
```diff
@@ -91,6 +91,17 @@ setTimeout(() => {
...
+      withbutton="true"
```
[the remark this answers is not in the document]
alexis-wei: Yeah so we're using it for adding new tags in image tagging

src/elements/select/select-button.svelte
alexis-wei: Do you know why it works when I use `v-select-button` but it doesn't when I import it as a `SelectButton`?

src/lib/sort.ts
alexis-wei: sg!

sbal13 (approved):
Looks good! Just update the lockfile with the new version and then ready to 🚢

PR #144, opened 2022-12-07
fix layout shift by adding placeholder pills in multi-select input

mrloureed:
This fixes a layout shift problem. Feedback is welcome.

https://user-images.githubusercontent.com/17263/206273352-4f68f82f-290b-4a58-98ca-0cef109521f1.mov

### DIFF
--- a/src/elements/pill.svelte
+++ b/src/elements/pill.svelte
@@ -22,7 +22,7 @@ const handleRemove = () => {
 </script>
 
 
-<div class='flex items-center max-w-fit gap-1 rounded-xl bg-[#C4C4C4] py-0.5 px-2 text-[10px] hover:bg-gray-300'>
+<div class='min-h-[20px] flex items-center max-w-fit gap-1 rounded-xl bg-[#C4C4C4] py-0.5 px-2 text-[10px] hover:bg-gray-300'>
   <span>
     {value}
   </span>
--- a/src/elements/select/select.svelte
+++ b/src/elements/select/select.svelte
@@ -326,7 +326,7 @@ $: {
           aria-disabled={isDisabled}
           readonly={isDisabled}
           type='text'
-          class='py-1.5 pl-2.5 pr-1 grow text-xs border-0 bg-white outline-none appearance-none'
+          class='py-1.5 pl-2 pr-1 grow text-xs border-0 bg-white outline-none appearance-none'
           on:input|preventDefault={handleInput}
           on:keyup|stopPropagation|preventDefault={handleKeyUp}
         >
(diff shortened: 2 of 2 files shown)


DTCurrie (approved):
✅

micheal-parks:
I think that Alexis is working on moving these pills outside of the multiselect with her current in-progress work, which should solve this issue. This is also the solution that Peter and I have been wanting to adopt for some time now.

DTCurrie:
> I think that Alexis is working on moving these pills outside of the multiselect with her current in-progress work, which should solve this issue. This is also the solution that Peter and I have been wanting to adopt for some time now.

Does Alexis have a ticket for that work? We should avoid having people fixing bugs that have tickets and not being assigned to them.

And was the solution you and Peter wanted to adopt documented somewhere to be added to the tickets?

micheal-parks:
Yep, that was definitely an oversight on my part. It looks like the solution that was desired was part of this ticket though?

> Layout shift does not occur (Possibly make dropdown box on a layer above with relative/absolute positioning.)
> Bonus: put the tags below the select box rather than within

peterborges:
> https://viam.atlassian.net/browse/APP-657



> > I think that Alexis is working on moving these pills outside of the multiselect with her current in-progress work, which should solve this issue. This is also the solution that Peter and I have been wanting to adopt for some time now.
> 
> Does Alexis have a ticket for that work? We should avoid having people fixing bugs that have tickets and not being assigned to them.
> 
> And was the solution you and Peter wanted to adopt documented somewhere to be added to the tickets?
> 
> Edit: I found the PR, it is DATA-804: https://viam.atlassian.net/browse/DATA-804
> 
> That only refers to sorting behavior and describes no additional solutions for the pills/selection. If we are going to make changes to the components we need to document and track what we intend those changes to be in the related tickets. No one is going to go look at open PRs before pulling in an unassigned ticket.
> 
> I would argue these changes are outside of scope for Alexis' ticket and should not have been included in her PR, and any design changes we wanted to make should have been put into https://viam.atlassian.net/browse/APP-657

Didn't realize there was a ticket for that work already. Will coordinate more closely with the team next time there is work being done to a PRIME components and make sure that work is split out into its own ticket.

mrloureed:
Yeah I need to look for tickets before fixing things too. My bad.

mrloureed:
FYI

`That's correct, and Jason's solution does put them below the select box as far as where the input goes, not at the bottom of the menu which Alexis' does. Another case of unclear requirements.`

If pills are placed as Alexis' does, they will disappear when the select options are closed btw...

micheal-parks:
> If pills are placed as Alexis' does, they will disappear when the select options are closed btw...

The solution that we discussed was to move the pills below the search input's rectangle. So whenever you select an item when the multiselect is open, no more layout shifts will occur because adding pills won't increase the size of the search input.

I don't think we should implement the approach above, because it conflicts with the design language. In our DLS pill = a selected something, whether it be a tag, or a component, etc. But it's used here to add emphasis to a placeholder for content that will arrive after interaction. I think this muddies the design language by training users to associate pills with not what we want to associate them with.

mrloureed:
@micheal-parks @peterborges I'm happy to do as Micheal suggests, "The solution that we discussed was to move the pills below the search input's borders." This was my first idea but I thought ya'll wanted the pills exposed when the options were exposed for whatever reason. Please advise.

micheal-parks:
I don't think it's a big deal if we don't see the pills while the dropdown is open, because we already know which items are selected by the visual feedback given by checked / unchecked items within the dropdown. If anything it's less noisy than the current implementation.

mrloureed:
I am starting a new branch in favor of a different solution.

PR #145, opened 2022-12-07
remove overflow-hidden on collapse component

mrloureed:
Currently items inside v-collapse are getting clipped. This removes overflow-hidden.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.0.88",
+  "version": "0.0.89",
   "lockfileVersion": 2,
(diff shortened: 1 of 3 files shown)


DTCurrie:
Does this mean they are scrollable now?

mrloureed:
This simply shows any overflow content, like absolutely positioned drop downs, and I just realized that I forgot to remove another... see video below

https://user-images.githubusercontent.com/17263/206298948-f59040c5-b3ae-4670-8bae-4e868f5091ce.mov

micheal-parks (approved):
I wonder why I added these overflow statements to begin with, but regardless, looks good!

PR #149, opened 2022-12-09
fix multiselect value check

sbal13:
Oops another fix! When displaying checked/unchecked value, was checking to see if the option was in the stringified list of values which meant that if you had the value set to `"gummy,bears"` and the options included `"y"`, `"y"` would be checked off because "gummy" has a "y" in it. Changed it to first convert the stringified list and then check to see if the option is in that array, which will do a full comparison of values as strings to the target option string

### DIFF
--- a/src/elements/select/multiselect.svelte
+++ b/src/elements/select/multiselect.svelte
@@ -118,11 +118,11 @@ const handleChange = (changedOption: string) => {
   if (parsedSelected.includes(changedOption)) {
     const newValue = [...parsedSelected.filter(item => item !== changedOption)];
     value = newValue.toString();
-    dispatch('input', { value , values: newValue, removed: changedOption });
+    dispatch('input', { value, values: newValue, removed: changedOption });
   } else {
     const newValue = [...parsedSelected, changedOption];
     value = newValue.toString();
-    dispatch('input', { value , values: newValue, added: changedOption });
+    dispatch('input', { value, values: newValue, added: changedOption });
   }
   input.focus();
 };
@@ -178,8 +178,8 @@ const handleIconClick = () => {
 
 const handlePillClick = (target: string) => {
   const newValue = [...parsedSelected.filter((item: string) => item !== target)];
-  value = newValue.toString()
-  dispatch('input', { value , values: newValue, removed: target });
+  value = newValue.toString();
+  dispatch('input', { value, values: newValue, removed: target });
   input.focus();
 };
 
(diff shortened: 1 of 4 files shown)


alexis-wei (approved):
good catch!!

src/elements/select/multiselect.svelte
```diff
@@ -179,8 +175,8 @@ const handleIconClick = () => {
 };
 
 const handlePillClick = (target: string) => {
-  value = [...parsedSelected.filter((item: string) => item !== target)].toString();
-  dispatch('input', { value, values: value.split(',') });
+  const newValue = [...parsedSelected.filter((item: string) => item !== target)];
```
micheal-parks, line 178: Out of curiosity, why are we not syncing the `value` attribute here (and below) anymore?

sbal13: I just felt it wasn't good for the component to manipulate the prop value on its own, especially if we intend for the user to pass in the value. I'm trying to make these components as pure as possible - given the same inputs, will always display the same and output the same values. To that end I'm trying to eliminate as many mutations of outside data as possible, instead dispatching events with mutated data and leaving it up to the point of implementation to decide what to do with the changes (use them, skip them, perform further mutations, etc.)

I'm generally in favor of this because from the usage side, solving bugs is hard when they occur within the blackbox of code that's coming from outside of the code base you're working in.

micheal-parks: I agree with all of this, but web components follow a different set of rules than react or vue components in this regard. The rule that I'm worried about breaking specifically here is this rule from "[Custom Elements Best Practices](https://web.dev/custom-elements-best-practices/)":

> Aim to keep primitive data attributes and properties in sync, reflecting from property to attribute, and vice versa. You never know how a user will interact with your element. They might set a property in JavaScript, and then expect to read that value using an API like getAttribute(). If every attribute has a corresponding property, and both of them reflect, it will make it easier for users to work with your element. In other words, calling setAttribute('foo', value) should also set a corresponding foo property and vice versa. There are, of course, exceptions to this rule. You shouldn't reflect high frequency properties, e.g. currentTime in a video player. Use your best judgment. If it seems like a user will interact with a property or attribute, and it's not burdensome to reflect it, then do so.

So if someone uses PRIME outside of a reactive framework, and attempts to read the latest `value`, then there's good chance that it could be stale. Even if they're in a reactive framework and decide not to use controlled forms then `value` could be stale.

PR #151, opened 2022-12-12
Add playwright with collapsible bugfix and test

micheal-parks:
This PR adds [Playwright](https://playwright.dev/) to PRIME, as well as a CI testing step, and introduces a simple test for the `<v-collapse>` element. By default our Playwright setup here does browser automation testing with Chromium and Safari (desktop), but adding mobile browsers and Firefox is also possible.

We've been wanting to experiment with Playwright as an alternative to Cypress. I think that the results here are promising:
* It was very easy and intuitive to get started. The docs are pretty fantastic and there wasn't any area of the API that I found overly complex or confusing.
* [Testing-library](https://testing-library.com/) like accessibility-based selector functions are built into the library. I think this is a big one for being able to make good selectors, and therefore make tests that don't need to be changed much as we change our code.
* `async / await` is natively supported instead of Cypress's really awkward proprietary promises. The result is much more terse tests that resemble our JS throughout the rest of our repos.
* Tests are parallel by default.
* It's fast. BLAZINGLY fast. (We could probably make this even faster by caching Github action steps, since most time is consumed by downloading browsers)
* All events are native! Instead of synthetic with Cypress. This is a big one IMO. Playing games with reality never ends well with tests and I suspect using native browser events may help with flakiness.
* No need to manually wait for DOM <3
* Will automatically retry (once in our case) if a test fails.

Because of this, I'm going to recommend that we move on to the next phase of experimentation: adding Playwright to APP!


tests-examples/demo-todo-app.spec.ts
```diff
@@ -0,0 +1,437 @@
+import { test, expect, type Page } from '@playwright/test';
```
DTCurrie, line 1: Do we need this?

playground/vite.config.ts
```diff
@@ -23,5 +27,11 @@ export default defineConfig({
   build: {
     target: 'esnext',
     assetsInlineLimit: 0,
+    rollupOptions: {
+      input: {
+        appSchoool: fileURLToPath(new URL('./index.html', import.meta.url)),
+        appStudent: fileURLToPath(new URL('./test.html', import.meta.url)),
```
DTCurrie, line 33: Is the school and student naming convention a playwright thing?

tests-examples/demo-todo-app.spec.ts
micheal-parks: Nope! Considered keeping it in for reference, but if we'd prefer to keep the codebase clean then I'm up for taking it out.

playground/vite.config.ts
micheal-parks: 😂 leftover from an example, I'll give them proper names

micheal-parks:
> I think this all sounds pretty good, also looks like a lot less config/boilerplate when compared to cypress.

Yes! The overall cost / benefits ratio here seems very nice. Quite exciting!

anjinai (approved):

alexis-wei (commented):
Can we add commands to the README on how to run the tests?

Other than that looks good!

michaellee1019:
Nice! Im really excited about:
- `await` making the test flow and structure easier.
- faster and less flaky

PR #152, opened 2022-12-13
Fix select bug and add test

micheal-parks:
This PR restores code that was accidentally removed during [another cleanup PR](https://github.com/viamrobotics/prime/pull/150) and adds a test case to guard against the bug it introduced.


DTCurrie (approved):
✅

anjinai (approved):

PR #155, opened 2022-12-15
Multiselect clear all bug

sbal13:
Multiselect was not clearing all without the parent modifying the value. Internally changes the value to `''` on clear


DTCurrie (approved):
✅

PR #166, opened 2023-01-03
Make "disabled" attributes behave more like their HTML counterparts

micheal-parks:
This PR mostly aligns how the `disabled` attribute for our custom elements behaves with native HTML elements. The goal here is 1. to reduce possible sources of confusion when adding `disabled` and 2. fix some cypress flakiness that may arise from using `aria-disabled`.


.github/workflows/publish.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
       - name: Install and Build 🔧
         run: |
-          npm ci --audit=false --legacy-peer-deps
+          npm ci --audit=false
```
micheal-parks, line 20: This isn't needed after the last storybook update. Legacy deps issues resolved 🎉

src/elements/button/button.svelte
```diff
@@ -54,15 +52,16 @@ const handleParentClick = (e: PointerEvent) => {
   :host { display: inline-block !important }
 </style>
 
+<!-- svelte-ignore a11y-click-events-have-key-events -->
 <svelte:element
   this={tooltip ? 'v-tooltip' : 'span'}
   text={tooltip}
-  on:click={handleParentClick}
+  on:click={isDisabled ? handleParentClick : undefined}
```
micheal-parks, line 59: Refactoring this so that we don't add unnecessary event listeners.

src/elements/slider.svelte
```diff
@@ -446,7 +446,6 @@ const onChange = () => {
         aria-valuetext={value?.toString()}
         aria-orientation='horizontal'
         aria-disabled={disabled}
-        {disabled}
```
micheal-parks, line 449: This is a span, which doesn't support this attribute

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.0.101",
+  "version": "0.0.102",
```
DTCurrie, line 3: Not an issue for this but I think we need to start considering what is a major/minor/patch and updating this accordingly. We have definitely added new APIs in the way of new elements/components and never did minor bump.

src/elements/button/button.svelte
```diff
@@ -54,15 +52,16 @@ const handleParentClick = (e: PointerEvent) => {
...
 >
   <button
     {type}
     aria-label={variant === 'icon' ? label : undefined}
-    aria-disabled={isDisabled}
+    disabled={isDisabled}
```
DTCurrie, line 64: Hm, this will mean the buttons will not be tab-navigable and won't be announced by assistive technology. It will essentially hide the button from them entirely, but I can imagine cases where a PRIME user would want a disabled button to still be recognizable/navigable.

src/elements/input/input.svelte
```diff
@@ -225,13 +226,14 @@ const handleNumberDragDown = async (event: PointerEvent) => {
...
+    {autocomplete}
     {placeholder}
     {name}
     {value}
     inputmode={isNumeric ? 'numeric' : undefined}
     pattern={inputPattern}
-    readonly={isReadonly || isDisabled}
-    aria-disabled={isDisabled}
+    readonly={isReadonly}
+    disabled={isDisabled}
```
DTCurrie, line 236: Same as above. For example for the fragment work we are doing, we want to use readonly inputs that are disabled. With this change fragment configs would be unnavigable and essentially hidden from users with assistive tech entirely. So they would have no idea what a fragments config values are.

package.json
micheal-parks: Yep! agreed that we should start proper versioning at this point.

src/elements/button/button.svelte
micheal-parks: Yeah, this is a good point. Thoughts on keeping `aria-disabled` for button and `disabled` for inputs?

src/elements/input/input.svelte
micheal-parks: For this one, why not use `<v-input readonly>`?

DTCurrie: That should be fine, but we need to support `aria-readonly` for checkbox inputs IIRC

DTCurrie: > The value of `<input type="checkbox">` can not be edited making readonly not relevant. However, when creating checkboxes with role="checkbox" the aria-readonly attribute is supported.

src/elements/button/button.svelte
DTCurrie: That works for me.

src/elements/input/input.svelte
```diff
@@ -13,13 +13,14 @@ import { addStyles } from '../../lib/index';
 import { dispatcher } from '../../lib/dispatch';
 
 export let type: Types = 'text';
+export let autocomplete: string;
 export let placeholder = '';
-export let readonly = 'false';
-export let disabled = 'false';
-export let label = '';
+export let readonly: string;
```
micheal-parks, line 18: This is to align more with `<input>`

micheal-parks:
@DTCurrie Decided to take a slightly less dangerous route for now: I've kept in aria-disabled to avoid regressing accessibility, but I've slightly tweaked how it gets set. Right now svelte will explicitly set that attribute as false if we do `aria-disabled={isDisabled}`. If we do `aria-disabled={isDisabled ? true : undefined}` then the attribute isn't rendered. I suspect that the existence of the attribute, even if false, is what's confusing Cypress.

DTCurrie (approved):
✅

PR #167, opened 2023-01-03
[DATA-995] Empty string check in multiselect

alexis-wei:
fix bug error where the array was not reducing correctly when empty string was passed in and sortoption='off'

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.0.100",
+  "version": "0.0.101",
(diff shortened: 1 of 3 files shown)


sbal13 (approved):
🍰 🥧

sbal13:
Oh wait did you manually update the lockfile? not seeing the same version upgrade in the package.json file

alexis-wei:
> Oh wait did you manually update the lockfile? not seeing the same version upgrade in the package.json file

hmm I'll update the package.json, but it was automatic when I ran `npm I`?

PR #172, opened 2023-01-12
Slider min/max labels no longer break bounds of parent container

mrloureed:
Currently our min and max labels are centered over the right and left edges of the slider module. This causes layout problems when used.

image

Fix:

- Left and right align labels instead, bound to parent container.

image


anjinai (approved):
LGTM

PR #178, opened 2023-02-13
APP-1268: Prime Breadcrumbs E2E Tests

anjinai:
**2023.02.13_APP-1268: Prime Breadcrumbs E2E Tests**

**Items to Note:**

- This PR is comprised of playwright tests to confirm the following behavior for prime breadcrumbs.

Behavior | Scenario
-- | --
Crumbs | GIVEN a "crumbs" attribute has been applied to the v-breadcrumbs element AND the value passed to the "crumbs" attribute is a comma-separated list of strings WHEN the element is rendered THEN the breadcrumbs should render the list of values as pills

<p></p>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.1.1",
+  "version": "0.1.2",
   "license": "Apache-2.0",
   "type": "module",
   "files": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.1.2",
+  "version": "0.1.1",
   "license": "Apache-2.0",
   "type": "module",
   "files": [
--- a/tests/breadcrumbs.spec.ts
+++ b/tests/breadcrumbs.spec.ts
@@ -1,0 +1,16 @@
+import { test, expect } from '@playwright/test';
+
+test('breadcrumbs should render the list of values as pills', async ({ page }) => {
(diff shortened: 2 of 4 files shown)


tests/breadcrumbs.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  await page.goto('/test.html');
+
+  // GIVEN a "crumbs" attribute has been applied to the v-breadcrumbs element
+  // AND the value passed to the "crumbs" attribute is a comma-separated list of strings
+  // WHEN the element is rendered
+  // THEN the breadcrumbs should render the list of values as pills
+
+  const breadcrumbs = page.getByText('Chocolate Chip Oatmeal Raisin')
+  await expect(breadcrumbs).toBeVisible()
+  await expect(breadcrumbs).toHaveCSS('border-color', 'rgb(0, 0, 0)')
```
micheal-parks, line 13: I don't think we should be testing CSS yet (if ever). It's not very valuable for detecting bugs and creates a bit of overheard when we update styles. I'd recommend removing lines 13-14

anjinai: Makes sense!  Thanks, removed.

micheal-parks (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.1.1",
+  "version": "0.1.2",
```
micheal-parks, line 3: nit: I don't think we need a bump since we're only adding tests and not PRIME features

PR #179, opened 2023-02-13
APP-1267: Prime Badge E2E Tests

anjinai:
**2023.02.13_APP-1267: Prime Badge E2E Tests**

**Items to Note:**

- This PR is comprised of playwright tests to confirm the following behaviors for Prime Badges.

Behavior | Scenario
-- | --
Label | GIVEN a "label" attribute has been applied to the v-badge element WHEN the element is rendered THEN the badge should include the passed label text
Variant: default | GIVEN no "variant" attribute has been applied to the v-badge element WHEN the element is rendered THEN the badge should render with a gray background AND the badge should render with gray text
Variant: gray | GIVEN a "variant" attribute has been applied to the v-badge element AND the "variant" attribute value is "gray" WHEN the element is rendered THEN the badge should render with a gray background AND the badge should render with gray text
Variant: green | GIVEN a "variant" attribute has been applied to the v-badge element AND the "variant" attribute value is "green" WHEN the element is rendered THEN the badge should render with a green background AND the badge should render with green text
Variant: orange | GIVEN a "variant" attribute has been applied to the v-badge element AND the "variant" attribute value is "orange" WHEN the element is renderedTHEN the badge should render with a orange background AND the badge should render with orange text
Variant: red | GIVEN a "variant" attribute has been applied to the v-badge element AND the "variant" attribute value is "red" WHEN the element is renderedTHEN the badge should render with a red background AND the badge should render with red text

<p></p>


Kschappacher (approved):
LGTM

tests/badge.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  // Gray Badge Test
+
+  const grayBadge = page.getByTestId("gray")
+  
+  await expect(grayBadge).toBeVisible()
+  await expect(grayBadge).toHaveText("Inactive")
+
+  const childGrayBadge = grayBadge.locator('small').nth(0)
+  
+  await expect(childGrayBadge).toHaveCSS("background-color", "rgb(229, 231, 235)")
```
micheal-parks, line 16: Similar comment here as the other test. I don't think we should be testing for specific CSS values.

micheal-parks: I see that we're trying to test the different states of this pill. I'd instead recommend testing the element's state via running assertions on attributes

micheal-parks: I'm also seeing that the test criteria is specifying color, so if we want to stay true to that then my alternate recommendation is that we should move these colors into the tailwind config file, add them into our theme, and use those theme colors in the pill element + import those theme colors here, so that when design updates happen we won't have to edit both the component and tests.

tests/badge.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  await page.goto('/test.html');
+
+  // Gray Badge Test
+
+  const grayBadge = page.getByTestId("gray")
+  
+  await expect(grayBadge).toBeVisible()
+  await expect(grayBadge).toHaveText("Inactive")
+
+  const childGrayBadge = grayBadge.locator('small').nth(0)
```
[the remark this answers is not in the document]
anjinai: Thanks, updated to use a text selector!

tests/badge.spec.ts
anjinai: I moved the colors needed for the badge variants into the tailwind config.  They render correctly when I run the storybook page locally.  However, when I run my playwright tests, the colors I added to the tailwind config weirdly don't render.  Do you happen to have any suggestions?

tests/badge.spec.ts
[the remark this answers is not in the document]
micheal-parks: Yay thank you!

tests/badge.spec.ts
micheal-parks: Oh that's probably because the "playground" has its own tailwind config. Getting that working should be as simple as exporting that theme object from the main config into the playground config.

anjinai: Thank you!!!! That totally did the trick!

tests/badge.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  // Gray Badge Test
+
+  const grayBadge = page.getByText('Inactive')
+  
+  await expect(grayBadge).toBeVisible()
+  await expect(grayBadge).toHaveText("Inactive")
+  
+
+  await expect(grayBadge).toHaveCSS("background-color", hexToRGB(config.theme.extend.colors['gray/200']))
```
anjinai, line 36: I believe that playwright css color assertions only support rgb so I needed to convert from hex to rgb to get this to work.

tests/badge.spec.ts
```diff
@@ -0,0 +1,79 @@
+import { test, expect } from '@playwright/test';
+import config from '../tailwind.config.cjs'
+
+const hexToRGB = (hex) => {
```
anjinai, line 4: I originally had this func in another file, but I was having difficulty importing it into badge.spec.ts.

micheal-parks (approved):

PR #180, opened 2023-02-13
APP-1274: Prime Tabs E2E Tests

anjinai:
**2023.02.13_APP-1274: Prime Tabs E2E Tests**

**Summary:** This PR is comprised of Playwright tests to confirm the following behaviors for Prime Tabs.

**Items to Note:**

- The tests for click and keydown enter call for tests which verify that the input event emits the selected tab value.  I was unable to figure out how to capture this info so I resolved to test that the clicked tab displays the correct class of bg-white (the selected tab should change to white while the rest of the tabs should be gray).  **@Kschappacher -** **I was wondering if you knew of a way to capture the selected tab value/do you think my class tests are sufficient?**
- Additionally, the ticket asks for a default tab test.  However, v-tabs has no default behavior.  Per MP's comment below, I intend to ask DC if adding default behavior to v-tabs is necessary when he's back from vacay. 

Behavior | Scenario
-- | --
Tabs | GIVEN a "tabs" attribute has been applied to the v-tabs element AND the value passed to the "tabs" attribute is a comma-separated list of strings WHEN the element is rendered THEN the tabs should render the list of values as tabs
**Selected: default** | **GIVEN a "tabs" attribute has been applied to the v-tabs element AND the value passed to the "tabs" attribute is a comma-separated list of strings AND no "selected" attribute has been applied to the v-tabs element WHEN the element is rendered THEN the tabs should render the list of values as tabs AND the first tab should render as selected**
Selected: value | GIVEN a "tabs" attribute has been applied to the v-tabs element AND the value passed to the "tabs" attribute is a comma-separated list of strings AND a "selected" attribute has been applied to the v-tabs element AND the value passed to the "selected" attribute matches one of the values passed to the "tabs" attribute WHEN the element is rendered THEN the tabs should render the list of values as tabs AND the tab that matches the "selected" value should render as selected
Click | GIVEN a "tabs" attribute has been applied to the v-tabs element AND the value passed to the "tabs" attribute is a comma-separated list of strings WHEN the element is rendered AND a tab is clicked THEN the tabs should set that tab as the selected tab AND the tabs should emit an "input" event with the selected tab value
Keydown | GIVEN a "tabs" attribute has been applied to the v-tabs element AND the value passed to the "tabs" attribute is a comma-separated list of strings WHEN the element is renderedAND a tab is focused AND the "Enter" key is pressed THEN the tabs should set that tab as the selected tab AND the tabs should emit an "input" event with the selected tab value

<p></p>


tests/tabs.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  const tab3Selected = waitForCustomEvent(page,'input')
+  await tab3.click()
+  await expect(tab3Selected).toBeTruthy()
+
+  // Check That New Selected Tab Has Correct Background Color of White (Tab 3)
+  await expect(tab1).not.toHaveCSS("background-color", "rgb(255, 255, 255)")
+  await expect(tab2).not.toHaveCSS("background-color", "rgb(255, 255, 255)")
+  await expect(tab3).toHaveCSS("background-color", "rgb(255, 255, 255)")
+
+  // Check That If No Selected Value, Default Selected Is First Value (Tab A)
```
anjinai, line 45: In regard to the default test, currently v-tabs has no default behavior. 
 When no selected attribute has been defined, no tab is rendered as selected.  The test asks that the first tab be selected.  Should I add this behavior into tabs.svelte?

micheal-parks: I think I disagree with that behavior. We can discuss with devin when he gets back but I think that nothing being selected is both useful and less surprising / trying to be less magical when there's no selected attribute. I'd instead test the current behavior for now.

tests/tabs.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  await expect(tab2).toHaveText('Tab 2')
+  await expect(tab3).toHaveText('Tab 3')
+
+  // Check Selected Attribute is Correct (Tab 2)
+  await expect(selectedTestTabs).toHaveAttribute("selected", "Tab 2")
+
+  // Check Selected Tab Has Correct Background Color of White (Tab 2)
+  await expect(tab1).not.toHaveClass(/bg-white/)
+  await expect(tab2).toHaveClass(/bg-white/)
+  await expect(tab3).not.toHaveClass(/bg-white/)
```
ehhong, line 33: [opt] not sure if it is redundant, but should we also check that the unselected tabs have the correct background color? e.g. `expect(tab1).toHaveClass(/bg-gray/)` or whatever the color is, or maybe we should check that the unselected tabs have the same background color

playground/src/tabs-test.vue
```diff
@@ -0,0 +1,31 @@
+<script setup lang="ts">
+
+</script>
+
+<template>
+  <main class="m-3">
+
+    <!-- Selected Value -->
+    <v-tabs
+      data-testid="Selected Test"
```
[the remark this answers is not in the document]
anjinai: All set - thanks!

playground/src/tabs-test.vue
```diff
@@ -7,21 +7,21 @@
 
     <!-- Selected Value -->
     <v-tabs
-      data-testid="Selected Test"
+      data-testid="selected-test"
```
ehhong, line 10: [nit] so sorry should've mentioned earlier, but probably also good to have the component type in the test name for clarity, maybe `tabs-selected`, `tabs-default`, `tabs-key-enter`

playground/src/tabs-test.vue
```diff
@@ -7,21 +7,21 @@
...
-      data-testid="Default Test"
+      data-testid="default-test"
       class="mt-2"
       tabs="Tab A, Tab B, Tab C"
     />
 
     <!-- Focus KeyDown Test -->
     <v-tabs
-      data-testid="Key Enter Test"
+      data-testid="key-enter-test"
```
ehhong, line 24: [q] is there a reason why you wanted a separate component for the focus keydown test rather than reusing `selected-test`? they seem functionally equivalent to me

tests/tabs.spec.ts
anjinai: In this case, the background color class for gray (bg-black/20) is on the parent and not on the individual buttons for the tabs so I can't test for gray the same way I'm testing for class bg-white on the selected tab.

playground/src/tabs-test.vue
anjinai: all set

playground/src/tabs-test.vue
Kschappacher: you are potentially altering the state of the component when clicking a component. So having them separated isn't a bad idea

Kschappacher:
@anjinai is this still relevant? ` @Kschappacher - I was wondering if you knew of a way to capture the selected tab value/do you think my class tests are sufficient?` I saw you used waitForCustomEventTimeout in your tests to track the input event.

ehhong (approved):
lgtm :)

Kschappacher (approved):
Have one question I tagged ya in. Otherwise looks good!

anjinai:
> @anjinai is this still relevant? ` @Kschappacher - I was wondering if you knew of a way to capture the selected tab value/do you think my class tests are sufficient?` I saw you used waitForCustomEventTimeout in your tests to track the input event.

Yes, so waitForCustomEvent captures if the event fired, but I can't seem to capture the value of the input event.  So in this case, on click an input event fires with the value of the selected tab.  Just wanted to check if you knew of a way to get that value.   

Screen Shot 2023-02-28 at 12 04 02 PM

tests/tabs.spec.ts
anjinai: @DTCurrie - just wanted to check if you had an opinion on v-tabs not having default behavior.  The ticket asks for a default test which shows the first tab as selected when no selected tab is specified.  Currently, that behavior doesn't exist in v-tabs.

Kschappacher:
> > @anjinai is this still relevant? ` @Kschappacher - I was wondering if you knew of a way to capture the selected tab value/do you think my class tests are sufficient?` I saw you used waitForCustomEventTimeout in your tests to track the input event.
> 
> Yes, so waitForCustomEvent captures if the event fired, but I can't seem to capture the value of the input event. So in this case, on click an input event fires with the value of the selected tab. Just wanted to check if you knew of a way to get that value.
> 
> Screen Shot 2023-02-28 at 12 04 02 PM

So there might be a way for us to wait on the value of the promise from waitForCustomEvent and see if it contains the value we are looking for. I dont think it is necessary in this pr since you are checking that things change when the button is clicked. Its a nice to have and def not a blocker for this pr

tests/tabs.spec.ts
anjinai: Per an offline discussion with DC, he confirmed it's fine to remove the default behavior test from the ticket.

PR #181, opened 2023-02-13
Data 1075 list box component

sbal13:
Adding this component to PRIME: https://www.figma.com/file/um4JSCwIrNL4vTx1c37Clm/Data?node-id=2764%3A28813&t=t6V3K0EWcWWOnbeK-4

I started writing tests - wanted to write one for disabled state but had some issues with non-unique test ids. Purely icon buttons don't have any identifiable text to locate so had to use test ids for those, hence the issue I was facing.

It might be useful to figure out a way to do isolated renders for each test, but not sure what the right setup for that is in Playwright. I'd research it didn't want to dally on this work as it's needed for features in app.


src/elements/list-box.svelte
```diff
@@ -0,0 +1,159 @@
...
+import { addStyles } from '../lib/index';
+import { htmlToBoolean } from '../lib/boolean';
+
+export let disabled: string;
+export let left = '';
+export let right = ''; 
+export let leftlabel = ''; 
+export let rightlabel = ''; 
+export let height = '200px';
+
```
alexis-wei, line 20: Would add empty state value options for both the left and right, so when there are no options, then the user can input "No models on robot" or any other info. Ideally make sure that it is linkable so people can get to the documentation or anything else that is necessary

src/elements/list-box.svelte
```diff
@@ -0,0 +1,159 @@
...
+      />
+    </button>
+  </div>
+  <div class="w-full flex flex-col gap-2 self-stretch" style={`height: ${height};`}>
+    <span class="text-xs">{ rightlabel }</span>
+    <div class="border border-[#D7D7D] grow px-4 py-3 bg-white flex flex-col overflow-auto">  
+      {#each options.right as option}
+      <button
+        class={cx('flex items-start', { 'bg-[#E2F1FD]': option.selected })}
+        on:click={() => handleOptionClick(option, RIGHT)}
```
alexis-wei, line 152: Double check with @peterborges and @agreenb , but the last I heard about the finalized design decision for this element, it was for the selection to have both the check boxes and the highlight. Where the highlight would be on click, but not select and checking the checkbox would actually select the item to the moveable to the other side

src/elements/list-box.svelte
```diff
@@ -0,0 +1,159 @@
...
+      data-testid="move-left"
+    >
+      <i
+        class="icon-arrow-up"
+      />
+    </button>
+  </div>
+  <div class="w-full flex flex-col gap-2 self-stretch" style={`height: ${height};`}>
+    <span class="text-xs">{ rightlabel }</span>
+    <div class="border border-[#D7D7D] grow px-4 py-3 bg-white flex flex-col overflow-auto">
```
alexis-wei, line 148: the padding looks a bit too thick (would try p-2). Would decrease padding and keep padding on top, bottom, right the same, and a bit thicker on the right due to the scroll bar?? Maybe the same in all 4 directions if the padding moves automatically when a scroll exists.

alexis-wei, line 148: the color is #D7D7D9, you're missing the 9 so I think the shade is a bit off

src/elements/list-box.svelte
```diff
@@ -0,0 +1,159 @@
...
+    options: JSON.parse(JSON.stringify(options)) as Options,
+  });
+};
+
+</script>
+<div
+  class={cx('w-full grid grid-cols-[1fr_32px_1fr] gap-8 items-center p-2', { '!text-black/50': isDisabled })}
+>
+  <div class="w-full flex flex-col gap-2 self-stretch" style={`height: ${height};`}>
+    <span class="text-xs">{ leftlabel }</span>
```
alexis-wei, line 110: Update color for this and also right label to #4E4F52 to match design

src/elements/list-box.svelte
```diff
@@ -0,0 +1,159 @@
...
+      on:click={() => handleMoveClick(LEFT)}
+      data-testid="move-left"
+    >
+      <i
+        class="icon-arrow-up"
+      />
+    </button>
+  </div>
+  <div class="w-full flex flex-col gap-2 self-stretch" style={`height: ${height};`}>
+    <span class="text-xs">{ rightlabel }</span>
```
alexis-wei, line 147: reminder to update color to #4E4F52

micheal-parks (approved):
Good stuff!

src/elements/list-box.svelte
```diff
@@ -0,0 +1,159 @@
...
+// disabled to pass untyped object to cx library in template
+
+import cx from 'classnames';
+import { afterUpdate } from 'svelte';
+
+import { dispatcher } from '../lib/dispatch';
+import { addStyles } from '../lib/index';
+import { htmlToBoolean } from '../lib/boolean';
+
+export let disabled: string;
```
alexis-wei, line 14: shouldn't we set the default of this to be false?

alexis-wei (changes requested):
wohoo nice adding this component! A couple of minor design details, and would double check with Peter and Alexa on the finalized design for this!

playground/src/Test.vue
```diff
@@ -19,5 +19,12 @@
       options="Jolly logs, Sleeping hogs, Sticky bogs, Fiendish dogs"
       placeholder="Select a Bobbins"
     />
+    <v-list-box
+      left="Charmander,Pikachu,Venusaur"
```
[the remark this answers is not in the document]
sbal13: it became Venusaur duh

src/elements/list-box.svelte
sbal13: Oh you right I actually completely forgot about this. Will add

src/elements/list-box.svelte
```diff
@@ -0,0 +1,159 @@
+<svelte:options immutable tag='v-list-box' />
```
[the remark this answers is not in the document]
sbal13: yeah i just went with what Peter called it but I couldn't think of anything either. box-select?? transfer-select?? exchange-select?? i kind of like that last one

src/elements/list-box.svelte
sbal13: Since `isDisabled` is the boolean converted value I'm leaving this defaulted to undefined to match how users would use it - they either have the disabled attribute on the component which will accept the value as a string or it's not there at all and therefore undefined.

src/stories/list-box.stories.mdx
```diff
@@ -0,0 +1,94 @@
...
+      <v-list-box
+        left="${left}"
+        right="${right}"
+        leftlabel="${leftlabel}"
+        rightlabel="${rightlabel}"
+        disabled="${disabled}"
+        height="${height}"
+      />
+    `}
+  </Story>
```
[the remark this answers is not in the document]
sbal13: + on the empty state but i intentionally left out the height one as its own story as its not really a state of the component and more like a visual modification (that they can control via the controls in Storybook if they want to change it) - we've created a story for almost every prop for every component, but I think we can just focus on the big ones that actually show case functional differences

sbal13:
@alexis-wei added empty state and tests, and i think i got all of the other small stuff

src/elements/list-box.svelte
```diff
@@ -0,0 +1,169 @@
...
+    options: JSON.parse(JSON.stringify(options)) as Options,
+  });
+};
+
+</script>
+<div
+  class={cx('w-full grid grid-cols-[1fr_32px_1fr] gap-8 items-center p-2', { '!text-black/50': isDisabled })}
+>
+  <div class="w-full flex flex-col gap-2 self-stretch" style={`height: ${height};`}>
+    <span class="text-xs text-[#4E4F52]">{ leftlabel }</span>
```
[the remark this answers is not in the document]
alexis-wei: Yeah we're grabbing the hex colors from Figma! +1 on adding them to the theme

sbal13:
@micheal-parks @alexis-wei added!

src/elements/list-box.svelte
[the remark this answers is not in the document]
sbal13: done!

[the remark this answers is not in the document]
micheal-parks: Nice!

src/elements/list-box.svelte
```diff
@@ -112,11 +137,14 @@ const handleMoveClick = (target: ListBoxSide) => {
...
        {#each options.left as option}
           <button
-            class={cx('flex items-center px-2 py', { 'bg-focus/highlight': option.selected })}
+            class={cx('flex items-center px-2 py text-sm', { 'bg-focus/highlight': option.selected })}
             on:click={() => handleOptionClick(option, LEFT)}
           > 
             <input type='checkbox' checked={option.selected} disabled={isDisabled}/>       
-            <span class="text-sm px-4">{ option.value }</span>
+            <span class="px-4">{ option.value }</span>
+            {#if displaySuffix && option.suffix}
```
alexis-wei, line 145: right now, if you put a super long title/suffix, it'll expand the box and squish the other one, although it won't occur much in current use cases, might be a nice thing to fix later. But personally would make the letters hidden at some point. Fine for now
Screen Shot 2023-02-15 at 2 02 46 PM

src/stories/list-box.stories.mdx
```diff
@@ -52,17 +57,19 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs'
       leftlabel: 'Team',
       rightlabel: 'Box 1',
       disabled: false,
-      height: '200px'
+      height: '200px',
+      suffix: false
```
alexis-wei, line 61: worth making this value false by default

src/stories/list-box.stories.mdx
```diff
@@ -102,21 +111,50 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs'
...
       >
         <div slot="left-empty">Left empty state</div>
         <div slot="right-empty">Right empty state</div>
       </v-list-box>
     `}
   </Story>
+</Canvas>
+
+<Canvas>
+  <Story
```
alexis-wei, line 136: yayy!

alexis-wei (approved):
Looking good!! tested locally and all the features work smoothly!! Just a comment about sizing and some questions

src/elements/list-box.svelte
```diff
@@ -36,25 +38,48 @@ const LEFT = 'left';
 const RIGHT = 'right';
 
 let isDisabled: boolean;
+// needs to be calculated on initial render
```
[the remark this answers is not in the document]
sbal13: Timing thing. It just doesn't matter as much for `isDisabled` because it gets set pretty soon after. The problem with `displaySuffix` is that it's needed for the `let options = {...}` calculation below, which isn't set by the time the `generateOptions` function runs, so is always `undefined` because the `$` block that sets it will only run after the initial render, which is too late

src/elements/list-box.svelte
sbal13: overflow-x?

PR #184, opened 2023-02-22
update button disabled

alexis-wei:
updating what the button looks like when disabled so it works for all variants. 

Screen Shot 2023-02-22 at 5 25 05 PM


DTCurrie (approved):
✅

PR #190, opened 2023-03-07
App 1341: Don't break the select component when adding special chars to the input

RoxyFarhad:
The bug was that when we add special characters to the v-select, we convert the string to a regex. This means if there are special characters in the string, it will break the component, and ultimately the page. This adds backslashes to any special character in the string, and will treat that as a character to match on rahter than a special character.

### DIFF
--- a/src/elements/select/utils.ts
+++ b/src/elements/select/utils.ts
@@ -1,3 +1,5 @@
+import { addSpecialCharacterEscapes } from '../../lib/sort';
+
 interface Match { 
   search: string[],
   option: string,
@@ -24,8 +26,10 @@ export const applySearchHighlight = (options: string[], value: string) => {
   const matches = [];
   const noMatches = [];
 
+  const valueCopy = addSpecialCharacterEscapes(value);
+
   for (const option of options) {
-    const match = option.match(new RegExp(value, 'i'));
+    const match = option.match(new RegExp(valueCopy, 'i'));
 
     if (match?.index === undefined) {
       noMatches.push({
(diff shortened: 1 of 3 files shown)


src/elements/select/utils.ts
```diff
@@ -24,8 +24,20 @@ export const applySearchHighlight = (options: string[], value: string) => {
...
+  let valueCopy = value;
+  let i = 0;
+  while (i < valueCopy.length) {
+    if (/[^\dA-Za-z]/.test(valueCopy[i] as string)) {
+      /* eslint-disable no-useless-escape */
+      valueCopy = `${valueCopy.slice(0, i)}\\${valueCopy.slice(i, valueCopy.length)}`;
+      i += 1;
+    }
+    i += 1;
+  }
```
ehhong, line 37: [opt] this feels hard to read, would add a quick comment explaining the logic more specifically here (prepend special characters with '/' to create appropriate regex). any reason you didn't construct `valueCopy` while iterating over `value`? might simplify this a bit

RoxyFarhad: yeah I can change it to do that too! just did the quickest that came to mind but will add thanks!

src/lib/sort.ts
```diff
@@ -1,9 +1,21 @@
 /* eslint-disable unicorn/prefer-regexp-test */
+export const addSpecialCharacterEscapes = (value: string) => {
```
RoxyFarhad, line 2: main change for a re-usable function

ehhong:
do you want to add a quick test for this?

ehhong (approved):

PR #191, opened 2023-03-07
[APP-1364] Add modal e2e tests

ehhong:
**Notes**
- I made a separate page for the modal tests, the open modal was interfering with all the other tests in `test.html` -- is there a better way to do this than replicate all the `.html`, `.ts`, `.vue` files? Is it worth the effort to make this cleaner / dynamically generate some of these files?
- Updated README to add details about running individual tests

Behavior | Scenario
-- | --
Open: default | GIVEN no "open" attribute has been applied to the v-modal elementWHEN the element is renderedTHEN the modal should be hidden
Open: false | GIVEN an "open" attribute of "false" has been applied to the v-modal elementWHEN the element is renderedTHEN the modal should be hidden
Open: true | GIVEN an "open" attribute of "true" has been applied to the v-modal elementWHEN the element is renderedTHEN the modal should be visible
Title | GIVEN a "title" attribute has been applied to the v-modal elementAND an "open" attribute of "true" has been applied to the v-modal elementWHEN the element is renderedTHEN the modal should be visibleAND it should have a title header with the "title" attribute value
Message | GIVEN a "message" attribute has been applied to the v-modal elementAND an "open" attribute of "true" has been applied to the v-modal elementWHEN the element is renderedTHEN the modal should be visibleAND it should have a message with the "message" attribute value at the top of the modal body
Click Overlay | GIVEN an "open" attribute of "true" has been applied to the v-modal elementWHEN the element is renderedAND the overlay behind the modal is clickedTHEN the modal should be hidden
Keydown Overlay | GIVEN an "open" attribute of "true" has been applied to the v-modal elementWHEN the element is renderedAND the overlay behind the modal is focusedAND the "Enter" key is pressedTHEN the modal should be hidden
Click Close | GIVEN an "open" attribute of "true" has been applied to the v-modal elementWHEN the element is renderedAND the "x" close icon is clickedTHEN the modal should be hidden
Keydown Close | GIVEN an "open" attribute of "true" has been applied to the v-modal elementWHEN the element is renderedAND the "x" close icon is focusedAND the "Enter" key is pressedTHEN the modal should be hidden
Slot | GIVEN a child element in included with the v-modalAND an "open" attribute of "true" has been applied to the v-modal elementWHEN the element is renderedTHEN the modal should render the child element in the modal body


playground/src/modal-test.ts
```diff
@@ -0,0 +1,6 @@
+import './main.css'
+import '../../src/main'
+import { createApp } from 'vue'
+import App from './modal-test.vue'
+
+createApp(App).mount('#app')
```
ehhong, line 6: is there a way we could dynamically generate this file and the `.html` file? we should probably move tests to their individual pages and off of `test.html`

tests/modal.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  await page.goto('/modal-test.html');
+
+  // No "open" attribute applied
+  await expect(page.getByTestId('modal-default')).toBeHidden()
+
+  // open="false" attribute applied
+  await expect(page.getByTestId('modal-closed')).toBeHidden()
+
+  // open="true" attribute applied
+  await expect(page.getByTestId('modal-open')).toHaveJSProperty('hidden', false)
```
ehhong, line 13: using `toBeVisible()` doesn't work here because Playwright's [definition of visible](https://playwright.dev/docs/actionability#visible) marks any element with a `hidden` attribute as _not_ visible, even if the value is `false`

README.md
```diff
@@ -67,6 +67,9 @@ To test:
 npm run test-dev # to test with the playwright debug UI
 # or
 npm run test # to test in headless mode
+
+# to run specific tests
+npm run test <test-name> # for <test-name>.spec.ts
```
micheal-parks, line 72: 👏

playground/src/modal-test.ts
micheal-parks: Yeah we should probably make the testbed into a single page app to avoid all this boilerplate. We could pretty easily do it with Nuxt or Sveltekit in the near future.

tests/modal.spec.ts
micheal-parks: Oh weird, it looks like we're not using the `hidden` attribute though, and instead toggling the `invisible` class (and thereby toggling `visibility: hidden;`) to set visibility, and the open modal should not have the CSS `visibility` computed property at all.

micheal-parks: Maybe it's considering it not visible because the `<v-modal>` element itself has zero size - its children do. Do visibility tests work when instead selecting the first element with a size within the modal instead?

ehhong: you're totally correct, fixed

playground/src/modal-test.ts
ehhong: should i make a ticket for this?

micheal-parks (approved):

playground/src/modal-test.ts
micheal-parks: Yeah that would be great! If you do add a comment here linking the ticket you created.

ehhong: https://viam.atlassian.net/browse/APP-1383

PR #192, opened 2023-03-07
APP-1370 - Prime Radio E2E Tests

anjinai:
**2023.03.07_APP-1370: Prime Radio E2E Tests**

**Items to Note:**

- This PR is comprised of playwright tests to confirm the following behaviors for Prime Radio Buttons.

Behavior | Scenario
-- | --
Options | GIVEN an "options" attribute has been applied to the v-radio elementAND the value passed to the "options" attribute is a comma-separated list of stringsWHEN the element is renderedTHEN the radio should render the list of options as buttons
Selected: value | GIVEN an "options" attribute has been applied to the v-radio elementAND the value passed to the "options" attribute is a comma-separated list of stringsAND a "selected" attribute has been applied to the v-radio elementAND the value passed to the "selected" attribute matches one of the values passed to the "options" attributeWHEN the element is renderedTHEN the radio should render the list of values as buttonsAND the button that matches the "selected" value should render as selected
Click | GIVEN a "options" attribute has been applied to the v-radio elementAND the value passed to the "options" attribute is a comma-separated list of stringsWHEN the element is renderedAND a button is clickedTHEN the radio should set that button as the selected optionAND the radio should emit an "input" event with the selected option value
Keydown | GIVEN a "options" attribute has been applied to the v-radio elementAND the value passed to the "options" attribute is a comma-separated list of stringsWHEN the element is renderedAND a button is focusedAND the "Enter" key is pressedTHEN the radio should set that button as the selected optionAND the radio should emit an "input" event with the selected option value
Label: default position | GIVEN a "label" attribute has been applied to the v-radio elementAND no "label-position" attribute has been applied to the v-radio elementWHEN the element is renderedTHEN the "label" attribute value should be rendered above the radio
Label: "top" position | GIVEN a "label" attribute has been applied to the v-radio elementAND a "label-position" attribute of "top" has been applied to the v-radio elementWHEN the element is renderedTHEN the "label" attribute value should be rendered above the radio
Label: "left" position | GIVEN a "label" attribute has been applied to the v-radio elementAND a "label-position" attribute of "left" has been applied to the v-radio elementWHEN the element is renderedTHEN the "label" attribute value should be rendered inline before the radio
Tooltip: default state | GIVEN a "tooltip" attribute has been applied to the v-radio elementAND no "state" attribute has been applied to the v-radio elementAND no "label-position" attribute has been applied to the v-radio elementWHEN the element is renderedTHEN an info icon should render above the radio
Tooltip: info state | GIVEN a "tooltip" attribute has been applied to the v-radio elementAND a "state" attribute of "info" has been applied to the v-radio elementAND no "label-position" attribute has been applied to the v-radio elementWHEN the element is renderedTHEN an info icon should render above the radio
Tooltip: warn state | GIVEN a "tooltip" attribute has been applied to the v-radio elementAND a "state" attribute of "warn" has been applied to the v-radio elementAND no "label-position" attribute has been applied to the v-radio elementWHEN the element is renderedTHEN an orange warn icon should render above the radio
Tooltip: error state | GIVEN a "tooltip" attribute has been applied to the v-radio elementAND a "state" attribute of "error" has been applied to the v-radio elementAND no "label-position" attribute has been applied to the v-radio elementWHEN the element is renderedTHEN a red error icon should render above the radio
Tooltip: "left" position | GIVEN a "tooltip" attribute has been applied to the v-radio elementAND a "label-position" attribute of "left" has been applied to the v-radio elementWHEN the element is renderedTHEN the tooltip icon should be rendered inline before the radio
Tooltip: hover | GIVEN a "tooltip" attribute has been applied to the v-radio elementWHEN the element is renderedAND the tooltip icon is hoveredTHEN the "tooltip" attribute value should render as a tooltip
Readonly | GIVEN a "options" attribute has been applied to the v-radio elementAND the value passed to the "options" attribute is a comma-separated list of stringsAND a "readonly" attribute of "true" has been applied to the v-radioWHEN the element is renderedTHEN it the radio should render as readonly
Click: readonly | GIVEN a "options" attribute has been applied to the v-radio elementAND the value passed to the "options" attribute is a comma-separated list of stringsAND a "readonly" attribute of "true" has been applied to the v-radioWHEN the element is renderedAND a button is clickedTHEN the radio should not changed the selected optionAND the radio should not emit an "input" event with the selected option value
Keydown: readonly | GIVEN a "options" attribute has been applied to the v-radio elementAND the value passed to the "options" attribute is a comma-separated list of stringsAND a "readonly" attribute of "true" has been applied to the v-radioWHEN the element is renderedAND a button is focusedAND the "Enter" key is pressedTHEN the radio should not changed the selected optionAND the radio should not emit an "input" event with the selected option value


tests/radio.spec.ts
```diff
@@ -0,0 +1,134 @@
+import { test, expect } from '@playwright/test';
+import { waitForCustomEvent, waitForCustomEventTimeout } from './lib/helper.ts'
+
+
+test('Radio E2E Test, Options, Selected, Click, Keydown ', async ({ page }) => {
```
ehhong, line 5: i think it's generally good practice to name tests by the action they're testing (i.e. Renders options appropriately, Responds to click, etc), kinda like naming PRs. this makes it easier to debug when tests break as well. it's fine to omit the component name because the tests are already named

a good example in the code is `listbox.spec.ts` -- lmk if you agree, we should probably standardize this moving forwards

ehhong: also would be good to split up these tests (options, selected, click, keydown) for the same debugging reasons

tests/radio.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  await page.keyboard.press('Enter')
+
+  await expect(opt4).toHaveClass(/bg-white/)
+  await expect(opt5).toHaveClass(/bg-white/)
+  await expect(opt6).toHaveClass(/bg-gray-9/)
+});
+
+
+test('Radio E2E Test, Label Position', async ({ page }) => {
+  await page.goto('/test.html');
```
ehhong, line 67: this line can be put in a [`beforeEach`](https://playwright.dev/docs/api/class-test#test-before-each) hook

tests/radio.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+test('Radio E2E Test, Readonly', async ({ page }) => {
+  await page.goto('/test.html');
+
+  // Readonly, Click
+  const readonly2 = page.getByRole('button', { name: 'Readonly 2' })
+  await expect(readonly2).toBeVisible()
+
+  const readonly2Selected = waitForCustomEventTimeout(page,'input')
+  await readonly2.press('Enter')
+  await readonly2Selected
```
ehhong, line 124: [q] do we need to assert this value against something? or will it just hang if the event is not emitted?

tests/radio.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  
+  // Readonly Keydown
+  const readonly5 = page.getByRole('button', { name: 'Readonly 5' })
+  await expect(readonly5).toBeVisible()
+
+  const readonly5Selected = waitForCustomEventTimeout(page,'input')
```
ehhong, line 130: i think `waitForCustomEventTimeout` adds an event listener under the event `input` and returns true if it is received once. not sure what calling `waitForCustomEventTimeout` twice with the same event name would do (are two event listeners with the same name added? does the old one get deleted and reset?), would just double check to see that `readonly5selected` does indeed fail if the enter press on `readonly5` doesn't happen

ehhong (commented):
looks good! just a couple style nits and questions. i didn't know about location-based testing before this and will be adding that to my tests 🙈

tests/radio.spec.ts
anjinai: Thanks, good point!  Will update my tests to match list box's naming convention + break tests up by action.  I now realize my tests are so boring...I wish I wrote pokemon themed tests now.  

I should probably go back and update the other PRIME tests I've written to also match this convention.

tests/radio.spec.ts
anjinai: We don't need to assert this value against something in this file as the assertion occurs within the helper function.  If the event is not emitted (which is the behavior we expect in this case), then it'll timeout.  

Screen Shot 2023-03-10 at 10 53 40 AM

tests/radio.spec.ts
ehhong: me too... lol

tests/radio.spec.ts
ehhong: ah sorry ignore this i was thinking of `waitForCustomEvent`, didn't realize this was readonly

tests/radio.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  await expect(opt2).toHaveText('Opt 2')
+  await expect(opt3).toHaveText('Opt 3')
+
+  // Confirm Selected Value - Opt 3
+  await expect(opt1).toHaveClass(/bg-white/)
+  await expect(opt2).toHaveClass(/bg-white/)
+  await expect(opt3).toHaveClass(/bg-gray-9/)
+
+  // Confirm Click Changes Selected to Opt 2
+  const opt2Selected = waitForCustomEvent(page,'input')
```
ehhong, line 30: btw, i just made a helper function for switch that tests the event dispatch and also the value dispatched, if you want to add that i can try to push that in first

anjinai: Nice!!!!  that would be great, thanks!

tests/radio.spec.ts
anjinai: Good call - all set!

tests/radio.spec.ts
ehhong: https://github.com/viamrobotics/prime/pull/194

tests/radio.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+  await expect(opt4).toHaveClass(/bg-white/)
+  await expect(opt5).toHaveClass(/bg-gray-9/)
+  await expect(opt6).toHaveClass(/bg-white/)
+
+  // Focus on Opt 6 - Keydown Test
+  const opt6Selected = waitForCustomEvent(page,'input')
+  await opt6.focus()
+  // Hit Enter
+  await page.keyboard.press('Enter')
+
```
Kschappacher, line 79: I would add a check to ensure that the 'input' event is emitted like this: await expect(opt6Selected).toBeTruthy()

Kschappacher (approved):
One small comment, otherwise LGTM

tests/radio.spec.ts
anjinai: Thanks, all set!

PR #194, opened 2023-03-10
APP-1387 Add helper function for testing dispatched custom events with parameters

ehhong:
Would love thoughts on if this is acceptably hack-y. My solution for the discrepancy between the event listener being called at time of dispatch and the initial `page.evaluate()` call is to store the parameter value on the page window in the event listener and then retrieve it after the expected dispatch with another `page.evaluate()`

Talked with @Kschappacher briefly earlier, would love to hear ideas for any other solutions


tests/lib/helper.ts
```diff
@@ -1,15 +1,29 @@
 import { expect,Page } from '@playwright/test';
 
 export const waitForCustomEvent = async(page:Page, customEventName:string) => {
-    return page.evaluate(eventName => new Promise(callback => window.addEventListener(eventName, callback,{ once: true })), customEventName)
+  return page.evaluate(eventName => new Promise(callback => window.addEventListener(eventName, callback,{ once: true })), customEventName)
```
ehhong, line 4: just changed tabs to 2 spaces instead of 4

tests/lib/helper.ts
```diff
@@ -1,15 +1,29 @@
...
+    new Promise((resolve) => {setTimeout(resolve, 500, "timeout")}),
+  ])
 
-    expect(value).toBe("timeout")
+  expect(value).toBe("timeout")
 }
 
+// This function adds an event listener for eventName and puts the value of paramName on the window
+// Retrieve the value of paramName by calling await page.evaluate('window.<paramName>')
+export const waitForCustomEventWithParam = (page:Page, eventName:string, paramName:string) => {
```
Kschappacher, line 18: Are there any concerns with placing the value on the window causing issues between tests? is there any clean up we could do of the value afterwards so another test doesnt pick it up?

ehhong: I think if the tests are broken down into one event of each type per test, we shouldn't have any problems (every `page.goto()` should create a new page environment), though I should test this function more. I imagine this would cause a problem if one test dispatches for example, two "input" events with the same value.

If that's causing a problem, maybe we can also create a helper function `getCustomEventValue` to retrieve the event value (`page.evaluate('window.eventName')`) and also clear it from the window?

micheal-parks (approved):
pending change discussed in slack :)

PR #195, opened 2023-03-10
App 1352: Adding component tests for monaco-editor

RoxyFarhad:
I had some issues making tests for the monaco editor because monaco does a lot of things under the hood which are hard to determine via CSS.
1. Syntax highlighting - through looking at CSS it's clear that monaco has different styles that it renders for group of text depending on its language & whether its correct syntax. It's unclear what classes match to what combinations for these, so for now I avoided creating tests for syntax highlighting. 
2. Syntax highlighting for incorrect syntax - for the same reason ^ its difficult to know whether the syntax is correct via the css classes for each line in the editor. Looking at the /code-editor-test.html, we can manually inspect whether the syntax is highlighted correctly. I maninly added separate tests for invalid syntax to check that they were still being rendered. 
3. Schema validation - seems to be correctly working, but again hard to determine if the correct things are being highlighted via CSS. 
4. dark-mode -- you cannot have multiple editors with different themes on the same page, so the tests work correctly but if you load /code-editor-test.html, all the editors will be dark.

### DIFF
--- a/playground/src/code-editor-test.vue
+++ b/playground/src/code-editor-test.vue
@@ -1,6 +1,58 @@
 <script setup lang="ts">
-const javascriptString = "function getComponents(){" + "\n\n" + "}"
-const typescriptString = "function get(value: string) {" + "\n\n" + "}"
+import { json } from 'body-parser';
+import { diff } from 'semver';
+
+const javascriptString = `function getComponents(){
+
+}`
+const typescriptString = `function get(value: string) {
+
+}`
+
+const minimapString = `{
+    "test": [], 
+    "test1": [], 
+    "test2": [], 
+    "test3": [], 
+    "test4": [], 
+    "test5": []
+}
+`
+
+const diffString = `{
+    "test": [], 
+    "test1": [], 
+    "test2": [], 
+    "test3": [], 
+    "test4": [], 
+}
+`
+
+
+const schema = JSON.stringify({
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  $ref: '#/definitions/AttrConfig',
+  definitions: {
+    AttrConfig: {
+      required: ['token', 'host'],
+      properties: {
+        hello: {
+            type: 'integer', 
+            description: 'integer'
+        }
+      },
+      additionalProperties: false,
+      type: 'object',
+    },
+  },
+});
+
+const basicJson = `
+{
+  "hello": "world"
+}
+`
+
 </script>
 
 <template>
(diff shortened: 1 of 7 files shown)


playground/src/code-editor-test.vue
```diff
@@ -0,0 +1,194 @@
...
+        <v-code-editor
+          data-testid='code-editor-golang-valid'
+          value="func (s *Custom) help(x int)"
+          language='go'
+        />
+      </div>
+      <div style='width: 400px; height: 200px;'>
+        <v-code-editor
+          data-testid='code-editor-golang-invalid'
+          value="def help(x int)"
```
RoxyFarhad, line 126: this didn't actually render anything syntactically wrong on the html

playground/src/code-editor-test.vue
```diff
@@ -0,0 +1,194 @@
...
+        <v-code-editor
+          data-testid='code-editor-python-valid'
+          value="def help()"
+          language='python'
+        />
+      </div>
+      <div style='width: 400px; height: 200px;'>
+        <v-code-editor
+          data-testid='code-editor-python-invalid'
+          value="const x = () => {}"
```
RoxyFarhad, line 112: this didn't actually render anything syntactically wrong on the html

playground/src/code-editor-test.vue
```diff
@@ -0,0 +1,194 @@
...
+        <v-code-editor
+          data-testid='code-editor-shell-valid'
+          value="echo test"
+          language='shell'
+        />
+      </div>
+      <div style='width: 400px; height: 200px;'>
+        <v-code-editor
+          data-testid='code-editor-shell-invalid'
+          value="help me!!"
```
RoxyFarhad, line 140: this didn't actually render anything syntactically wrong on the html

tests/code-editor.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+
+test('Renders JSON code editor that is invalid', async ({page}) => {
+    await page.goto('/code-editor-test.html');
+    const validJSONEditor = page.getByTestId('code-editor-json-invalid')
+    await expect(validJSONEditor).toHaveCount(1)
+    await expect(validJSONEditor).toHaveAttribute('language', 'json')
+    await expect(validJSONEditor).toHaveAttribute('value', '{"component"}')
+    const monacoContainer = validJSONEditor.locator('div').first()
+    await expect(monacoContainer).toHaveCount(1);
+    await expect(monacoContainer).toHaveAttribute('data-mode-id', 'json')
```
Kschappacher, line 26: I would add a check to ensure that the invalid portion of the "value" attribute value should be highlighted as an error. Not sure if there is an attribute you can check for this but I recommend loading the code-editor-test.vue file in a browser and inspecting the invalid code editor element to seesee

Kschappacher: I would add this to the other invalid cases as well

tests/code-editor.spec.ts
```diff
@@ -0,0 +1,248 @@
+import { test, expect } from '@playwright/test';
+
+test('Renders JSON code editor correctly', async ({ page }) => {
+  await page.goto('/code-editor-test.html');
+  const validJSONEditor = page.getByTestId('code-editor-json-valid')
+  await expect(validJSONEditor).toHaveCount(1)
+  await expect(validJSONEditor).toHaveAttribute('language', 'json')
+  await expect(validJSONEditor).toHaveAttribute('value', '{"component": "JSON"}')
+  const monacoContainer = validJSONEditor.locator('div').first()
+  await expect(monacoContainer).toHaveCount(1);
```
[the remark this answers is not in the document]
Kschappacher: And anywhere else toHaveCount is being used

Kschappacher (commented):
Left some comments, let me know if you have any questions!

playground/src/code-editor-test.vue
Kschappacher: I would check with @micheal-parks on this, it looks like he built the prime component.

tests/code-editor.spec.ts
RoxyFarhad: yeah, so i added a list of some of the issues with testing monaco on the general comment but checking for syntactical errors is quite difficult to do from a testing standpoint as monaco doesn't add human readable css classnames or attributes to pieces of text that they have highlighted as an error. 

I spent some time looking at this and couldn't figure out how to make a meaningful test for this, so decided that the only real thing that we could test was that the text was actually rendered even if its not syntactically correct.

tests/code-editor.spec.ts
```diff
@@ -0,0 +1,248 @@
+import { test, expect } from '@playwright/test';
+
+test('Renders JSON code editor correctly', async ({ page }) => {
+  await page.goto('/code-editor-test.html');
+  const validJSONEditor = page.getByTestId('code-editor-json-valid')
+  await expect(validJSONEditor).toHaveCount(1)
```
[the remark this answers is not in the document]
RoxyFarhad: So actually the visibility check doesn't work here because the div has width 0 x height 0, but i just instead removed these checks because they were mainly to make sure that the container existed but if the attribute check passes then it implicitly states that it does.

tests/code-editor.spec.ts
Kschappacher: One thing you could add is a check for the text color styling. It looks like the class mtk8 is added to any text that is styled (See screenshot). This could allow you to check for this class in cases where the text is valid and ensure the class is not applied when the text is not valid. 
image

tests/code-editor.spec.ts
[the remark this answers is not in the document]
Kschappacher: 👍 Makes sense

playground/src/code-editor-test.vue
DTCurrie: More than likely we are only doing validation for JSON based on the workers we are including. We may need to include further validation logic from monaco, or remove validation checks for non-JSON editors.

RoxyFarhad: so  i see that we define language like this: 
`export let language: MonacoSupportedLanguages;`
where: 
```
export type MonacoSupportedLanguages =
| 'json'
| 'javascript'
| 'typescript'
| 'python'
| 'go'
| 'shell';
```

so could be something to do with the way that we pass in the go language? I do see on the FE when I manually write bad golang in the code samples tab there is also no invalid syntax highlighting

tests/code-editor.spec.ts
RoxyFarhad: I added a check for valid language syntax highlighting from what i could see for the highlighting on the code-editor-test page. I left out most of the invalid highlighting because most of them are done via underlines and its unclear what class is actually setting that.

Kschappacher (approved):
LGTM

PR #197, opened 2023-03-13
APP-1393: Adding Select tests

RoxyFarhad:


tests/select.spec.ts
```diff
@@ -1,20 +1,204 @@
...
-  const input = page.getByPlaceholder(/select a bobbins/i).last()
+  // don't expect the options to be rendered until a click on input
+  const optionsContainer = select.locator('.options-container').first()
+  await expect(optionsContainer).not.toBeVisible()
 
-  input.focus()
-  input.type('bogs')
+  // click on the input
+  const input = select.locator('input').first()
+  await input.focus()
```
ehhong, line 18: should there be separate tests for display on focus vs. click?

tests/select.spec.ts
```diff
@@ -1,20 +1,204 @@
...
+import { waitForCustomEvent } from './lib/helper.ts';
 
-test('user is able to make a selection with mouse', async ({ page }) => {
-  await page.goto('/test.html');
+test.beforeEach(async ({page}) => {
+  await page.goto('/select-test.html')
+})
 
-  await expect(page.getByLabel(/bobbins/i)).toBeVisible()
+test('Given an options attribute, on select the options should be visible', async ({ page }) => {
```
ehhong, line 8: can we split this test up? feels like 3 separate tests to me: options initially not visible, options visible on focus/click, click on option works properly

tests/select.spec.ts
```diff
@@ -1,20 +1,204 @@
...
+
+  // text is not visible
+  await expect(tooltip.getByText('default tip').first()).not.toBeVisible()
+  // make
+  await tooltip.hover()
+  await expect(tooltip.getByText('default tip').first()).toBeVisible()
+
+})
+
+test('Given info state and tooltip value, there should be an info tooltip next to the select label', async ({page}) => {
```
ehhong, line 108: not sure if this is necessary, but should we test the info state here? maybe checking for the right class or color or something.

tests/select.spec.ts
```diff
@@ -1,20 +1,204 @@
...
+  const input = select.locator('input').first()
+  await input.focus()
+  const optionsContainer = select.locator('.options-container').first()
+  expect(await optionsContainer.locator('label').all()).toHaveLength(3)
+
+  await input.type('one')
+
+  expect(await optionsContainer.locator('label').all()).toHaveLength(2)
+  await expect(optionsContainer.locator('label').first()).toHaveText('one')
+  await expect(optionsContainer.locator('label').last()).toHaveText('one chicken')
```
ehhong, line 180: 🐓

ehhong (commented):
looks good! just a few minor comments ~

tests/select.spec.ts
```diff
@@ -1,20 +1,204 @@
...
+
+  // text is not visible
+  await expect(tooltip.getByText('error tip').first()).not.toBeVisible()
+  // make
+  await tooltip.hover()
+  await expect(tooltip.getByText('error tip').first()).toBeVisible()
+
+})
+
+test('When prefix is set to True', async ({page}) => {
```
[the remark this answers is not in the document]
RoxyFarhad: Yeah -- so i am getting thsi really weird error in the console:

```
runtime-core.esm-bundler.js:40 [Vue warn]: Failed setting prop "prefix" on <v-select>: value true is invalid. TypeError: Cannot set property prefix of #<Element> which has only a getter
    at patchDOMProp (runtime-dom.esm-bundler.js:266:12)
    at patchProp (runtime-dom.esm-bundler.js:384:9)
    at mountElement (runtime-core.esm-bundler.js:5204:21)
    at processElement (runtime-core.esm-bundler.js:5178:13)
    at patch (runtime-core.esm-bundler.js:5098:21)
    at mountChildren (runtime-core.esm-bundler.js:5285:13)
    at mountElement (runtime-core.esm-bundler.js:5195:13)
    at processElement (runtime-core.esm-bundler.js:5178:13)
    at patch (runtime-core.esm-bundler.js:5098:21)
    at mountChildren (runtime-core.esm-bundler.js:5285:13) 
  at <SelectTest>
```

If i change the keyword prefix to something else, it fixes the issue -- i checked and it doesn't seem like we are using prefix anywhere in app at least for v-select, but not sure why this is happening?

tests/select.spec.ts
RoxyFarhad: so click is tested further down when you click on an option

tests/select.spec.ts
RoxyFarhad: The reason i didn't is because separating out tests would just duplicate boilerplate setup code and because this test essentially tests the default expected behavior of the component it made sense to just put it together. In general, I am anti separating tests when its duplicating setup stuff because in the L-R it slows down CI but can separate it out if there is a strong opinion to do so.

tests/select.spec.ts
ehhong: we should also test the initial click to display options though

RoxyFarhad: added!

tests/select.spec.ts
```diff
@@ -1,21 +1,209 @@
...
+  expect(await optionsContainer.locator('label').all()).toHaveLength(3)
 
-  await expect(option).toBeVisible()
+  await expect(optionsContainer).toHaveText(/one/)
+  await expect(optionsContainer).toHaveText(/two/)
+  await expect(optionsContainer).toHaveText(/three/)
 
-  option.click()
+  // selecting a value should emit an input event + render that as the input
+  const oneSelected = waitForCustomEvent(page, 'input')
```
ehhong, line 29: i juust merged in a helper function to also test for the any dispatched values -- would be good to update this to check for the value passed with the dispatched event

ehhong (approved):
pending change to check dispatched event params -- lgtm!

PR #204, opened 2023-03-17
APP-1427 Add input e2e tests

ehhong:
**updated notes**
- updated input component to:
  - only take in characters that are number related given type number (`[0-9]`, `.`, `+`, `-`, `e`)
  - only dispatch input events on new and valid number values
  - display error state on blur when number value is invalid
- fixed bugs with inconsistent decimal points and enforced step in slider
---
**notes**
- updated input component to only take in characters that are number related given type number (`[0-9]`, `.`, `+`, `-`, `e`)
- noticed some current weird behaviors with the component:
  - max / min parameters only apply with use of the slider
  - the step parameter only determines the interval changed by the up/down arrows or slider (doesn't enforce "legal values")
  - ~~inconsistent number of decimal points, unclear what the default / expected behavior is~~ **fixed**
    - ~~on a standard number input component with step=1 and no starting value, hitting the up arrow goes to "1", then incrementing again goes to "2.0" and maintains 1 decimal point thereafter~~
  - ~~using the slider with `type=integer` introduces decimal points~~ **fixed**
- should we add some cleaning / verification that the values sent can always be cast to numbers? 

@micheal-parks could you let me know if these follow the desired behavior?

Behavior | Scenario
-- | --
Text | GIVEN a "type" attribute of "text" has been applied to a v-input elementWHEN the element is renderedTHEN it should render as a text input
Default | GIVEN no "type" attribute has been applied to a v-input elementWHEN the element is renderedTHEN it should render as a text input
Text: Value | GIVEN a "type" attribute of "text" has been applied to a v-input elementAND a "value" attribute with some text of your choosing has been appliedWHEN the element is renderedTHEN it should render as a text inputAND the starting value of the input should reflect the value text
Required | GIVEN a "required" attribute of "true" has been applied to a v-input elementAND a "label" attribute of some text of your choosing has been appliedWHEN the element is renderedTHEN a * should render next to the label text
Label: Default | GIVEN a "label" attribute of some text of your choosing has been appliedWHEN the element is renderedTHEN a the label should render the label text above the input
Label: Top | GIVEN a "label" attribute of some text of your choosing has been appliedAND a "labelposition" attribute of "top" has been specifiedWHEN the element is renderedTHEN a the label should render the label text above the input
Label: Left | GIVEN a "label" attribute of some text of your choosing has been appliedAND a "labelposition" attribute of "left" has been specifiedWHEN the element is renderedTHEN a the label should render the label text to the left of the input
Number | GIVEN a "type" attribute of "number" has been applied to a v-input elementAND a "step" attribute is specifiedWHEN the element is renderedTHEN it should render as a number inputAND the input should increment/decrement by the specified step value
Number: Value | GIVEN a "type" attribute of "number" has been applied to a v-input elementAND a "value" attribute with a number of your choosing has been appliedWHEN the element is renderedTHEN it should render as a number inputAND the starting value of the input should reflect the value attribute
Number Default Step | GIVEN a "type" attribute of "number" has been applied to a v-input elementAND no "step" attribute is specifiedWHEN the element is renderedTHEN the step should default to 1
Placeholder | GIVEN a "placeholder" attribute has been applied to a v-input elementWHEN the element is renderedTHEN the placeholder attribute value should render as placeholder text within the v-input element
Date | GIVEN a "type" attribute of "date" has been applied to a v-input elementWHEN the element is renderedTHEN it should render as a date input
Time | GIVEN a "type" attribute of "time" has been applied to a v-input elementWHEN the element is renderedTHEN it should render as a time input
Datetime Local | GIVEN a "type" attribute of "datetime-local" has been applied to a v-input elementWHEN the element is renderedTHEN it should render as a datetime-local input
Email | GIVEN a "type" attribute of "email" has been applied to a v-input elementWHEN the element is renderedTHEN it should render as an email input
No Label | GIVEN no "label" attribute has been applied to a v-input elementWHEN the element is renderedTHEN no text should render above the input
Tooltip: default | GIVEN a "tooltip" attribute has been applied to the v-input elementAND no "state" attribute has been applied to the input elementWHEN the element is renderedTHEN an info icon should render above the input
Tooltip: info state | GIVEN a "tooltip" attribute has been applied to the v-input elementAND a "state" attribute of "info" has been applied to the input elementWHEN the element is renderedTHEN an info icon should render above the input
Tooltip: warn state | GIVEN a "tooltip" attribute has been applied to the v-input elementAND a "state" attribute of "warn" has been applied to the input elementWHEN the element is renderedTHEN an warn icon should render above the input
Tooltip: error state | GIVEN a "tooltip" attribute has been applied to the v-input elementAND a "state" attribute of "error" has been applied to the input elementWHEN the element is renderedTHEN an error icon should render above the input
Readonly | "GIVEN a "readonly" attribute has been applied to the v-input elementAND the "readonly" attribute is "readonly"WHEN the element is renderedTHEN the input should render as readonly"
Disabled | "GIVEN a "disabled" attribute has been applied to the v-input elementAND the "disabled" attribute is "true"WHEN the element is renderedTHEN the input should render as disabled"
Message | GIVEN a "message" attribute has been applied to the v-input elementWHEN the element is renderedTHEN the message text should display below the input
Message: Error State | GIVEN a "message" attribute has been applied to the v-input elementAND a "state" attribute of "error" has been applied to the v-input elementWHEN the element is renderedTHEN the message text should display below the input in red
On: Input | GIVEN a v-input elementWHEN the element is renderedWHEN an input is updatedTHEN an event should fire


src/elements/input/input.svelte
```diff
@@ -95,12 +95,23 @@ const handleInput = () => {
     return;
   }
 
-  if (type === 'number' && input.value.endsWith('.')) {
-    return;
+  if (type === 'number') {
```
ehhong, line 98: @micheal-parks would appreciate your eyes on this, not sure if this was the right approach

tests/input.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+  await expect(inputLabelLeft.getByText('D:')).toBeVisible()
+  await expect(await inputLabelLeft.locator("input:right-of(:text('D:'))").first()).toBeVisible()
+});
+
+test('Given type number, initializes with given number value', async ({ page }) => {
+  const inputNumber = await page.getByTestId('input-number')
+  await expect(inputNumber).toBeVisible()
+  await expect(inputNumber.locator('input').first()).toHaveValue('3.14159')
+});
+
```
ehhong, line 63: did not include testing for type=`integer`, it's not included in prime stories but i saw it in the code. should i add?

tests/input.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+  await expect(inputMessage.locator('input:above(:text("it\'s the cliiimb"))')).toBeVisible()
+});
+
+test('With error state, displays message below input box in red', async ({ page }) => {
+  const inputMessageError = page.getByTestId("input-message-error")
+  await expect(inputMessageError.getByText('fell off the cliiimb')).toHaveClass(/text-red-600/)
+  await expect(inputMessageError.locator('input:above(:text("fell off the cliiimb"))')).toBeVisible()
+});
+
+test('Fires input event with value on input', async ({ page }) => {
```
ehhong, line 204: for type=number, new input cleaning does not fire an event for every keystroke but rather only on valid formatted numbers, will that be a problem?

src/elements/input/input.svelte
RoxyFarhad: where do we set type? should we not use some typescript validation instead of string checking?

src/elements/input/input.svelte
```diff
@@ -95,12 +95,23 @@ const handleInput = () => {
...
+    let numString = '';
+    for (const c of input.value) {
+      if (/[\d+.e-]/.test(c)) {
+        numString += c;
+      }
+    }
+    input.value = value = numString;
+
+    // only send value if formatted as valid number
+    if (Number.isNaN(Number.parseFloat(value))) {
```
RoxyFarhad, line 109: have you checked this doesn't throw an exception on weird input types?

src/elements/input/input.svelte
ehhong: this is a parameter to the component called "type" that determines the type of the HTML input element. so, the variable `type` holds a string value of either `"text"`, `"number"`, or `"input"`

src/elements/input/input.svelte
ehhong: the weirdest thing about this function is that it'll greedily parse the float for a valid number. so on input `"1..."` returns as valid, and it'll parse `1`. similarly, `1...1` will parse as `1`. `1.1.1` parses as `1.1`

[`Number.parseFloat()` returns `NaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseFloat) if the string passed in is not a number at all (ex. "...." or "+" etc)

we should probably expand this validation for `type=integer`

ehhong: oh actually, looks like we can use `Number()` if we want this check to be stricter
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat#description

tests/input.spec.ts
ehhong: added

src/elements/input/input.svelte
```diff
@@ -95,19 +95,30 @@ const handleInput = () => {
...
-
-  value = input.value;
-
   internals.setFormValue(value);
 
   dispatch('input', { value });
 };
 
 const getDecimals = (value = '') => {
-  return Math.max(value.split('.').pop()?.length ?? 0, stepDecimalDigits);
```
ehhong, line 110: there was a bug where if value was a string representing an integer `"2"`, `value.split('.')` was `["2"]`, and then "value.split('.').pop()" was `"2"` which is a string of length `1`

src/elements/input/input.svelte
```diff
@@ -138,8 +149,7 @@ const handleNumberDragMove = (event: PointerEvent) => {
   const x = event.clientX;
   const deltaString = (-(startX - x) * stepNumber / 10).toFixed(type === 'integer' ? 0 : stepDecimalDigits);
   const delta = type === 'integer' ? Number.parseInt(deltaString, 10) : Number.parseFloat(deltaString);
-
-  value = input.value = (startValue + delta).toFixed(getDecimals(input.value));
+  value = input.value = (startValue + (delta * stepNumber)).toFixed(getDecimals(input.value));
```
ehhong, line 152: modified to `delta * stepNumber` to account for the step

src/elements/input/input.svelte
micheal-parks: Nice catch!

tests/input.spec.ts
micheal-parks: Ah yep, we should add type=integer to the stories as well. Not a blocker though

src/elements/input/input.svelte
ehhong: it depends on the check in [line 128](https://github.com/viamrobotics/prime/pull/204/files#diff-86dbb74c12f51262a7099a9487197a4c34301a0844e630b15d297ca2a0ebbe76R128 )

the current code would fire an event with value `1.`, but wanted to check in with you to see what the desired event firing behavior is. for example, on input string `1.`, should no event be fired or should an event fire with value `1`?

src/elements/input/input.svelte
```diff
@@ -95,19 +95,49 @@ const handleInput = () => {
...
+  if (type === 'number') {
+    // ensure input of chars that form valid numbers
+    let numString = '';
+
+    let pointCount = 0;
+    let minusCount = 0;
+    let plusCount = 0;
+    let eCount = 0;
+
+    for (const c of input.value) {
```
ehhong, line 107: this is throwing a lint error for cognitive complexity
```Refactor this function to reduce its Cognitive Complexity from 19 to the 18 allowed  sonarjs/cognitive-complexity```
but I'm struggling to find a way to make it simpler... would appreciate any ideas

src/elements/input/input.svelte
RoxyFarhad: I would argue here that we want to set 1. -> 1.0 or 1 the former maybe seems more logical?

src/elements/input/input.svelte
RoxyFarhad: what is the goal of this function? To make sure the string is a valid float or a number? Should maybe ad the goal of the loop here as well as a comment

src/elements/input/input.svelte
ehhong: can I directly compare against the previous value somehow to prevent an eager event fire?

tests/input.spec.ts
micheal-parks: that sounds fine

src/elements/input/input.svelte
ehhong: decided to use `Number()` to align user experience

DTCurrie (approved):
✅ pending the conversation around trailing `.`

src/elements/input/input.svelte
ehhong: @DTCurrie we decided that `1.` will evaluate to `1`, but the event will only be sent if the value is different from the previous.
ex. 
1. `input.value`: `1` --> send value `1`
2. `input.value`: `1.` --> no value sent (repeated)

alternatively,
1. `input.value`: `1.2` --> send value `1.2`
2. `input.value`: `1.` --> send value `1`

PR #205, opened 2023-03-17
APP-1386: Add tooltip e2e tests

anjinai:
**20230320_APP-1386: Add tooltip e2e tests**

Behavior | Scenario
-- | --
Hover | GIVEN a text attribute has been applied to a v-tooltip elementWHEN the element is renderedAND the element is hoveredTHEN the text attribute value should render
Top | GIVEN a location attribute of "top" has been applied to a v-tooltip elementWHEN the element is renderedAND the element is hoveredTHEN the text attribute value should render above the element
Bottom | GIVEN a location attribute of "bottom" has been applied to a v-tooltip elementWHEN the element is renderedAND the element is hoveredTHEN the text attribute value should render below the element
Left | GIVEN a location attribute of "left" has been applied to a v-tooltip elementWHEN the element is renderedAND the element is hoveredTHEN the text attribute value should render to the left of the element
Right | GIVEN a location attribute of "bottom" has been applied to a v-tooltip elementWHEN the element is renderedAND the element is hoveredTHEN the text attribute value should render above the element
Default | GIVEN no location attribute has been applied to a v-tooltip elementWHEN the element is renderedAND the element is hoveredTHEN the text attribute value should render above the element

<p></p>


tests/tooltip.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+test('Confirms tooltip text is visible upon hover', async ({ page }) => {
+  const tooltip = page.getByRole('tooltip', { name: 'This is the hover test'})
+  await expect(tooltip).not.toBeVisible()
+  const wordToHover = page.getByText('Hover Test').first()
+  await wordToHover.hover();
+  await expect(tooltip).toBeVisible()
+});
+
+test('Confirms tooltip text is above text to hover if location attribute is specified as "top"', async ({ page }) => {
```
anjinai, line 15: This hover test passes playwright tests in CI and playwright webkit tests locally, but does not pass chromium tests (in headless mode) locally.  I've tried many different things to try to get this to pass (using different locators and selectors for the tooltip, using waitforselector, adding a delay, clicking on the hover text after hover and before hover, looking at screenshots at different points of this test, looking at the video for the test).  **Do you happen to know why something would pass webkit, but not chromium in headless mode?**

tests/tooltip.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  const textLocatedBelow = await page.locator('p:near(:text("This is the top test"))').textContent()
+  expect(textLocatedBelow).toBe('Top Test')
+});
+
+test('Confirms tooltip text is below text to hover if location attribute is specified as "bottom"', async ({ page }) => {
```
anjinai, line 20: Additionally, the ticket calls for tests that confirm the position of the tooltip is to the top, bottom, left, and right of the hover text depending on the specified location attribute in the v-tooltip.  However, location based selectors weren't working for me.  Do you know of another way to test for location besides using location based selectors like (await page.locator('p:above(:text("This is the right test"))').textContent()).

ehhong: not optimal, but are there particular classes on inner elements that determine their relative positions?

tests/tooltip.spec.ts
ehhong: can confirm these many valiant efforts

micheal-parks: Have you tried accessibility based selectors such as getting by role=tooltip?

tests/tooltip.spec.ts
micheal-parks: I haven't tried this yet with playwright but my first intuition for a test like this would be to compare the viewport offset of the tooltip against its target with an api like `getBoundingClientRect()`. Not sure if it's available in this environment though.

anjinai: That totally worked!!! Thank you both!

tests/tooltip.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+test('Confirms tooltip text is above text to hover if location attribute is specified as "top"', async ({ page }) => {
+  const text = page.locator('p:near(:text("This is the top test"))')
+  const tooltip = page.getByRole('tooltip', { name: 'This is the top test'})
+  const wordToHover = page.getByText('Top Test').first()
+
+  const textLocation = await text.evaluate(element => element.getBoundingClientRect());
+  wordToHover.hover()
+  const tooltipLocation = await tooltip.evaluate(element => element.getBoundingClientRect());
+  
+  expect(tooltipLocation.y).toBeLessThan(textLocation.y)
```
anjinai, line 24: Just a heads up that some properties provided by getBoundingClientRect (x, y, top, bottom, left, right values) are relative to the top-left of the viewport.  Normally, you would think that tooltip location y would be greater than the hover text location y when you're expecting the tooltip to be above the hover text.  But in this case, it's the opposite.

tests/tooltip.spec.ts
anjinai: Sadly, I have.  I'm just going to skip this test for now so I can get this PR in.  As of now, the hover test passes in CI and webkit locally.  Weirdly, when I conditionally skip the test when the browser is chromium, the webkit then fails locally.  Something wonky is going on.

ehhong (approved):
lgtm!

PR #207, opened 2023-03-22
Add prettier to autoformat source, docs, and config

mcous:
## Overview

After a _very_ scientific poll in Slack with four 👍's and no 👎's, this PR adds [prettier](https://prettier.io) to this repository. I attempted to follow the config found in the `eslint` settings, the source itself, and [viamrobotics/viam-typescript-sdk](https://github.com/viamrobotics/viam-typescript-sdk).

I also added some brief usage instructions for development tasks/scripts to `CONTRIBUTING.md`.

## Change log

This PR is split into ~three~ various commits to aid reviewing:

1. The config changes themselves
2. Running `npm run format`
3. A single file rename I found where a test file did not match existing conventions of lowercase file names
4. ...fixups and review requests

If helpful, I can split these changes into multiple PRs.

## Test plan

Given automated tests in place (though they're a bit flaky on my machine; requiring increased timeouts for the code editor tests), I think giving Storybook a quick visual comparison to prod should be pretty good here.

You may notice that the browser table in the README is broken in Storybook. This an existing issue on the site due to Storybook v7 dropping GFM by default, and not caused by this PR.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,6 +4,7 @@
     "csstools.postcss",
     "bradlc.vscode-tailwindcss",
     "svelte.svelte-vscode",
-    "unifiedjs.vscode-mdx"
+    "unifiedjs.vscode-mdx",
+    "esbenp.prettier-vscode"
   ]
 }


.vscode/extensions.json
```diff
@@ -4,6 +4,6 @@
     "csstools.postcss",
     "bradlc.vscode-tailwindcss",
     "svelte.svelte-vscode",
-    "silvenon.mdx"
+    "unifiedjs.vscode-mdx"
```
micheal-parks, line 7: Ah, I assume `silvenon.mdx` is now deprecated?

mcous: Yup, this is the new one

CONTRIBUTING.md
```diff
@@ -35,3 +35,41 @@ More detailed docs exist on [the storybook](https://www.viam.com/prime).
 Linked below are some articles that provide knowledge for how to best build reusable, generic web components.
 
 - [Custom Elements Best Practices](https://web.dev/custom-elements-best-practices/)
+
+## Development tasks
```
micheal-parks, line 39: 🎉

src/elements/collapse.svelte
```diff
@@ -29,45 +28,47 @@ const handleClick = (event: Event) => {
   isOpen = !isOpen;
   dispatch('toggle', { isOpen });
 };
-
 </script>
 
-<div class='relative w-full'>
+<div class="relative w-full">
```
micheal-parks, line 33: Suggestion: Since we're interspersing JS + HTML here could we have prettier using single quotes for everything? I'd rather not have to mentally switch between the two in the same file.

src/elements/input/input.svelte
```diff
@@ -197,33 +213,48 @@ const handleNumberDragDown = async (event: PointerEvent) => {
   window.addEventListener('pointermove', handleNumberDragMove);
   window.addEventListener('pointerup', handleNumberDragUp, { once: true });
 };
-
 </script>
 
+<style>
```
micheal-parks, line 218: Oof, I don't like this one 😭 , because it's inconsistent with the svelte docs / convention of putting style last. I know prettier is opinionated, but is there any way to change?

mcous: Oh sure, the default for prettier-plugin-svelte is, in fact, to put style last. I only changed it since I saw style in a different place in various files so thought it might be Viam convention.

I'll remove the override which will re-order `style` to be last.

src/elements/collapse.svelte
mcous: Definitely agree, and attempted to do so by setting `jsxSingleQuote` to `true`, but ran right into https://github.com/sveltejs/prettier-plugin-svelte/issues/253.

Options, depending on how much we care about this:

1. Live with it until issue is resolved
2. Don't format `.svelte` files until issue is resolved
3. Ditch the PR 😭 

In my personal experience working in projects that used `'` in JS and `"` in JSX, I stopped thinking about it quite quickly once "format on save" was turned on, but YMMV

src/elements/input/input.svelte
DTCurrie: Yeah I would like to follow svelte convention here.

src/elements/collapse.svelte
micheal-parks: I'm cool with 1., my only rec would be to link https://github.com/sveltejs/prettier-plugin-svelte/pull/271 somewhere in the eslint config so that we don't forget to track

micheal-parks (approved):
Cool! Are you planning on replicating on APP and RDK next?

PR #212, opened 2023-03-24
Fix flakiness in slider tests from comparing floats

ehhong:
~~also created a new helper function for converting strings to numbers~~ decided as redundant

@DTCurrie I can wait until you push your SPA testbed to merge this

### DIFF
--- a/tests/table.spec.ts
+++ b/tests/table.spec.ts
@@ -1,5 +1,5 @@
 import { test, expect } from '@playwright/test';
-import { hexToRGB } from './lib/helper.ts';
+import { hexToRGB } from './lib/helper.js';
 
 test.beforeEach(async ({ page }) => {
   await page.goto('/table-test.html');


tests/lib/helper.ts
```diff
@@ -1,6 +1,12 @@
 import { expect, Page } from '@playwright/test';
 import config from '../../tailwind.config.cjs';
 
+// Convert from '<number>px' to actual number
+// ex. '3.14px' --> 3.14
+export const pxStringToNumber = (pxString: string) => {
+  return Number(pxString.substring(0, pxString.indexOf('px')));
+};
+
```
mcous, line 9: Do we need a helper for this? I think this is the default behavior of `Number.parseFloat`

```js
> Number('3.14px')
NaN
> Number.parseFloat('3.14px')
3.14
```

ehhong: ah you are totally correct. thank you for catching!

mcous (approved):
Test changes LGTM, logic looks correct to allow fuzziness in the actual pixel value.

My suggestions are just the stuff that #210 is going to highlight, so I don't think it's important whether or not they land in this PR

PR #213, opened 2023-03-24
Fix event waiting helper code and expectations in tests

mcous:
## Overview

While reviewing newly surfaced linting and typechecking errors in #210, I noticed that many of the "wait for an event to be fired" logic in the tests were wrong. The problems boiled down to a few main themes:

1. Asserting that a Promise object is truthy rather than awaiting it

```ts
const someEventPromise = waitForCustomEvent(...)
expect(someEventPromise).toBeTruthy() // always passes, because every `Promise` object is truthy
```

2. `waitForCustomEventTimeout` helper always resolved, either by the event happening or by the timeout, even though in usage it was used to express "this event shouldn't happen"

3. `waitForCustomEventWithParam` resolved immediately, rather than waiting for the event to happen

On top of that, the `waitForCustomEventWithParam` + `getCustomEventParam` pair was:

- Putting internal stuff on `window` to test
- Not really necessary, because `waitForCustomEvent` returned the event

In order to fix these issues, I made the following changes:

- Modified `waitForCustomEvent` to actually returned the `detail` value from the event, which is wasn't doing because `Event` objects don't serialize intuitively
- Added a default timeout failure to `waitForCustomEvent`
- Replaced `waitForCustomEventTimeout` with `waitForCustomEvent` by giving `waitForCustomEvent` a way to say "this event shouldn't happen"
- Removed `waitForCustomEventWithParam` and `getCustomEventParam` in favor of using `waitForCustomEvent` directly
- Removed all incorrect usage of `toBeTruthy` on `Promise` objects

## New usage

To check that an event has happened, the following can be used:

```ts
const eventHandler = waitForCustomEvent(...)

// if we just care that it was fired
await eventHandler.detail()

// if we care about the payload
const detail = await eventHandler.detail()
expect(detail).toEqual(...)
// or
await expect(eventHandler.detail()).resolves.toEqual(...)
```

To check that an event did not occur:

```ts
const eventHandler = waitForCustomEvent(...)

await eventHandler.didNotOccur()
// or
await expect(eventHandler.didNotOccur()).resolves.toBe(true)
```

## Review requests

Happily, most test cases continued to pass after I made this change. Two tests, however, did not. I need help identifying whether these failures are due to incorrect expectations in the tests of some failure of the actual code.

- `input.spec.ts` -> `Given type number, only dispatches valid and new number values`
    - `NaN` is input into the field, and no `input` event was expected
    - However, an `input` event is being fired
    - IIRC `NaN` is a valid number field input in JS, so I _think_ this is a bad test expectation
- `multi-select.spec.ts` -> `Test sort options for container`
    - An `input` event is not firing where the test expects on
    - I don't really know what should be happening here 

## Next steps

I've spun up a scratch pad doc in Confluence for various testing practices, and I'll be incorporating Playwright learnings from this PR into this document. The reality of the situation is that we're going to be adding tests to existing code, which is a pretty difficult thing!

When adding tests to existing code, it can be pretty tough to write assertions that will properly fail, because the code (hopefully) is already working. You lack the signals and progression of a test-first approach where you know your test fails because you wrote the test in a failing state.

One helpful trick in this (necessary) test-after context is to intentionally break the code while you're writing tests -  for example, by commenting out important things - and make sure the tests fail in the way you expect them to. You can also try to introduce obvious bugs into the tests themselves and make sure they fail when written improperly - in this case, for example, by listening for the wrong event type.


tests/input.spec.ts
```diff
@@ -347,34 +342,30 @@ test('With error state, displays message below input box in red', async ({
...
 }) => {
-  const inputNumber = await page.getByTestId('input-number');
-  const input = await inputNumber.locator('input').first();
+  const inputNumber = page.getByTestId('input-number');
+  const input = inputNumber.locator('input').first();
 
-  const noInputEvent1 = await waitForCustomEventTimeout(page, 'input');
-  await input.fill('NaN');
-  await noInputEvent1;
+  // TODO(mc, 2023-03-24): `NaN` is a valid number input in JS
```
mcous, line 348: One test failure noted in PR description. Anyone have any thoughts here?

tests/multi-select.spec.ts
```diff
@@ -277,10 +276,11 @@ test('Test sort options for container', async ({ page }) => {
   await expect(optionsContainer.locator('label').nth(1)).toHaveText('sad');
   await expect(optionsContainer.locator('label').last()).toHaveText('angry');
 
-  // check that an event is fired
-  const event = waitForCustomEventWithParam(page, 'input', 'sa');
+  // TODO(mc, 2023-03-24): an event is not fired, figure out what's going on here
```
mcous, line 279: The other test failure

ehhong:
thanks so much for doing this! wrestled with this for a while, this is so clearly a much better solution

tests/input.spec.ts
ehhong: I realized this is firing an event because the value goes from a valid starting number (set as a parameter in `input-test.vue`) to a blank string (the component rejects the characters "NaN")

I have a fix for this, maybe leave a TODO comment here and I can merge my solution after

ehhong (approved):

anjinai (approved):

tests/lib/helper.ts
```diff
@@ -20,76 +20,82 @@ export const hexToRGB = (color?: string) => {
   })`;
 };
 
+export interface CustomEventPayload {
+  detail: unknown;
+}
+
 export const waitForCustomEvent = async (
```
Kschappacher, line 27: Thanks so much for cleaning all this up! The changes look good. One ask, could we add tests for the helper functions?

mcous: Sure, I'll give it a shot, but...

...`waitForEvent` / `expectNoEvent` are very coupled to the Playwright framework itself - they call `page.evaluate`, which seems to do something complicated to serialize the function over IPC to the browser, run in the browser context, and then serialize something back - so I'm not 100% certain of the value of adding such tests.

Such tests would likely be redundant with the test suite itself, because they would look something like...

1. Use Playwright to trigger an event on a DOM node, probably via `page.evaluate`
2. Use `waitForCustomEvent` to listen for that event

...which is more or less what every test that uses `waitForEvent` does

mcous: Update: the value of writing the tests turned out to be "Teach Michael how Playwright works," so I'll chalk that one up as a win

Kschappacher: That's totally fair, it would be a bit redundant. I think the one case they may be helpful is if someone is trying to change the helper functions. The basic tests would be a good way to outline the intended behavior of the function and the tests wouldn't be coupled to a component test. 

That said totally your call if you want to add them!

Kschappacher (approved):
LGTM

tests/lib/helper.ts
mcous: I agree, the tests are helpful documentation, and Playwright is _quirky_, so they function well as contract tests, too. I'll be adding them

PR #214, opened 2023-03-27
APP-1476: Make drop down select focus on top result

anjinai:
**20230327_Make drop down select focus on top result**

**Items to Note:** 

- Updates select and multi-select components to have the top (ordered) result already in focus such that a user can press enter and select that first item.
- PR includes tests for this behavior as well.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime",
-  "version": "0.1.4",
+  "version": "0.1.5",
```
mcous, line 3: What's the bump/release strategy for this library?

mcous (approved):
Tested behavior manually, and it works as advertised. Tests look good, too!

package.json
anjinai: I think we’re trying to follow semver.  Anytime I make a bug fix/minor change to a component, I increase the patch version in prime and then go to app and update the version of prime that app uses.

anjinai:
> Tested behavior manually, and it works as advertised. Tests look good, too!

Thanks for reviewing!

ehhong (approved):

PR #216, opened 2023-03-27
Fix failing tests revealed by fixing custom event helpers

ehhong:
Fix 2 bugs in input and multi-select tests, see conversation in #213


mcous (approved):
Nice, LGTM!!

PR #217, opened 2023-03-28
App 1133 Create PRIME code snippet element

mrloureed:
https://viam.atlassian.net/browse/APP-1133

Simplified code snippet component that leverages prismjs.


src/elements/code-snippet.svelte
```diff
@@ -0,0 +1,62 @@
...
+    await import(/* @vite-ignore */cdn('prism.min.js'))
+  }
+
+  await import(/* @vite-ignore */cdn(`components/prism-${language}.min.js`))
+
+  Prism.highlightElement(el)
+}
+
+export class VCode extends HTMLElement {
+  observer!: MutationObserver
```
mrloureed, line 25: @micheal-parks I was able to get this to work as a Svelte file rather than Typescript, however I am getting a Svelte parse error for "observer." 

```
(property) VCode.observer: MutationObserver
Unexpected token

If you expect this syntax to work, here are some suggestions: 
If you use typescript with `svelte-preprocess`, did you add `lang="ts"` to your `script` tag? 
Did you setup a `svelte.config.js`? 
See https://github.com/sveltejs/language-tools/tree/master/docs#using-with-preprocessors for more info.svelte(parse-error)
```

src/elements/code-snippet.ts
```diff
@@ -0,0 +1,56 @@
+import pkg from '../../package.json';
```
mrloureed, line 1: This will be going away in lieu of svelte file

PR #222, opened 2023-03-31
App 1133 Create PRIME code snippet element

mrloureed:
Create PRIME code snippet element. This is always read only and offers our normal vs theme and a dark version. 

_Thank you @micheal-parks for your code and guidance on this one!_
https://viam.atlassian.net/browse/APP-1133


src/elements/code-snippet.ts
```diff
@@ -48,15 +50,17 @@ export class VCode extends HTMLElement {
   }
 
   render = () => {
-    const language = this.getAttribute('language') ?? '';
+    language = this.getAttribute('language') ?? '';
```
micheal-parks, line 53: Why are we moving these vars to the outer scope? If there are multiple snippets on the same page we'll have one code snippet's `code` and `language` vars overriding another.

src/elements/code-snippet.svelte
```diff
@@ -0,0 +1,89 @@
...
+    // eslint-disable-next-line require-atomic-updates
+    loadedLanguages[language] = true;
+  }
+
+  if (element !== undefined) {
+    window.Prism.highlightElement(element);
+  }
+};
+
+$: void highlight(element);
```
mrloureed, line 66: Added `void` to fix Typescript error: fyi @micheal-parks 

```
Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.eslint[@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises)
```

micheal-parks (approved):

micheal-parks:
Should wait for cousins since my review is kinda pointless 😂

PR #223, opened 2023-04-06
Update fonts and design sync

micheal-parks:
This PR was done after another design sync with Nat + Jason. Changes:
* Notify element now has an action slot for buttons. This was done for upcoming billing features.
* PRIME now makes no opinions about fonts, this was done to transition us gracefully to Public Sans.
* Input labels are now only first letter capitalized
* Storybook fonts are changed to Public Sans
* Various smaller design syncs


mrloureed (approved):
LGTM

PR #225, opened 2023-04-13
DATA-1365 - Update v-select behavior

alexis-wei:
** NOTE: RELATIVELY URGENT, PLS CHECK ASAP **
Changes:
- added heading in v-select to use in app! It currently exists in multi-select, but not the normal select
- added a 'change' dispatch call to know when a value was actually selected vs just a change in the input bar
- Updated handleEnter behavior, now it will clear the input if it matches none of the items, I know this also caused issues on the config page. 
- updated the behavior to match multi select where it doesn't correctly detect an empty string as options

Video for select behavior: 

https://user-images.githubusercontent.com/29152328/232108609-177b82f5-83b6-4009-ac0c-86c2172a7c62.mov

Would like to make sure this doesn't break current existing functions in app


src/elements/select/select.svelte
```diff
@@ -105,12 +106,14 @@ const handleEnter = () => {
 
     if (result) {
       value = result;
+      dispatch('select', { value });
```
mcous, line 109: To match [vanilla `<select>`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement#events), should this event be `change` instead of `select`?

> When dealing with a vanilla `<select>`, the [following events](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement#events) can be emitted:
>
> - [change](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) event
> Fires when the user selects an option.
> - [input](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) event
> Fires when the value of an [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input), [<select>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select), or [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) element has been changed.

mcous (commented):
I think this PR should include updates to the tests, too

src/elements/select/select.svelte
alexis-wei: that makes sense to match the behavior, will change naming to "change"

src/stories/select.stories.mdx
```diff
@@ -80,6 +85,10 @@ import gpioOptions from './assets/gpio-options';
     'on:search': {
       description: 'Event fired when the search input is changed',
     },
+    'on:select': {
+      description:
+        'Event fired when an option in the list is selected or an input value matches an option on enter key press',
+    },
   }}
 />
```
anjinai, line 94: Can you add a story for with heading?

```
<Canvas>
  <Story
    name='With heading'
    args={{
      label: 'With heading',
      options: 'Option 1, Option 2, Option 3',
      heading: 'Test',
      value: 'Option 3',
    }}
  >
    {({ label, options, value, heading }) => `
      <v-select
        label='${label}'
        options='${options}'
        value='${value}'
        heading='${heading}'
      />
    `}
  </Story>
</Canvas>
```

src/elements/select/select.svelte
```diff
@@ -105,12 +106,14 @@ const handleEnter = () => {
...
       value = result;
+      dispatch('change', { value });
+    } else {
+      value = '';
     }
   }
   if (open) {
     input.blur();
   }
-  dispatch('input', { value });
```
mcous, line 113: Minor point, but should this `input` event remain? Experimenting with a vanilla `<select>`, selecting an option with [enter] fires both events

micheal-parks: there are likely v-selects in app that are listening to this as well

tests/select.spec.ts
```diff
@@ -222,12 +222,15 @@ test('On pressing enter over an option, that option should be selected', async (
   expect(await input.inputValue()).toEqual('');
   // press down over an option
   await page.keyboard.press('ArrowDown');
+  // const valueChanged = waitForCustomEvent(page, 'change');
```
mcous, line 225: Should these lines be commented out? Also, do we want t listen for an `input` event here rather than `change`? (see other comment)

tests/select.spec.ts
```diff
@@ -206,7 +206,7 @@ test('When there is a reduce sort option, the elements should reduce on matching
   );
 });
 
-test('On pressing enter over an option, that option should be selected', async ({
+test('On pressing enter over an option, that option should be selected, change event emitted', async ({
```
mcous, line 209: Looks like this was already the case, but I don't think this title is correct, test appears to be be pressing an arrow, not [enter]`.

If a down arrow is pressed, I think the spec should be only that an `input` event is emitted (because the value changed), but not a `change` event, because the user has not selected the option yet by pressing enter or clicking away. What do you think?

mcous (commented):
Looking good! Just got a couple more questions about the tests

tests/select.spec.ts
alexis-wei: I think we would want to listen to both since the value will be changed after it got selected on enter. Will fix the commented out part though

tests/select.spec.ts
alexis-wei: on the title and existing test, it already has the portion where the user presses down the 'enter' key, where they select the option and the value is changed, shouldn't a `change` event and an `input` event be expected?

src/elements/select/select.svelte
```diff
@@ -100,16 +113,22 @@ const handleKeyUp = (event: KeyboardEvent) => {
 const handleEnter = () => {
   if (navigationIndex > -1) {
     value = sortedOptions[navigationIndex]!;
+    dispatch('change', { value });
   } else {
+    console.log('handle enter pressed');
```
mcous, line 118: Errant console log

tests/select.spec.ts
mcous: My bad, missed the line where enter was pressed

mcous (approved):
LGTM once the console log is removed

src/stories/select.stories.mdx
alexis-wei: done!

sbal13 (approved):
🔥

src/elements/select/select.svelte
```diff
@@ -105,12 +106,14 @@ const handleEnter = () => {
...
       value = result;
+      dispatch('select', { value });
+    } else {
+      value = '';
     }
   }
   if (open) {
     input.blur();
   }
-  dispatch('input', { value });
```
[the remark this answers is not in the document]
alexis-wei: hmmm that's weird, I put it back but it's not showing up here. I did want to leave it so no regressions happen

src/elements/select/select.svelte
```diff
@@ -105,12 +106,14 @@ const handleEnter = () => {
 
     if (result) {
       value = result;
+      dispatch('select', { value });
+    } else {
+      value = '';
```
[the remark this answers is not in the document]
alexis-wei: hmm tested that is-exact would reset it for us so it should be okay to remove

PR #228, opened 2023-04-19
update select

alexis-wei:
updating change behavior so it is also dispatched when the input value matches the option selected


sbal13 (approved):
🔥

PR #232, opened 2023-05-02
fix z-index issue

mrloureed:
Remove relative parent and absolutely positioned child to avoid z-index issue on app.


DTCurrie (approved):
✅

PR #233, opened 2023-05-02
Bugfix notify z index button

mrloureed:
This removes `will-change: auto` CSS from button, as it is causing a layout issue in App, and also:

_The will-change property is intended to be used as a last resort when dealing with known performance problems. Avoid using these utilities too much, or simply in anticipation of performance issues, as it could actually cause the page to be less performant._


micheal-parks (approved):
Just needs `npm run format`

PR #234, opened 2023-05-05
chore: remove redundant vite configuration

mcous:
## Overview

This PR simplifies the Prime development setup. I'm on a background mission to remove Vue from the playground, since it's really just creating busy work for writing tests. Every playground page could be created with vanilla HTML, most by just copy-paste from the Vue templates.

In advance of making that change, I did some build cleanup! We had two mutually exclusive `vite.config.ts` files, one in the root, and one in `playground`.

- `vite.config.ts`
    - Used only for library build
    - Unused `vite dev` script configured in `package.json` that didn't work
- `playground/vite.config.ts`
    - Used only for playground dev server
    - Unused `vite build` script configured in `playground/package.json`

When using library mode with vite, it's typical for `vite dev` to launch a playground, so I consolidated all playground dev server logic into the top-level `vite.config.ts`. This allowed my to remove all build-related config in `playground`.

## Change log

Other than the build changes described above, I made **one runtime change that was required for this to work properly**. The Prime docs state:

> You must host `prime.css` ... at the root directory of your domain

However, if `window.PRIME_CONFIG.base` is not configured, the font loading logic does not respect this requirement. The current code instead loads a URL relative to the current path. 

- Old default: relative to current path: `url(icons.woff2)`
- New default: relative to root: `url(/icons.woff2)`

This PR changes the default runtime behavior to match the documentation, because otherwise the build simplification changes don't work. I also tweaked the URL formation logic so that the old URL behavior of "relative to current path" can be constructed by passing `window.PRIME_CONFIG.base = ''`.

I would consider this a potentially breaking change, so I bumped to `0.2.0` (in prerelease `0.x.x` versions, a "minor" bump is allowed to signify a break according to the semver spec).

## Review Requests

- Make sure the font loading changes don't affect app!


DTCurrie (approved):
🚀

micheal-parks:
This is a great simplification! Especially concerning `PRIME_CONFIG`. One concern: now that the vite configs have been merged, can you confirm that tailwind classes compiled only for the playground are not making their way into the `dist` that we ship to NPM?

mcous:
> One concern: now that the vite configs have been merged, can you confirm that tailwind classes compiled only for the playground are not making their way into the dist that we ship to NPM?

Good call to check this. **The `prime.css` generated from `main` is identical to the `prime.css` generated on this branch**. I also tested by adding a class that does not already exist in the codebase (`aspect-auto`) to see if it showed up in `prime.css`

- Added to random svelte component in `src` - showed up in `prime.css`
- Added to random Vue component in `playground` - did not show up in `prime.css`

micheal-parks:
Nice! Thanks for checking!

PR #237, opened 2023-05-10
Design sync May 10 2023

micheal-parks:
* Align font sizes and padding sizes with figma
* Remove capitalization rules to allow for first letter only capitalized statements
* Add pressed states to buttons
* Add hover and pressed states to icon buttons
* Align some background colors with figma

### DIFF
--- a/src/elements/button/button.svelte
+++ b/src/elements/button/button.svelte
@@ -68,8 +68,8 @@ const handleParentClick = (event: PointerEvent) => {
     {title}
     class={cx({
       'h-[30px] w-[30px]': variant === 'icon',
-      'px-4': !icon,
-      'pl-3 pr-4': icon,
+      'px-4': !icon || variant === 'icon',
+      'pl-3 pr-4': icon && variant !== 'icon',
       'inline-flex items-center justify-center gap-1.5 py-1 text-sm border':
         variant !== 'icon',
       'bg-light border-light hover:bg-medium hover:border-medium active:bg-gray-2':
--- a/src/elements/button/button.svelte
+++ b/src/elements/button/button.svelte
@@ -8,6 +8,7 @@
 type Variants =
   | 'primary'
   | 'inverse-primary'
+  | 'ghost'
   | 'danger'
   | 'outline-danger'
   | 'success'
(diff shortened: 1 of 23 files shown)


src/elements/button/button.svelte
```diff
@@ -65,17 +66,22 @@ const handleParentClick = (event: PointerEvent) => {
...
+      'h-[30px] w-[30px]': variant === 'icon',
+      'px-4': !icon || variant === 'icon',
+      'pl-3 pr-4': icon && variant !== 'icon',
+      'inline-flex items-center justify-center gap-1.5 py-1 text-sm border':
         variant !== 'icon',
-      'bg-light border-light hover:bg-medium hover:border-medium':
+      'bg-light border-light hover:bg-medium hover:border-medium active:bg-gray-2':
         variant === 'primary',
-      'bg-gray-9 border-gray-9 text-white hover:bg-black hover:border-black':
+      'bg-gray-9 border-gray-9 text-white hover:bg-black hover:border-black active:bg-[#000]':
```
mrloureed, line 77: I would add a black or dark variant here rather than using [#000]

```
    backgroundColor: {
      extralight: '#FBFBFC',
      light: '#F7F7F8',
      medium: '#F1F1F4',
    },
```

src/elements/button/button.svelte
```diff
@@ -65,17 +66,22 @@ const handleParentClick = (event: PointerEvent) => {
...
         variant === 'inverse-primary',
-      'bg-danger-dark text-white border-danger-dark hover:bg-[#aa2a2b]':
+      'border-transparent hover:bg-[rgba(0,0,0,0.04)] active:bg-[rgba(0,0,0,0.08)] active-border-[rgba(0,0,0,0.08)]':
+        variant === 'ghost' || variant === 'icon',
+      'bg-danger-dark text-white border-danger-dark hover:bg-[#aa2a2b] active:bg-[#9e2728]':
```
mrloureed, line 81: Why the custom colors?

src/elements/button/button.svelte
```diff
@@ -65,17 +66,22 @@ const handleParentClick = (event: PointerEvent) => {
...
         variant === 'danger',
       'bg-success-dark border-success-dark text-white': variant === 'success',
-      'bg-danger-light border-danger-medium text-danger-dark hover:bg-[#f5dfdc]':
+      'bg-danger-light border-danger-medium text-danger-dark hover:bg-[#f5dfdc] active:bg-[#f6d7d3]':
```
mrloureed, line 84: same as above

mrloureed (changes requested):

src/elements/button/button.svelte
micheal-parks: These are colors that design wants for specific states in this button, but shouldn't be used elsewhere. For that reason it's best to keep them here and out of the theme.

src/elements/button/button.svelte
micheal-parks: same as above

src/elements/button/button.svelte
micheal-parks: We don't want to use pure black outside of this file

mrloureed (approved):
LGTM

PR #238, opened 2023-05-11
APP-1844 Create context menu component

ehhong:
Context menu represents the menu itself -- any positioning / floating behavior must be externally managed (see test example)

design https://www.figma.com/file/ERHSYDGTXiEQOVGY8ORg6I/PRIME?type=design&node-id=2703-5151&t=PUd4XMOf36ba4gDT-0

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   },
   "devDependencies": {
     "@floating-ui/dom": "^1.2.6",
+    "@floating-ui/vue": "^1.0.0",
     "@playwright/test": "^1.32.3",
     "@storybook/addon-a11y": "^7.0.7",
     "@storybook/addon-actions": "^7.0.7",
--- a/playground/src/context-menu-test.vue
+++ b/playground/src/context-menu-test.vue
@@ -1,0 +1,73 @@
+<script setup lang="ts">
+import { useFloating } from '@floating-ui/vue';
+import { ref } from "vue";
+
+let showMenu = $ref<boolean>(false)
+
+const parent = ref<HTMLElement>();
+const child = ref<HTMLElement>();
+const { floatingStyles } = useFloating(parent, child, {placement: 'bottom-start'});
+
+const handleSelect = (event: any) => {
+  console.log('select', event.detail.value)
+}
+
+</script>
+
+<template>
+  <main>
+    <v-context-menu
+      data-testid="context-menu-default"
+    >
+      <v-context-menu-item
+        label="label 1"
+      />
+      <v-context-menu-separator />
+      <v-context-menu-item
+        label="label 2"
+        variant="primary"
+      />
+      <v-context-menu-item
+        icon="trash"
+        label="label 3"
+      />
+      <v-context-menu-item
+        icon="x"
+        label="danger"
+        variant="danger"
+      />
+    </v-context-menu>
+
+
+    <!-- Context menu with floating UI example -->
+    <div
+      ref="parent"
+      @click="showMenu = !showMenu"
+    >
+      hi there this is the parent
+    </div>
+    <v-context-menu 
+      v-if="showMenu" 
+      ref="child"
+      :style="floatingStyles"
+      @click="showMenu = false"
+    >
+      <v-context-menu-item 
+        icon="send"
+        label="and this is this child"
+        @select="handleSelect"
+      />
+      <v-context-menu-item-separator />
+      <v-context-menu-item 
+        label="label really long item words really really really long"
+        @select="handleSelect"
+      />
+      <v-context-menu-item 
+        icon="x"
+        label="label"
+        variant="danger"
+        @select="handleSelect"
+      />
+    </v-context-menu>
+  </main>
+</template>
--- a/src/elements/context-menu/context-menu-item.svelte
+++ b/src/elements/context-menu/context-menu-item.svelte
@@ -1,0 +1,46 @@
+<svelte:options immutable tag="v-context-menu-item" />
+
+<script lang="ts">
+import cx from 'classnames';
+import { addStyles } from '../../lib/index';
+import { dispatcher } from '../../lib/dispatch';
+
+const dispatch = dispatcher();
+
+addStyles();
+
+export let icon: string;
+export let variant: 'primary' | 'danger' = 'primary';
+export let label: string;
+
+const handleSelect = () => {
+  dispatch('select', { value: label });
+};
+</script>
+
+<button
+  role="menuitem"
+  aria-labelledby={label}
+  class="flex flex-row px-2 py-1.5 hover:bg-light w-full text-left"
+  on:click={handleSelect}
+>
+  {#if icon}
+    <v-icon
+      name={icon}
+      class={cx('p-1', {
+        'text-gray-6': variant === 'primary',
+        'text-danger-dark': variant === 'danger',
+      })}
+    />
+  {/if}
+
+  <p
+    id={label}
+    class={cx('font-normal', {
+      'text-default': variant === 'primary',
+      'text-danger-dark': variant === 'danger',
+    })}
+  >
+    {label}
+  </p>
+</button>
--- a/src/elements/context-menu/context-menu.svelte
+++ b/src/elements/context-menu/context-menu.svelte
@@ -1,0 +1,11 @@
+<svelte:options immutable tag="v-context-menu" />
+
+<script lang="ts">
+import { addStyles } from '../../lib/index';
+
+addStyles();
+</script>
+
+<div role="menu" class="py-1 border border-medium shadow-sm max-w-xs">
+  <slot class="filter-none" />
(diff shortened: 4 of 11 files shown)


playground/src/context-menu-test.vue
```diff
@@ -0,0 +1,50 @@
+<script setup lang="ts">
+import { useFloating } from '@floating-ui/vue';
+import { ref } from "vue";
+
+let showMenu = $ref<boolean>(false)
+
+const parent = ref<HTMLElement>();
+const child = ref<HTMLElement>();
+const { floatingStyles } = useFloating(parent, child, {placement: 'bottom-start'});
```
micheal-parks, line 9: I wouldn't recommend testing anything relating to floating UI here. The library itself could break or change, and maintenance would then would be complicated for someone trying to introduce an unrelated change to the library.

ehhong: what do you think of leaving the floating UI stuff in this test file as an example? I am not writing any playwright tests related to floating UI

tests/icon.spec.ts
```diff
@@ -20,62 +20,62 @@ test('Renders and icon as extra small if size attribute is specified as xs', asy
   page,
 }) => {
   await expect(page.getByTestId('icon-size-xs').locator('i')).toHaveClass(
-    /text-xs/
+    /text-xs/u
```
ehhong, line 23: fixing lint warnings

tests/select.spec.ts
```diff
@@ -23,9 +23,9 @@ test('Given an options attribute, on select the options should be visible and cl
   // within the optionsContainer there should be 3 different options
   expect(await optionsContainer.locator('label').all()).toHaveLength(3);
 
-  await expect(optionsContainer).toHaveText(/one/);
-  await expect(optionsContainer).toHaveText(/two/);
-  await expect(optionsContainer).toHaveText(/three/);
+  await expect(optionsContainer).toHaveText(/one/u);
+  await expect(optionsContainer).toHaveText(/two/u);
+  await expect(optionsContainer).toHaveText(/three/u);
```
ehhong, line 28: lint warnings

src/elements/context-menu/context-menu.svelte
```diff
@@ -0,0 +1,11 @@
+<svelte:options immutable tag="v-context-menu" />
+
+<script lang="ts">
+import { addStyles } from '../../lib/index';
+
+addStyles();
+</script>
+
+<div class="py-1 border border-medium shadow-sm max-w-xs">
```
mcous, line 9: Should this have an `aria-role` of [`menu`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role)?

src/elements/context-menu/context-menu-item.svelte
```diff
@@ -0,0 +1,53 @@
...
+
+const handleSelect = () => {
+  dispatch('select', { value: label });
+};
+</script>
+
+<div>
+  {#if isDivider}
+    <hr class="border-light mt-1 mb-0.5 drop-shadow-none" />
+  {:else}
```
mcous, line 28: I think `v-context-menu-divider` should be a separate component; it muddies the water a little by overloading the `menu-item`, which is a different role; `separator` vs `menuitem`

tests/context-menu.spec.ts
```diff
@@ -0,0 +1,69 @@
+import { test, expect } from '@playwright/test';
+import { waitForCustomEvent } from './lib/helper.js';
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('/playground/context-menu-test.html');
+});
+
+test('Renders labels, dividers, and icons', async ({ page }) => {
+  const contextMenu = page.getByTestId('context-menu-default');
```
mcous, line 9: I think in these tests we should strive to minimize `getByTestId`. I think using it to select to overall component for the given `test` is good, but other usages are questionable

tests/context-menu.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  await expect(contextMenu).toBeVisible();
+  await expect(contextMenu.getByText('label 1')).toBeVisible();
+  await expect(contextMenu.getByTestId('context-menu-divider')).toBeVisible();
```
mcous, line 12: An `hr` should have an implicit role of `separator`, so we should use that instead of `test-id`

src/elements/context-menu/context-menu-item.svelte
```diff
@@ -0,0 +1,53 @@
...
+export let label: string;
+
+$: isDivider = htmlToBoolean(divider, 'divider');
+
+const handleSelect = () => {
+  dispatch('select', { value: label });
+};
+</script>
+
+<div>
```
mcous, line 25: Should have a `role` of [`menuitem`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role)

tests/context-menu.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+import { waitForCustomEvent } from './lib/helper.js';
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('/playground/context-menu-test.html');
+});
+
+test('Renders labels, dividers, and icons', async ({ page }) => {
+  const contextMenu = page.getByTestId('context-menu-default');
+  await expect(contextMenu).toBeVisible();
+  await expect(contextMenu.getByText('label 1')).toBeVisible();
```
mcous, line 11: Can these search by `menuitem` role and check the text, instead?

tests/context-menu.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  // renders with icon
+  await expect(contextMenu.getByText('label 3')).toBeVisible();
+  await expect(
+    contextMenu.getByTestId('context-menu-icon').locator('i')
+  ).toBeVisible();
+  await expect(
+    contextMenu.getByTestId('context-menu-icon').locator('i')
+  ).toHaveClass(/icon-trash/u);
+  await expect(
+    contextMenu.locator('i:left-of(:text("label 3"))').first()
```
mcous, line 24: Not super familiar with the locator syntax, but could this be `i.icon-trash:left-of...` or similar to collapse these two assertions?

mcous (commented):
Styling matches the Figma perfectly, as far as I can tell. Two suggestions for the component itself from an accessibility standpoint:

- Add roles
- Split apart "item" and "divider" into two separate components

With these roles added, you should also be able to reduce usage of `getByTestId` in the tests

package.json
```diff
@@ -89,5 +89,8 @@
     "vite": "4.3.1",
     "vue": "^3.2.45",
     "vue-tsc": "^1.2.0"
+  },
+  "dependencies": {
+    "@floating-ui/vue": "^1.0.0"
```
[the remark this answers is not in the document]
ehhong: good catch! thank you

tests/context-menu.spec.ts
ehhong: yup, this worked!

mcous (approved):
Nice!

playground/src/context-menu-test.vue
```diff
@@ -0,0 +1,73 @@
...
+      ref="child"
+      :style="floatingStyles"
+      @click="showMenu = false"
+    >
+      <v-context-menu-item 
+        icon="send"
+        label="and this is this child"
+        @select="handleSelect"
+      />
+      <v-context-menu-item divider/>
```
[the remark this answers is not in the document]
ehhong: oops, thanks for catching!

src/elements/context-menu/context-menu-item.svelte
```diff
@@ -0,0 +1,48 @@
...
+export let icon: string;
+export let variant: 'primary' | 'danger' = 'primary';
+export let label: string;
+
+const handleSelect = () => {
+  dispatch('select', { value: label });
+};
+</script>
+
+<div>
```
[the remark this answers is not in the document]
ehhong: good catch! don't need

src/elements/context-menu/context-menu-item.svelte
```diff
@@ -0,0 +1,48 @@
...
+    {/if}
+
+    <p
+      id={label}
+      class={cx('font-normal', {
+        'text-default': variant === 'primary',
+        'text-danger-dark': variant === 'danger',
+      })}
+    >
+      {label}
```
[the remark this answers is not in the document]
ehhong: ah yea, i like the idea of using IDs and children for content. even when i was leveraging `label` for the `onselect` behavior here, it felt a bit wrong because your display content might be unwieldy and not great for an event id.

i'm going to leave `label` here for now, but agree we should have this conversation and update components as we move to a more atomic design. is there somewhere, maybe in jira, where we can add a reminder?

PR #239, opened 2023-05-12
Fix lint warnings in test files

mcous:
This PR fixes all lint warnings in test files. Broke apart test file changes and `src` changes for safety

PR #245, opened 2023-06-05
Fix ESLint errors

mcous:
## Overview

Follows up on #239 and fixes all remaining ESLint errors. I left warnings (and `--quiet` in CI) in place in CI due to false warnings from our outdated Svelte plugin, to be resolved in a future PR.

~Depends on viamrobotics/js-config#11~

## Change log

Mostly this amounted to running `eslint --fix` and then cleaning up the stragglers. **This PR caught one bug**, which this PR also fixes:

- Fix regex escape in select utils
    - Adding the `/u` option also enables a sort of "strict mode" for regex
    - Turns out, our regex escape logic produces invalid regex
    - I replaced our logic with the escape logic from MDN's page on regex

## Review requests

- Be on the lookout for any auto-changes that may not be safe
    - e.g. ESLint tried to remove some `nullish?.access` question marks due to inaccurate typing
- Highlight any changes you think may not be covered by automated tests


mcous:
Closing due to extreme staleness

PR #255, opened 2023-06-17
Switch component fix -- event handlers for labels and switch button

mrloureed:
https://viam.atlassian.net/browse/APP-2069

Clicking on labels is not triggering desired event.


DTCurrie (approved):
✅

PR #257, opened 2023-06-22
APP-2072 Use pnpm instead of npm in PRIME

ethanlook:
Migrates PRIME to pnpm. No structural changes to the package structure of PRIME yet. I validated that all scripts still run as expected.

Verified GitHub workflow changes: [link](https://github.com/viamrobotics/prime/actions/runs/5357167551/jobs/9717661074)


ethanlook:
Hang tight on the review - I need to update the GitHub workflow.

ethanlook:
> Hang tight on the review - I need to update the GitHub workflow.

Done ✅

.github/actions/setup/action.yml
```diff
@@ -5,16 +5,23 @@ inputs:
...
+  pnpm-version:
+    description: 'pnpm version to install'
+    default: 8
 
 runs:
   using: 'composite'
   steps:
-    - name: 'Install Node.js'
+    - name: 'Install pnpm ${{ inputs.pnpm-version }}'
+      uses: 'pnpm/action-setup@v2'
```
mcous, line 16: Following GH Actions security best practices, we should pin this to a specific action commit

```suggestion
      uses: pnpm/action-setup@ c3b53f6a16e57305370b4ae5a540c2077a1d50dd # v2.2.4
```

.github/actions/setup/action.yml
```diff
@@ -5,16 +5,23 @@ inputs:
   node-version:
     description: 'Node.js version to install'
     default: '18'
+  pnpm-version:
+    description: 'pnpm version to install'
+    default: 8
```
mcous, line 10: I think we should leverage `corepack` and the `packageManager` field in package.json instead of specifying this in the action

The `pnpm/action-setup` action will read this field automatically, so we don't have to specify it here

package.json
mcous, line 1: Per my other comment, we should add...

```json
{
  "packageManager": "pnpm@8.6.3"
}
```

...to package.json so we (and CI) are all on the same version, and `corepack` can install it automatically during local dev

mcous (commented):
Looks great, just discussed this in person

package.json
ethanlook: Done

.github/actions/setup/action.yml
ethanlook: Done

.github/actions/setup/action.yml
ethanlook: Done. Did the same for the node setup action.

mcous: I tend to be a little more relaxed with the GitHub maintained actions, but certainly no harm in locking! I'd rather we lock everything than nothing

micheal-parks (approved):

DTCurrie (approved):
✅

PR #258, opened 2023-06-22
App 2068 bugfix undefined replace

mrloureed:
In app there's an error occurring that does not occur in playground or storybook.

`prime.js:2787 Uncaught TypeError: Cannot read properties of undefined (reading 'replaceAll')`


ehhong (approved):
lgtm!

DTCurrie (approved):
✅

PR #263, opened 2023-06-29
APP-1996 listen to elements instead of the page in PRIME tests

ethanlook:
This migrates the tests to listen for custom events on the element itself instead of the page.

When migrating to Svelte 4, we will no longer be able to dispatch a custom event that bubbles up the page. (This is consistent with Svelte's event dispatcher that events do not bubble.) To prepare for that, the tests should listen for events on the element itself.

Some of the tests began failing for unknown reasons, particularly those that test "enter" key press. I disabled those tests instead of figuring out why the custom events are not propogating correctly. We will re-enable those tests once the components export Svelte components and we can migrate the tests to use those directly.

### DIFF
--- a/tests/lib/helper.spec.ts
+++ b/tests/lib/helper.spec.ts
@@ -2,12 +2,22 @@
 import { test, expect } from '@playwright/test';
 import { waitForCustomEvent } from './helper.js';
 
-test('Resolves with the details of a custom event', async ({ page }) => {
-  const eventHandler = waitForCustomEvent(page, 'custom');
-
+test.beforeEach(async ({ page }) => {
   await page.evaluate(() => {
+    const div = document.createElement('div');
+    // eslint-disable-next-line unicorn/prefer-dom-node-dataset
+    div.setAttribute('data-testid', 'test-div');
+    document.body.append(div);
+  });
+});
+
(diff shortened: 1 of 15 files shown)


tests/lib/helper.spec.ts
```diff
@@ -2,12 +2,22 @@
 import { test, expect } from '@playwright/test';
 import { waitForCustomEvent } from './helper.js';
 
+test.beforeEach(async ({ page }) => {
+  await page.evaluate(() => {
+    const div = document.createElement('div');
+    // eslint-disable-next-line unicorn/prefer-dom-node-dataset
+    div.setAttribute('data-testid', 'test-div');
```
micheal-parks, line 9: Why not use `dataset`?

micheal-parks (approved):

tests/lib/helper.spec.ts
ethanlook: It's a runtime error, sadly:

```
Error: page.evaluate: DOMException: Failed to set a named property on 'DOMStringMap': 'data-testid' is not a valid property name.
```

micheal-parks: weird, carry on!

mcous (approved):
LGTM 🎈

PR #264, opened 2023-06-30
Reset toast notifications rather than duplicating them

micheal-parks:
This PR changes the behavior of toasts so that if a duplicate message is pushed, an existing toast will reset its timer and migrate to the top of the list rather than stacking duplicate messages.

toasts


DTCurrie (approved):
✅

PR #265, opened 2023-07-05
APP-2072 Migrate PRIME to Svelte 4.

ethanlook:
@mcous thanks for the inspiration for PR descriptions :)

## Overview

This migrates PRIME to Svelte 4. I wanted to do this migration before changing the package structure.

## Change log

- `npx svelte-migrate@latest svelte-4`
- Fix package version mismatches
- Import styles directly instead of `addStyles` helper
- Fix button and input `attachInternals`
- Fix code editor injected styles
- Change slider `range` prop to only accept a string - mixed types doesn't play well with web components
- Use Svelte `createEventDispatcher` directly - `get_current_component` no longer works
- Fix problematic tests

## Review requests

Pull down the branch and make sure:
- `pnpm run test` passes
- Elements look/behave correctly with `pnpm run storybook` and `pnpm start`

Questions:
- Is the minor version bump to 0.3.0 appropriate? I imagine we'll bump to 1.0.0 with the package restructure since that will be breaking.


package.json
```diff
@@ -76,22 +76,22 @@
...
     "monaco-editor": "^0.37.1",
     "postcss": "^8.4.23",
     "prettier": "^2.8.8",
-    "prettier-plugin-svelte": "^2.10.0",
+    "prettier-plugin-svelte": "^2.10.1",
     "prismjs": "^1.29.0",
     "react": "^18.2.0",
     "remark-gfm": "^3.0.1",
     "storybook": "^7.0.7",
-    "svelte": "^3.58.0",
```
micheal-parks, line 90: Can we upgrade to latest? There are some patch upgrades out 🕺

package.json
```diff
@@ -76,22 +76,22 @@
...
-    "svelte-check": "^3.2.0",
+    "svelte": "^4.0.0",
+    "svelte-check": "^3.4.3",
     "svelte-preprocess": "^5.0.3",
     "tailwindcss": "^3.3.1",
-    "typescript": "^5.0.4",
```
micheal-parks, line 94: y go back?

src/elements/tooltip.svelte
```diff
@@ -24,7 +23,7 @@ $: {
 let x = 0;
 let y = 0;
 
-export const recalculateStyle = async () => {
+export let recalculateStyle = async () => {
```
micheal-parks, line 26: Why the change to `let`?

ethanlook: Sadly, using `const` produces a runtime error on the `create_custom_element` call:

```
TypeError: Cannot redefine property: recalculateStyle
```

mcous:
> thanks for the inspiration for PR descriptions :)

The template is here for the taking, if we want it!
<https://github.com/Opentrons/opentrons/blob/edge/.github/PULL_REQUEST_TEMPLATE.md>

DTCurrie:
> > thanks for the inspiration for PR descriptions :)
> 
> The template is here for the taking, if we want it! https://github.com/Opentrons/opentrons/blob/edge/.github/PULL_REQUEST_TEMPLATE.md

If there is no opposition to it I don't mind setting it as our template.

package.json
ethanlook: I believe this was changed by the svelte upgrade script. I upgraded to the latest.

src/elements/tooltip.svelte
micheal-parks: ahhh interesting

.eslintrc.cjs
```diff
@@ -5,6 +5,7 @@ module.exports = {
   env: {
     browser: true,
   },
+  parser: '@typescript-eslint/parser',
```
mcous, line 8: [Nit] I think this might be set by `@viamrobotics/eslint-config`, and may not be needed. However, it is not harmful to include

src/elements/badge.svelte
```diff
@@ -25,3 +22,7 @@ addStyles();
 >
   {label}
 </small>
+
+<style>
+@import './../../prime.css';
```
mcous, line 27: [nit] does `./..` === `../`?

```suggestion
@import '../../prime.css';
```

src/elements/badge.svelte
```diff
@@ -25,3 +22,7 @@ addStyles();
 >
   {label}
 </small>
+
+<style>
```
mcous, line 26: Is this running through PostCSS? I thought [this was necessary](https://tailwindcss.com/docs/guides/sveltekit):

```suggestion
<style lang="postcss">
```

ethanlook:
Bundle size is down with Svelte 4!

Before:
```
dist/prime.css   25.84 kB │ gzip:  5.49 kB
dist/prime.js   329.75 kB │ gzip: 72.66 kB
dist/prime.umd.cjs  210.17 kB │ gzip: 59.10 kB
```

After:
```
dist/prime.css   25.84 kB │ gzip:  5.49 kB
dist/prime.js   318.95 kB │ gzip: 71.46 kB
dist/prime.umd.cjs  190.12 kB │ gzip: 56.52 kB
```

mcous (approved):
Tests and playground LGTM!

> Is the minor version bump to 0.3.0 appropriate? I imagine we'll bump to 1.0.0 with the package restructure since that will be breaking.

Works for me. Since we're pre 1.0, `0.3.0` is _not_ semver compatible with `^0.2.0`, which should cover our bases

micheal-parks (approved):
🎉 glad we arrived at this solution

src/main.ts
```diff
@@ -5,44 +5,91 @@ import '../prime.css';
...
+ * https://github.com/sveltejs/svelte/blob/master/packages/svelte/src/runtime/internal/public.d.ts#L72-L73
+ */
+type ComponentWithElement = SvelteComponent & {
+  element: typeof HTMLElement;
+};
+
+const main = async () => {
+  // These components are exported as pure Svelte components.
+  const componentModules = {
+    'v-badge': await import('./elements/badge.svelte'),
```
mcous, line 24: Can these become static imports since we removed the tag side effect?

mcous (approved):
Looks great! One question about the need for dynamic imports in main, but that can be moved to the restructuring PR

src/main.ts
ethanlook: Yup, that's totally reasonable. We'll do that when we restructure the package.

PR #266, opened 2023-07-07
Design maintenance Jul 7 2023

micheal-parks:
Done in a session with nat. Mostly involves notification and modal tweaks.


DTCurrie:
What was the fix for the notify "art" bug?

micheal-parks:
@DTCurrie 

We restructured slots to be more intuitive, but this will actually require a follow up in app. There's an example added that creates a notification that would exist on the history page without that visual bug 🎉

src/stories/notify.stories.mdx
```diff
@@ -178,11 +178,37 @@ Notify message of type info, success, warning, or error
     {() => `
 <v-notify
   variant='info'
-  title='This is the title.'
-  message='This is the message.'
+  title='You have an incoming life update'
+  message='LifeOS has been upgraded to a new version that includes great new things such as friends and family and you will not be alone anymore.'
```
mrloureed, line 182: 🤣

mrloureed (approved):
LGTM!

PR #267, opened 2023-07-10
App 2049 icons update

mrloureed:
Update icon library to use @mdi/js

Audit current icon set


mrloureed:
@nathartman How do you like this iteration of the current icon set in Prime? I tried to make more consistent and use outline versions where it made sense. That being said, I'm happy to do whatever you suggest!

image

nathartman:
Looks great from what I can tell! Yay

package.json
```diff
@@ -95,5 +95,9 @@
     "vite": "4.3.1",
     "vue": "^3.2.45",
     "vue-tsc": "^1.2.0"
+  },
+  "dependencies": {
+    "@mdi/js": "^7.2.96",
```
micheal-parks, line 100: These should be devDependencies since we don't want them added into the `node_modules` of anything downstream like app or rc

src/elements/icon.svelte
```diff
@@ -3,22 +3,132 @@
...
+  class={cx(sizes[size], {
+    'inline-block': hasNameProperty,
+    hidden: !hasNameProperty,
   })}
-/>
+  viewBox="0 0 24 24"
+  aria-labelledby="name"
+  role="img"
+>
+  {#if sizes[size]}
```
micheal-parks, line 130: Why do we need to check for size here?

src/elements/button/button.svelte
```diff
@@ -46,9 +45,29 @@ const handleClick = () => {
 const handleParentClick = (event: PointerEvent) => {
   event.stopImmediatePropagation();
 };
+
+let fill = '';
+
+switch (variant) {
```
micheal-parks, line 51: We should wrap this in a `$` so that it's reactive (e.g. if a user updates the `variant` prop we'd want fill to update)

```ts
$: {
  switch (variant) { ... }
}

micheal-parks (approved):

PR #277, opened 2023-07-21
[DATA-1742] Add full width option to button and radio

alexis-wei:
add a `width` field to both `v-button` and `v-radio` to allow the elements to stretch the full width of its container

### DIFF
--- a/src/elements/button/button.svelte
+++ b/src/elements/button/button.svelte
@@ -126,6 +126,6 @@ $: {
 <style>
 :host {
   --button-display: {display};
-  display: var(--button-display) !important;
+  display: var(--button-display);
 }
 </style>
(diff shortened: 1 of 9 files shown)


src/elements/button/button.svelte
```diff
@@ -121,6 +125,7 @@ $: {
 
 <style>
 :host {
-  display: inline-block !important;
+  --button-display: {display};
+  display: var(--button-display) !important;
```
micheal-parks, line 129: You can just prefix the classes above with `!` to avoid having to do this:
```
$: display = width === 'full' ? '!block' : '!inline-block';
```
why is important necessary though?

micheal-parks (approved):

PR #282, opened 2023-07-25
APP-2280 Add @viamrobotics/prime-core package

ethanlook:
This adds a new `@viamrobotics/prime-core` package to PRIME. The package is a SvelteKit component library.

## Change log

- `npm create svelte@latest core`
    - Use "component library" option
- Add Badge component
    - Add Badge tests using Vitest
    - Use Badge component in dev environment (not published)
- Update GitHub Workflow to run all checks, builds, and tests
- Update GitHub Workflow to publish NPM package

## Testing

- Added `Badge` tests
- Dry run [npm publish](https://github.com/viamrobotics/prime/actions/runs/5673468740/job/15375029127?pr=282) - please check that the package is structured as expected

## Review requests

Please try to pull down the branch and see if you can use the component in your own local work. Instructions are in the README.


packages/core/.eslintignore
mcous, line 1: I think this was probably the generated SvelteKit default, but it always bugs me that they'll just jam everything in these ignore files without thought, leading to bloated files that don't do anything

packages/core/.eslintignore
```diff
@@ -0,0 +1,14 @@
+.DS_Store
```
mcous, line 1: Not a JS file, ESLint doesn't care

packages/core/.eslintignore
```diff
@@ -0,0 +1,14 @@
...
+node_modules
+/build
+/.svelte-kit
+/package
+.env
+.env.*
+!.env.example
```
mcous, line 8: Not files ESLint cares about

packages/core/.npmrc
```diff
@@ -0,0 +1,2 @@
+engine-strict=true
+resolution-mode=highest
```
mcous, line 2: Haven't messed with the `resolution-mode` setting, was this something you set? What's reasoning? (Might be worth a comment, especially if we want to set this as a default across other projects)

packages/core/.prettierignore
mcous, line 1: Same comments about the `.eslintignore`, with the caveat that that the `pnpm-lock.yaml` ignore is actually important here

packages/core/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Installation
+
+Add PRIME core using your package manager of choice:
+
+```
+pnpm add --save-dev @viamrobotics/prime-core
+```
+
+Install [Tailwind][]. In the `tailwind.config.js`, add the components to the content and include the theme:
```
mcous, line 15: Are we missing a `[tailwind]` anchor href down below?

packages/core/package.json
```diff
@@ -0,0 +1,73 @@
...
+    },
+    "./prime.css": "./src/prime.css",
+    "./theme.js": "./src/theme.js"
+  },
+  "files": [
+    "dist",
+    "!dist/**/*.test.*",
+    "!dist/**/*.spec.*",
+    "src/prime.css",
+    "src/theme.js"
```
mcous, line 30: If we're generating our output files with source maps, I like to include the whole `src` directory. It increases the download size a little, but doesn't harm the actual bundle size, and can help with debugging

packages/core/package.json
```diff
@@ -0,0 +1,73 @@
...
+  ],
+  "peerDependencies": {
+    "svelte": "^4.0.0"
```
mcous, line 33: This is definitely a correct spec, but I like to spec peers with explicit bottom and top values, to make it easier in the future to simply widen the range, rather than mindlessly update the peer and accidentally introduce a breaking change to the package contract.

```suggestion
    "svelte": ">=4.0.0 <5"
```

packages/core/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@viamrobotics/prime-core",
+  "version": "0.0.1",
+  "scripts": {
+    "dev": "vite dev",
+    "build": "vite build && npm run package",
```
mcous, line 6:
```suggestion
    "build": "vite build && pnpm run package",
```

packages/core/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "preview": "vite preview",
+    "package": "svelte-kit sync && svelte-package && publint",
+    "prepublishOnly": "npm run package",
+    "check": "concurrently -g pnpm:check-*",
+    "check-svelte": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json",
+    "check-lint": "prettier --check . && eslint .",
+    "format": "prettier --write .",
```
mcous, line 13: Are we missing extensions on either of these calls? IIRC (could be wrong) using `.` will only check prettier and eslint defaults, which won't include `.svelte`, `.ts`, etc

packages/core/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "test": "vitest run",
+    "test:watch": "vitest"
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "svelte": "./dist/index.js"
```
mcous, line 20: I assume this is what SvelteKit outputs? A little suprising there isn't a `default` or `import` export, but I guess it makes sense if this only works with Svelte

packages/core/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@viamrobotics/prime-core",
+  "version": "0.0.1",
+  "scripts": {
```
mcous, line 4: We'll want a publish config section (there are a few ways to do this, `publishConfig` is my favorite because it's portable) to tell `npm` on the initial publish that this namespaced package should be public, because the npm default for namespaced packages is private

```suggestion
  "publishConfig": {
    "access": "public"
  },
  "scripts": {
```

packages/core/vite.config.ts
```diff
@@ -0,0 +1,11 @@
+import { sveltekit } from '@sveltejs/kit/vite';
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  plugins: [sveltekit()],
+  test: {
+    include: ['src/**/*.{test,spec}.{js,ts}'],
+    globals: true,
```
mcous, line 8: Rather than setting `globals: true`, we can set up a `vitest.setup.ts` file to install `@testing-library/jest-dom` and configure the testing-library reset. I softly prefer it to globals mode which feels to me like a silly holdover from pre-module times

mcous (approved):
Looks fantastic to me! Got some random package management nits and thoughts, but otherwise LGTM!

(Except for that `publishConfig` comment, which is needed or else the first publish will fail)

packages/core/.prettierignore
ethanlook: Why is the `pnpm-lock.yaml` important? It's located in the parent workspace

packages/core/package.json
ethanlook: Yup, this is what SvelteKit outputs. Also a little surprising but it makes sense.

packages/core/src/theme.js
```diff
@@ -0,0 +1,52 @@
+export const theme = {
```
micheal-parks, line 1: Theme is used both in core and legacy correct? are we having legacy import this theme from core? If not I'd probably recommend that since problems created by design drift I think would be noisier worse than problems caused by cross pollinating data between packages in this case

ethanlook: Great idea, I just made legacy import from core.

.github/workflows/ci.yml
```diff
@@ -62,21 +67,26 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - name: 'Install Node.js'
-        uses: actions/setup-node@v3
+      - name: Install 🔧
+        uses: ./.github/actions/setup
         with:
-          node-version: 18
-      - name: Download npm artifacts
+          node-version: 16
```
mcous, line 73: Node 16 is EOL'ing in a few months; was this change intentional?

packages/legacy/.eslintrc.cjs
```diff
@@ -8,7 +10,10 @@ module.exports = {
   parserOptions: {
     tsConfigRootDir: __dirname,
```
mcous, line 11: Is this casing correct?

```suggestion
    tsconfigRootDir: __dirname,
```

packages/legacy/.eslintrc.cjs
```diff
@@ -8,7 +10,10 @@ module.exports = {
...
     extraFileExtensions: ['.svelte'],
-    project: ['./tsconfig.json', './playground/tsconfig.json'],
+    project: [
+      path.join(__dirname, './tsconfig.json'),
+      path.join(__dirname, './playground/tsconfig.json'),
```
mcous, line 15: These `path.join`'s shouldn't be required; might be related to casing comment above

packages/legacy/package.json
```diff
@@ -67,6 +63,7 @@
     "@typescript-eslint/parser": "^5.59.0",
     "@viamrobotics/eslint-config": "^0.0.4",
     "@viamrobotics/prettier-config": "^0.0.1",
+    "@viamrobotics/prime-core": "workspace:../core",
```
mcous, line 66: Oh duh, hadn't considered that we'd added a `workspace` dependency. This makes our usage of `pnpm publish -r` _required_ rather than "nice to have" (to replace `workspace` deps with the actual versions).

Glad you took care of that!

mcous (approved):
LGTM, might be a little (unharmful) weirdness remaining in the legacy ESLint config because of course there is it's ESLint

packages/legacy/package.json
ethanlook: I just added this dependency so we could share the theme across packages, but we didn't do that when you first reviewed. It's much better now anyways!

.github/workflows/ci.yml
ethanlook: Nope - copy pasta

packages/legacy/.eslintrc.cjs
ethanlook: It was not! Great catch.

packages/legacy/.eslintrc.cjs
ethanlook: Unfortunately the casing did not affect the need for the `path.join`s. Leaving as-is.

PR #285, opened 2023-07-27
APP-2283 Migrate Breadcrumbs to @viamrobotics/prime-core

ethanlook:
Migrates Breadcrumbs component to `@viamrobotics/prime-core`.

## Change log

- Copy `breadcrumbs.svelte` to `packages/core`
    - Change props to accept an array
- Add `Breadcrumbs` export from `index.ts`
- Copy unit tests
    - Change to use vitest
- Copy storybook story to `packages/storybook`
    - Change to use Svelte component directly

## Testing

- Added unit tests

## Review requests

- None

### DIFF
--- a/packages/core/src/lib/breadcrumbs.spec.ts
+++ b/packages/core/src/lib/breadcrumbs.spec.ts
@@ -1,0 +1,11 @@
+import { describe, it, expect } from 'vitest';
+import { render, screen } from '@testing-library/svelte';
+import Breadcrumbs from './breadcrumbs.svelte';
+
+describe('Breadcrumbs', () => {
+  it('Renders breadcrumbs with the list of values specified in crumbs attribute as pills', () => {
+    render(Breadcrumbs, { crumbs: ['Chocolate Chip', 'Oatmeal Raisin'] });
+    expect(screen.getByText('Chocolate Chip')).toBeInTheDocument();
+    expect(screen.getByText('Oatmeal Raisin')).toBeInTheDocument();
+  });
+});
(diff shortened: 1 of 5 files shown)


packages/core/src/lib/breadcrumbs.spec.ts
```diff
@@ -0,0 +1,11 @@
...
+import { render, screen } from '@testing-library/svelte';
+import Breadcrumbs from './breadcrumbs.svelte';
+
+describe('Breadcrumbs', () => {
+  it('Renders breadcrumbs with the list of values specified in crumbs attribute as pills', () => {
+    render(Breadcrumbs, { crumbs: ['Chocolate Chip', 'Oatmeal Raisin'] });
+    expect(screen.getByText('Chocolate Chip')).toBeInTheDocument();
+    expect(screen.getByText('Oatmeal Raisin')).toBeInTheDocument();
+  });
+});
```
ethanlook, line 11: Add a test for empty crumbs array

ethanlook: No longer necessary since the crumbs prop is now `[string, ...string[]]`

mcous (approved):
Lgtm

packages/core/src/routes/+page.svelte
```diff
@@ -23,3 +24,6 @@ import Badge from '$lib/badge.svelte';
   variant="blue"
   label="Info"
 />
+
+<!-- Breadcrumbs -->
+<Breadcrumbs crumbs={['Chocolate Chip', 'Oatmeal Raisin']} />
```
micheal-parks, line 29: Maybe we should start splitting this into individual routes?

micheal-parks (approved):

packages/core/src/routes/+page.svelte
ethanlook: I could go either way, let's start here and split once it becomes unruly?

PR #287, opened 2023-07-27
inherit fill color from text-color

mrloureed:
Rather than explicitly setting the icon color every time, inherit from parent, or overwrite with text color.


micheal-parks (approved):
❤️

ethanlook (approved):

PR #292, opened 2023-07-31
Fix input events in select and multiselect

micheal-parks:
Fixes instances of both native and custom input events firing, causing state issues or errors to components that listen to these events due to the `detail` object only existing on custom events.

Note: this fixes camera issues we're seeing in RC.

### DIFF
--- a/packages/core/src/lib/button.svelte
+++ b/packages/core/src/lib/button.svelte
@@ -45,7 +45,7 @@ export let width: 'full' | 'default' = 'default';
   aria-disabled={disabled}
   {disabled}
   {title}
-  class={cx('whitespace-nowrap mx-auto', {
+  class={cx('mx-auto whitespace-nowrap', {
     'flex w-full': width === 'full',
     'inline-flex': width !== 'full' && variant !== 'icon',
     'h-[30px] w-[30px] text-gray-6 hover:text-gray-7 active:text-gray-8':
--- a/packages/core/src/lib/button.svelte
+++ b/packages/core/src/lib/button.svelte
@@ -45,13 +45,14 @@ export let width: 'full' | 'default' = 'default';
   aria-disabled={disabled}
   {disabled}
   {title}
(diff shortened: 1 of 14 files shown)


ehhong:
I don't entirely understand the bug -- was this able to be locally tested with app?

micheal-parks:
@ehhong The issue is that v-select and v-multiselect have hidden checkbox inputs that represent if an item within the selection list has been selected. These hidden inputs will fire `input` events, which will bubble up the DOM to whatever is listening for the `input` custom event that the these elements also fire. There's no reason to listen for these checkbox events so they're silenced by calling `stopPropagation`.

ehhong:
@micheal-parks why would this bug have prevented the manual refresh button from handling that event properly? because there was an exception thrown by these bubbled `input` events?

micheal-parks:
There are unrelated issues with full width buttons that are breaking checks. Not sure how they were merged but working on fixes.

ehhong:
> @ehhong The exception was thrown because a component that was listening to `input` was receiving events that did not contain the `details` object, which it was reading from for event data. `details` is only a feature of the `CustomEvent` class, which is what we're instantiating when firing our custom `input` event, and not the `Event` class, which is firing in native `input` events unless `stopPropagation` is called.

got it! thanks so much for the explanation

ehhong (approved):

mrloureed (approved):
LGTM

packages/core/src/lib/button.svelte
```diff
@@ -89,10 +90,3 @@ export let width: 'full' | 'default' = 'default';
     </span>
   {/if}
 </button>
-
-<style>
-:host {
-  --button-display: {display};
```
micheal-parks, line 95: This isn't valid svelte syntax and was breaking formatting in CI. Not sure how it passed ever. Anyway, there's a much simpler solution that doesn't involve setting styles on the host, and hosts don't even exist in a non-web-component context. Fix works in both core and legacy.

PR #294, opened 2023-07-31
Add storybook output to eslintignore and prettierignore

ethanlook:
Quick one found in development. The storybook output shouldn't get formatted or checked!


mrloureed (approved):
LGTM

PR #297, opened 2023-08-01
Bug fix: <v-switch> events

mrloureed:
This fix is related to https://github.com/viamrobotics/prime/pull/292

### DIFF
--- a/packages/legacy/src/elements/switch.svelte
+++ b/packages/legacy/src/elements/switch.svelte
@@ -26,7 +26,6 @@ $: isReadonly = htmlToBoolean(readonly, 'readonly');
 
 const handleClick = (event: Event) => {
   if (!(isDisabled || isReadonly)) {
-    console.log(event);
     value = on ? 'off' : 'on';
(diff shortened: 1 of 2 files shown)


packages/legacy/src/elements/switch.svelte
```diff
@@ -26,6 +26,7 @@ $: isReadonly = htmlToBoolean(readonly, 'readonly');
 
 const handleClick = (event: Event) => {
   if (!(isDisabled || isReadonly)) {
+    console.log(event);
```
micheal-parks, line 29: ✂️

micheal-parks (approved):

packages/legacy/src/elements/switch.svelte
mrloureed: 🪨

micheal-parks: 🧻

mrloureed: 💀

PR #298, opened 2023-08-01
Add tooltip

DTCurrie:
Adds the `<Tooltip />` to `core`. 

Notes:
- The layout for the component has been changed to be accessible. The major changes are there is no singular root component, and the default slot is now nested in an unstyled `<button />` element. This is so that the tooltip target can be focused using keyboard navigation, and so we can use proper `aria` syntax to assist with accessibility. 
- The `text` prop was removed, and now the tooltip text must be passed using the `slot="text"` syntax. This allows using rich html inside of tooltips and removes confusion on which to use.
- Removed the `<style />` block and instead relied on `tailwindcss` classes.
- Added unit tests, will add `e2e` tests for tooltip location in a follow-up that introduces `playwright` to `core`
  - Due to the nature of rendering with `@testing-library/svelte`, the tooltip has a `tooltip.spec.svelte` component to handler rendering slots
- Added a `svelte.config.js` to `storybook` to allow creating story view components with `svelte` and `typescript`
  - Due to the nature of rendering with `<Story />`, the tooltip has a multiple story view components to handler rendering slots
  
Testing:

You should be able to run `pnpm test` in `core` and see all tests passing, and `pnpm dev` to see the tooltip in the playground. Also you should be able to run `pnpm storybook` in `storybook` to test out the tooltip there.

Screenshot 2023-08-01 at 2 30 50 PM
Screenshot 2023-08-01 at 2 30 42 PM

### DIFF
--- a/packages/core/src/lib/tooltip.spec.svelte
+++ b/packages/core/src/lib/tooltip.spec.svelte
@@ -1,0 +1,30 @@
+<!-- 
+  @component
+
+  This component allows us to render a Tooltip with its slotted
+  children, due to limitations with rendering slots using
+  `@testing-library`.
+
+  See:
+  - https://sveltesociety.dev/recipes/testing-and-debugging/unit-testing-svelte-component
+  - https://github.com/testing-library/svelte-testing-library/issues/48
+ -->
+<script lang="ts">
+import Tooltip, { type TooltipLocation } from './tooltip.svelte';
+
+export let location: TooltipLocation = 'top';
+export let state: 'invisible' | 'visible' = 'invisible';
+</script>
+
+<Tooltip
+  {location}
+  {state}
+>
+  <p>This element has a tooltip.</p>
+  <span
+    slot="icon"
+    role="img"
+    class="information-outline"
+  />
+  <p slot="text">This is the tooltip text!</p>
+</Tooltip>
--- a/packages/core/src/lib/tooltip.svelte
+++ b/packages/core/src/lib/tooltip.svelte
@@ -1,0 +1,139 @@
+<!--
+  @component
+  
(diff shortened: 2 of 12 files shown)


packages/core/src/lib/tooltip.svelte
```diff
@@ -0,0 +1,156 @@
...
+>
+export type TooltipLocation = 'top' | 'bottom' | 'right' | 'left';
+export type TooltipState = 'visible' | 'invisible';
+</script>
+
+<script lang="ts">
+import { computePosition, flip, shift, offset, arrow } from '@floating-ui/dom';
+
+export let location: TooltipLocation = 'top';
+export let state: TooltipState = 'invisible';
```
ethanlook, line 29: Do you mind adding JSDoc comments to these props?

packages/core/src/lib/tooltip.svelte
```diff
@@ -0,0 +1,156 @@
...
+  w-max
+  max-w-[250px]
+  items-center gap-1
+  border
+  border-gray-9
+  bg-gray-9
+  px-2
+  py-1
+  text-left text-xs text-white
+`}
```
ethanlook, line 136: Did prettier make this multiline like this? Feels funny

packages/core/src/lib/tooltip.spec.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
```
ethanlook, line 1: Do you mind adding a lil docstring explaining the need for an extra test-only component?

ethanlook (approved):
This looks great, thank you!

ethanlook (approved):

PR #299, opened 2023-08-02
Add label to core

DTCurrie:
This PR adds a new element to `core`, the `Label`. This is the first step of migrating the `Input` to core, by first breaking it apart from its wrapping `Label` to simplify both components. This PR also includes fixes to the tooltip to make sure it works as expected inside of a `Label` (making sure the tooltip text is not read aloud as a part of the label text.)

### DIFF
--- a/packages/core/src/lib/label.svelte
+++ b/packages/core/src/lib/label.svelte
@@ -1,0 +1,33 @@
+<!-- 
+  @component
+  
+  For labelling inputs
+  
+```svelte
+<Label>
+  Name:
+  <Input name="name" />
+</Label>
+```
+ -->
+<svelte:options immutable />
+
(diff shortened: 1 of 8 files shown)


packages/core/src/lib/label.svelte
```diff
@@ -0,0 +1,57 @@
...
+  <span> Name:</span>
+  <Input slot="input" name="name" />
+</Label>
+```
+ -->
+<svelte:options immutable />
+
+<script
+  lang="ts"
+  context="module"
```
ethanlook, line 17: nice!

ethanlook (approved):
Nice, thank you!

ethanlook (approved):

PR #301, opened 2023-08-03
App 2290 migrate icon

mrloureed:
Add icons to Prime core package

image

https://viam.atlassian.net/browse/APP-2290


packages/core/package.json
```diff
@@ -73,5 +73,8 @@
   },
   "svelte": "./dist/index.js",
   "types": "./dist/index.d.ts",
-  "type": "module"
+  "type": "module",
+  "dependencies": {
+    "@mdi/js": "^7.2.96"
```
micheal-parks, line 78: Should be a devDependency

packages/core/src/lib/icon/icons.ts
```diff
@@ -0,0 +1,109 @@
+import {
```
micheal-parks, line 1: up for switching to a wildcard import here?

```ts
import * as MDI from from '@mdi/js';

export const paths: Record<string, string> = {
  alert: MDI.mdiAlert,
  ...

File: packages/legacy/src/elements/select/select.svelte [hunk 3]
```diff
@@ -271,10 +271,10 @@ $: {
     {#if tooltip}
       <v-tooltip text={tooltip}>
         <v-icon
-          class={cx('text-xs', {
+          class={cx('gextext-xs', {
```
micheal-parks, line 274: typo?

ethanlook (commented):
Do you mind adding the new story to packages/storybook?

packages/core/src/lib/icon.spec.ts
ethanlook, line 1: Could you please put this in the icon directory?

packages/core/src/lib/icon/icon.svelte
```diff
@@ -0,0 +1,69 @@
...
+  />
+```
+-->
+<svelte:options immutable />
+
+<script lang="ts">
+import cx from 'classnames';
+import { paths } from './icons';
+
+/**
```
ethanlook, line 18: Non-exported variables do not need a docstring unless it makes the code more readable. I think the docstring is unnecessary in the case of Size, hasNameProperty, and sizes, but it's ultimately up to you.

ethanlook (changes requested):
Do you mind adding a new story to package/storybook? We're also migrating the storybook as we go.

mrloureed:
> Do you mind adding a new story to package/storybook? We're also migrating the storybook as we go.

Added . Could def be improved but shows some icons . Can do more when I return from vaca.

ethanlook (approved):

PR #303, opened 2023-08-07
Add resource icons

ehhong:
These icons are needed for the new org settings UI to support location/robot permissions


anjinai:
> These icons are needed for the new org settings UI to support location/robot permissions

[q] - do you need to up your prime version?

anjinai (approved):
✅

PR #304, opened 2023-08-07
APP-2297: migrate radio to viamrobotics prime core

anjinai:


packages/core/src/lib/radio.svelte
```diff
@@ -0,0 +1,145 @@
...
+        </div>
+      </Tooltip>
+    {/if}
+  </div>
+
+  <div class="flex flex-nowrap">
+    {#each options as option}
+      <button
+        aria-label="Select {option}"
+        class={cx('whitespace-nowrap border px-3 py-1.5 text-xs', {
```
anjinai, line 128: So storybook is giving me an accessibility violation on readonly radio.  Should I reach out to Nat and ask if she wants to tweak the style?  Or is this fine as is?

Screen Shot 2023-08-08 at 8 15 05 PM

ethanlook: This seems like an opportunity for improvement! I'd reach out to Nat and we can go from there

packages/core/src/lib/radio.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  it('Allows option click if not readonly', async () => {
+    const { component } = render(Radio, {
+      options: ['Option1', 'Option2'],
+      selected: 'Option1',
+    });
+    const onInput = vi.fn();
+    component.$on('input', onInput);
+    await fireEvent.click(screen.getByText('Option2'));
+    expect(onInput).toHaveBeenCalledOnce();
```
ethanlook, line 38: Check here that onInput was called with Option2?

packages/core/src/lib/radio.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  it('Applies correct color for warn state', () => {
+    render(Radio, { state: 'warn', tooltip: 'Strong warning' });
+    const icon = screen.getByRole('button');
+    expect(icon.firstChild).toHaveClass('text-warning-bright');
+    const svg = screen.getByRole('img', { name: 'alert-circle-outline icon' });
+    expect(svg).toBeVisible();
+  });
+
+  it('Applies correct color for error state', () => {
+    render(Radio, { state: 'error', tooltip: 'Ahhhhh error!' });
```
ethanlook, line 86: Ahhhhhhh! 🤣

packages/core/src/routes/+page.svelte
ethanlook, line 1: Thanks for cleaning these up!

ethanlook (approved):
Nice!! Could you please bump the version of prime-core as well?

packages/core/src/lib/radio.spec.ts
anjinai: 😂

PR #310, opened 2023-08-09
APP-2295 - Notify core

Kschappacher:


packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,104 @@
...
+      {#if variant === 'danger'}
+        <v-icon size="lg" class="text-danger-dark" name="alert-circle" />
+      {:else if variant === 'info'}
+        <v-icon size="lg" class="text-info-dark" name="information" />
+      {:else if variant === 'success'}
+        <v-icon size="lg" class="text-success-dark" name="check-circle" />
+      {/if}
+
+      {#if variant === 'warning'}
+        <v-icon size="lg" class="text-warning-bright" name="alert" />
```
anjinai, line 67: Now that Icon has been migrated to prime/core, can you use that instead of v-icon?

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,104 @@
...
+          {#if $$slots.action}
+            <div class="pt-1 pb-2">
+              <slot name="action" />
+            </div>
+          {/if}
+        </div>
+      </figure>
+
+      {#if isExitable}
+        <v-button
```
anjinai, line 91: Now that button has been migrated to prime/core, can you use button instead of v-button?

packages/core/src/lib/notify.svelte
Kschappacher: changed to Icon!

packages/core/src/lib/notify.svelte
Kschappacher: Using Button Now!

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,132 @@
...
+import { createEventDispatcher } from 'svelte';
+import { Button, Icon } from '$lib';
+
+export let title = '';
+export let message = '';
+export let variant: Variants = 'info';
+export let progress = 1;
+export let exitable = 'false';
+
+export const htmlToBoolean = (value: string | undefined, key: string) => {
```
anjinai, line 16: We can now accept boolean props so can you remove the htmlToBoolean?

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,132 @@
...
+            </div>
+          {/if}
+        </div>
+      </figure>
+
+      {#if isExitable}
+        <!-- svelte-ignore a11y-unknown-role -->
+        <div
+          class="absolute right-1 top-1 text-gray-7"
+          role="notify"
```
anjinai, line 118: [q] - is notify a valid aria role?

Kschappacher: changed to alert

packages/core/src/lib/notify.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    render(Notify, {
+      title: 'This is the title',
+      message: 'This is the message.',
+    });
+    expect(screen.getByText('This is the title')).toBeVisible();
+    expect(screen.getByText('This is the message.')).toBeVisible();
+  });
+});
+
+describe('Notify', () => {
```
ethanlook, line 17: Could you please put all `it` statements in the same `describe`?

describe('Notify', () => {
  it('...', () => {...});
  it('...', () => {...});
  it('...', () => {...});
});

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,127 @@
...
+
+import cx from 'classnames';
+import { createEventDispatcher } from 'svelte';
+import { Button, Icon } from '$lib';
+
+export let title = '';
+export let message = '';
+export let variant: Variants = 'info';
+export let progress = 1;
+export let exitable = false;
```
ethanlook, line 14: Could you please add docstrings to all props? [Example](https://github.com/viamrobotics/prime/blob/main/packages/core/src/lib/button.svelte#L15-L39)

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,127 @@
+<svelte:options immutable />
+
+<script lang="ts">
+type Variants = 'danger' | 'warning' | 'success' | 'info';
```
ethanlook, line 4: Can you inline this type on L12? It will make IDE suggestions easier to use

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,127 @@
...
+
+export let title = '';
+export let message = '';
+export let variant: Variants = 'info';
+export let progress = 1;
+export let exitable = false;
+
+$: isExitable = exitable;
+
+const dispatch = createEventDispatcher();
```
ethanlook, line 18: Can you please add typings and docstrings? [Example](https://github.com/viamrobotics/prime/blob/main/packages/core/src/lib/pill.svelte#L26-L29)

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,127 @@
...
+import { createEventDispatcher } from 'svelte';
+import { Button, Icon } from '$lib';
+
+export let title = '';
+export let message = '';
+export let variant: Variants = 'info';
+export let progress = 1;
+export let exitable = false;
+
+$: isExitable = exitable;
```
anjinai, line 16: I think you shouldn't need isExitable at all now and you can just use exitable.

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,127 @@
...
+
+const dispatch = createEventDispatcher();
+
+const handleClose = (event: Event) => {
+  dispatch('close', event);
```
ethanlook, line 21: There's no need to pass the event as the value - just call `dispatch('close')`.

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,127 @@
...
+            <p class="text-sm text-subtle-1">{message}</p>
+          {/if}
+
+          <slot />
+
+          {#if $$slots.action}
+            <div class="pb-2 pt-1">
+              <slot name="action" />
+            </div>
+          {/if}
```
ethanlook, line 106: Do you have a test for this slot?

ethanlook (changes requested):
Awesome stuff! Do you mind also bumping the version in `core/packages.json`?

anjinai:
Can you write storybook stories too?

packages/core/src/lib/notify.svelte
Kschappacher: yupp

packages/core/src/lib/notify.svelte
Kschappacher: removed and am now just using exitable

packages/core/src/lib/notify.svelte
Kschappacher: the linter didnt like dispatch('close'). Error: Expected 2-3 arguments, but got 1.ts(2554)

ethanlook: Ah, I'd do `dispatch('close', {})` and define the dispatcher as:

```
const dispatch = createEventDispatcher<{
  close: {}
}>();
```

packages/core/src/lib/notify.svelte
Kschappacher: what do you mean by this?

packages/core/src/lib/notify.svelte
Kschappacher: yupp updated the name

Kschappacher:
> Can you write storybook stories too?

just added the storybook to the pr

packages/storybook/src/stories/notify/notify.stories.mdx
```diff
@@ -0,0 +1,192 @@
...
+    },
+    title: {
+      description: 'Title',
+      table: { defaultValue: { summary: '' } },
+    },
+    message: {
+      description: 'Message',
+      table: { defaultValue: { summary: '' } },
+    },
+  }}
```
anjinai, line 27: [opt] You could add an argType for your close event similar to how Ethan does it in the pill story he wrote:

Something like:

`'on:close': {
      description: 'Event fired when user presses close button.',
    },`

anjinai (approved):
LGTM

anjinai:
> LGTM

Can you bump the version number up in `packages/core/package.json`?

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.8",
+  "version": "0.0.9",
```
ethanlook, line 3: `prime-core` is already on [`0.0.9` now](https://github.com/viamrobotics/prime/blob/main/packages/core/package.json#L3) - please bump to `0.0.10`!

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,133 @@
...
+import cx from 'classnames';
+import { createEventDispatcher } from 'svelte';
+import { Button, Icon } from '$lib';
+
+/** The name of the notification you want to alert users about*/
+export let title = '';
+/** The specific details about your notifications*/
+export let message = '';
+/** The severity of the notification you want to show users*/
+export let variant: Variants = 'info';
```
ethanlook, line 15:
```suggestion
export let variant: 'danger' | 'warning' | 'success' | 'info' = 'info';
```

packages/core/src/lib/notify.svelte
ethanlook: See below!

packages/core/src/lib/notify.svelte
```diff
@@ -0,0 +1,133 @@
...
+/** The scaling applied on the y access of the page*/
+export let progress = 1;
+/** If true a user can close out of the notification. If false they cannot*/
+export let exitable = false;
+
+const dispatch = createEventDispatcher<{
+  /** when the notification close button is pressed*/
+  close: object;
```
ethanlook, line 23:
```suggestion
  close: Record<string, never>;
```

packages/core/src/routes/+page.svelte
ethanlook, line 1: Could you please wrap each in a div like the other components? Hopefully that will help keep this file organized.

ethanlook (approved):

packages/core/src/lib/notify.svelte
Kschappacher: Makes sense 👍

packages/core/src/lib/notify.svelte
Kschappacher: compiler wont let me make this change. it says Event is not type string

ethanlook: Let me know if you'd like to look together!

PR #311, opened 2023-08-09
Add Input and TextInput

DTCurrie:
Adds the base `<Input />` and `<TextInput />` components to `core`. Does not add stories yet, holding off until all inputs are done to do a more comprehensive story with docs/usage notes. Also did some cleanup on the `<Tooltip />` to remove the `icon` slot and just expect it to be included as a part of the default slot.

### DIFF
--- a/packages/core/src/lib/index.ts
+++ b/packages/core/src/lib/index.ts
@@ -5,8 +5,15 @@ export { default as Icon } from './icon/icon.svelte';
 export { default as Label } from './label.svelte';
 export { default as Pill } from './pill.svelte';
 export { default as Radio } from './radio.svelte';
+
 export {
   default as Tooltip,
   type TooltipLocation,
   type TooltipState,
 } from './tooltip.svelte';
+
+export { default as Input } from './input/input.svelte';
+export {
+  default as TextInput,
+  type TextInputTypes,
+} from './input/text-input.svelte';
--- a/packages/core/src/lib/input/input.svelte
+++ b/packages/core/src/lib/input/input.svelte
@@ -1,0 +1,55 @@
+<!--
+@component
+  
+For user inputs.
+
+This is the base input component that accepts all input properties without any 
+additional behaviors attached. Generally, other typed inputs like the 
+NumberInput or DateInput are preferable.
+
+```svelte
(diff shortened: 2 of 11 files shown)


packages/core/src/lib/index.ts
```diff
@@ -5,8 +5,15 @@ export { default as Icon } from './icon/icon.svelte';
 export { default as Label } from './label.svelte';
 export { default as Pill } from './pill.svelte';
 export { default as Radio } from './radio.svelte';
+
```
ethanlook, line 8: Why the extra newlines?

packages/core/src/lib/input/input.svelte
```diff
@@ -0,0 +1,57 @@
...
+import cx from 'classnames';
+
+/** Whether or not the input should be rendered as readonly. */
+export let readonly = false;
+
+/**
+ * Whether or not the input should be rendered as readonly and have
+ * `aria-disabled` set to true.
+ */
+export let disabled = false;
```
ethanlook, line 26: Is there any reason to have separate props? It looks like these do the same things.

ethanlook (approved):

anjinai (approved):
✅

packages/core/src/lib/input/input.svelte
DTCurrie: Yes, mainly for accessibility. For readonly inputs, we generally still allow things like selecting and copying. We don't want to indicate to screen readers those inputs are disabled or inactionable. For disabled inputs, we generally do not allow any actions on the inputs, but they still need to be focusable so screen readers can see them. If we make the input disabled, it won't be focusable or perceivable to assistive devices. So we make it readonly so it can still be perceived, then set `aria-disabled` to indicate to them there is not any action they can take on the input other than being aware of it.

We could go a step further for disabled and assign a specific cursor as well, I went back and forth on that but may be useful for non-screenreader users to differentiate.

DTCurrie: @ethanlook OK so now `disabled` also sets `pointer-events-none` making it non-operable. I updated the prop docs to clarify: 

readonly = renders as readonly but is still operable
disabled = renders as readonly but is not operable

PR #319, opened 2023-08-11
App 2284 core button cleanup

mrloureed:
Cleanup core button component:

- break logic up into buttons with icons VS without icons
- update logic appropriately


ethanlook (approved):
LGTM, but I have a really nit-picky question: would it make more sense to call it an `IconButton`? To me, `ButtonIcon` sounds like it's an icon of a button, but `IconButton` sounds like a button with an icon.

PR #320, opened 2023-08-14
Add numeric and slider inputs

DTCurrie:
Also includes updates to the input and label to match designs here: https://www.figma.com/file/ERHSYDGTXiEQOVGY8ORg6I/PRIME?node-id=3149%3A5823&mode=dev


packages/core/src/lib/input/numeric-input.svelte
```diff
@@ -0,0 +1,87 @@
...
+
+export let value: number | undefined = undefined;
+
+export let step = 1;
+
+const dispatch = createEventDispatcher<{
+  input: number;
+  blur: void;
+  keydown: number;
+}>();
```
ethanlook, line 28: nit: docstrings, please!

packages/core/src/lib/input/slider-input.svelte
```diff
@@ -0,0 +1,157 @@
...
+export let step = 1;
+
+export let min = -Infinity;
+export let max = Infinity;
+
+const dispatch = createEventDispatcher<{
+  input: number;
+  blur: void;
+  keydown: number;
+}>();
```
ethanlook, line 36: nit: docstrings, please!

packages/core/src/lib/input/slider-input.svelte
```diff
@@ -0,0 +1,157 @@
...
+  value = next;
+  dispatch('input', value);
+
+  // The tooltip may not be mounted the first time this handler is called.
+  numberDragTooltip.recalculateStyle();
+};
+
+const handleNumberDragUp = () => {
+  isDragging = false;
+  window.removeEventListener('pointermove', handleNumberDragMove);
```
ethanlook, line 98: Looks like the pointerup listener is not removed

packages/core/src/lib/tooltip.svelte
```diff
@@ -47,7 +50,7 @@ let x = 0;
 let y = 0;
 let arrowCss = '';
 
-export let recalculateStyle = async () => {
+export const recalculateStyle = async () => {
```
ethanlook, line 53: I was noticing that the tooltip is missing the arrow - does this resolve that?

ethanlook (approved):

packages/core/src/lib/input/numeric-input.svelte
DTCurrie: This is a draft PR, I will add tests and docs later I just put this up so you can grab the state code from the input.

packages/core/src/lib/input/slider-input.svelte
DTCurrie: It is set to only fire once, not sure if it is necessary to remove it: https://github.com/viamrobotics/prime/pull/320/files#diff-159329097f01d3404df5ec038d4ca2a7beeac4648d1595cdbe32f70860351958R118

packages/core/src/lib/input/input.svelte
```diff
@@ -13,33 +13,92 @@ NumberInput or DateInput are preferable.
...
+    case 'error': {
+      icon = 'alert-circle';
+      break;
+    }
+    default: {
+      icon = '';
+      break;
+    }
+  }
+}
```
micheal-parks, line 62: this is a cool pattern that's a bit more compact:
```suggestion
$: icon = {
  info: 'information',
  warn: 'alert'
  error: 'alert-circle'
}[state] ?? ''
```
but either works for me!

packages/core/src/lib/input/numeric-input.svelte
```diff
@@ -0,0 +1,91 @@
...
+  
+For numeric user inputs.
+
+```svelte
+<NumericInput type="integer"  on:input={onInput} />
+```
+-->
+<svelte:options
+  immutable
+  accessors
```
micheal-parks, line 12: How are we using accessors?

packages/core/src/lib/input/numeric-input.svelte
```diff
@@ -0,0 +1,91 @@
...
+$: isNumber = type === 'number';
+$: pattern = isNumber ? '^([-+,0-9.]+)' : '[0-9]+';
+
+const setValue = (next: number) => {
+  value = next;
+  stringValue = `${value}`;
+  dispatch('input', next);
+};
+
+const handleKeydown = (event: KeyboardEvent) => {
```
micheal-parks, line 55: Isn't this behavior native to the number input since we're using it now?

packages/core/src/lib/tooltip.svelte
```diff
@@ -12,7 +12,10 @@
   </Tooltip>
   ```
 -->
-<svelte:options immutable />
+<svelte:options
+  immutable
+  accessors
```
micheal-parks, line 17: How are we using accessors here as well?

packages/core/src/lib/input/numeric-input.svelte
```diff
@@ -0,0 +1,91 @@
...
+ * these props are treatef as required and force users to add value={undefined}
+ * when no initial value is set.
+ */
+// eslint-disable-next-line no-undef-init,unicorn/no-useless-undefined
+export let value: number | undefined = undefined;
+
+/** The amount to increment/decrement when using the up/down arrows. */
+export let step = 1;
+
+const dispatch = createEventDispatcher<{
```
micheal-parks, line 40: I think we may also need blur

PR #323, opened 2023-08-14
APP-2302 Migrate table to prime-core

ehhong:
### Questions
- I don't love the exported names that follow the capitalization convention: 
    - `v-thead` --> `THead`
    - `th` --> `TH`
    - any ideas for what we should do instead?
- Should `tr` variant colors be updated?
- the `style` attribute on these components feels outdated, but `class` is a reserved word so cannot be exported as an attribute
- lint error for using `array.entries()` -- not sure how best to resolve?
```
Type 'IterableIterator<[number, string]>' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.
```

### Changes
- Fixed + added some documentation for context menu

### DIFF
--- a/packages/core/src/lib/index.ts
+++ b/packages/core/src/lib/index.ts
@@ -31,8 +31,8 @@ export {
 } from './input/text-input.svelte';
 
 export { default as Table } from './table/table.svelte';
-export { default as THead } from './table/thead.svelte';
-export { default as TBody } from './table/tbody.svelte';
-export { default as TD } from './table/td.svelte';
-export { default as TH } from './table/th.svelte';
-export { default as TR } from './table/tr.svelte';
+export { default as TableHeader } from './table/table-header.svelte';
+export { default as TableHeaderCell } from './table/table-header-cell.svelte';
+export { default as TableBody } from './table/table-body.svelte';
+export { default as TableRow } from './table/table-row.svelte';
+export { default as TableCell } from './table/table-cell.svelte';
--- a/packages/core/src/lib/index.ts
+++ b/packages/core/src/lib/index.ts
@@ -29,3 +29,10 @@ export {
   default as TextInput,
   type TextInputTypes,
 } from './input/text-input.svelte';
+
+export { default as Table } from './table/table.svelte';
+export { default as THead } from './table/thead.svelte';
+export { default as TBody } from './table/tbody.svelte';
+export { default as TD } from './table/td.svelte';
+export { default as TH } from './table/th.svelte';
+export { default as TR } from './table/tr.svelte';
--- a/packages/core/src/lib/table/table-body.svelte
+++ b/packages/core/src/lib/table/table-body.svelte
@@ -13,8 +13,9 @@ A table body.
 <script lang="ts">
 /**
  * Custom table body style
- * @example
- * style="border-collapse"
+ * ```
+ * <TableBody style="text-align:center">
+ * ```
  */
 export let style = '';
 </script>
--- a/packages/core/src/lib/table/table-body.svelte
+++ b/packages/core/src/lib/table/table-body.svelte
@@ -1,11 +1,10 @@
 <!--
 @component
   
-A table header.
+A table body.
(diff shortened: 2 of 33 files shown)


packages/core/src/lib/index.ts
```diff
@@ -19,6 +19,14 @@ export {
...
 export { default as ContextMenu } from './context-menu/context-menu.svelte';
 export { default as ContextMenuItem } from './context-menu/context-menu-item.svelte';
 export { default as ContextMenuSeparator } from './context-menu/context-menu-separator.svelte';
+
+export { default as Table } from './table/table.svelte';
+export { default as THead } from './table/thead.svelte';
+export { default as TBody } from './table/tbody.svelte';
+export { default as TD } from './table/td.svelte';
+export { default as TH } from './table/th.svelte';
+export { default as TR } from './table/tr.svelte';
```
ehhong, line 32: looking for better ideas for these exported names 👀

packages/core/src/lib/table/table.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    render(Table, { variant: 'fixed' });
+    expect(screen.getByRole('table')).toBeVisible();
+    expect(screen.getByRole('table')).toHaveClass('table-fixed');
+  });
+
+  it('Renders columns with widths set by cols attribute', () => {
+    const cols = ['20%', '30%', '50%'];
+    const { container } = render(TableWith3Cols, { cols });
+
+    for (const [i, width] of cols.entries()) {
```
ehhong, line 29: this is recommended by the linter but using `cols.entries()` is firing a different lint error about JS versions

packages/core/src/lib/index.ts
ethanlook: One of my favorite UI libraries is [Material UI](mui.com). I like how they name their table components - `Table`, `TableBody`, `TableRow`, etc: https://mui.com/material-ui/react-table/#api

packages/core/src/lib/table/table-body.svelte
```diff
@@ -0,0 +1,30 @@
...
+</Table>
+```
+-->
+<svelte:options immutable />
+
+<script lang="ts">
+/**
+ * Custom table body style
+ * @example
+ * style="border-collapse"
```
ethanlook, line 17: How does the intellisense look with this example? Should it be wrapped in a ``` block?

packages/core/src/lib/table/table.spec.ts
ethanlook: Interesting! What's the error? Let me know if you want to take a look together.

packages/core/src/lib/table/table.svelte
```diff
@@ -0,0 +1,60 @@
+<!--
+@component
+  
+A table.
+```svelte
+<Table variant='fixed' cols=["30%", "30%", "40%"]>
```
ethanlook, line 6: How does the fixed table behave when contents overflow a cell?

ethanlook (approved):
Looks great! Thank you! Could you please update the prime-core package version as well!

packages/core/src/lib/table/table.spec.ts
ehhong: the error is:
```
Type 'IterableIterator<[number, string]>' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.
```

not sure which tag we should use to resolve -- what version of JS are we using?

packages/core/src/lib/table/table-body.svelte
ehhong: with @example
image

with ```
image

updating everything to ``` -- it looks cleaner

packages/core/src/lib/table/table.svelte
ehhong: it wraps to maintain the widths, this example is with widths 10% and 90%
image

PR #325, opened 2023-08-15
APP-2287: Migrate collapse to @viamrobotics/prime-core

anjinai:
Migrates collapse component to `@viamrobotics/prime-core`.

## Change log

- Copied `collapse.svelte` to `packages/core.`
- Removed need for htmlToBoolean - uses boolean for open.
- Updated unit tests.
- Copied storybook story to `packages/storybook` and added missing stories.
- Bumped prime core version.

### Notes

- **I'm finding it difficult to properly show collapse's slot usage in storybook and given the point of storybook is to demonstrate how to use a component, my stories should be better.  Do you happen to have any advice?**

Screen Shot 2023-08-17 at 3 19 20 PM

## Testing

- Added unit tests

## Review requests

- None

[APP-2287]: https://viam.atlassian.net/browse/APP-2287


packages/core/src/lib/collapse.svelte
```diff
@@ -0,0 +1,68 @@
...
+      {/if}
+
+      <slot name="title" />
+    </div>
+
+    <div class="h-full flex items-center gap-3">
+      <slot name="header" />
+
+      <div
+      class={cx(
```
anjinai, line 47: I would expect this to work/rotate my icon, but it's not.  Do I need to update the icon svelte component to accept class props?

packages/core/src/lib/collapse.svelte
```diff
@@ -0,0 +1,94 @@
...
+    </div>
+  </div>
+
+  <div
+    class={cx('text-black transition-all duration-500', {
+      'bg-white': variant === 'default',
+      hidden: !open,
+    })}
+  >
+    <slot />
```
micheal-parks, line 92: I'd like to change how we handle this slot as we revise `<Collapse>` - specifically moving to conditionally rendering this slot rather than making it hidden. Before we had prime-core I created a revised version of collapse [here](https://github.com/viamrobotics/rdk/blob/main/web/frontend/src/lib/components/collapse.svelte).

Would you be up for moving the slot changes here? I'd like to work toward removing most of the RDK component and replacing it with this one.

The optimization is mostly for performance; rendering fewer nodes, especially if they're not used, almost always means better performance.

micheal-parks: There is also some local storage stuff but I have doubts that it belongs in Prime, we may just have to make a wrapper component for <Collapse> in RDK that persists state.

anjinai: Absolutely, will update accordingly

packages/core/src/lib/collapse.svelte
```diff
@@ -0,0 +1,95 @@
...
+```
+-->
+<svelte:options immutable />
+
+<script lang="ts">
+import cx from 'classnames';
+import { createEventDispatcher } from 'svelte';
+import { Icon } from '$lib';
+
+type Variants = 'default' | 'minimal';
```
micheal-parks, line 24: I'd recommend getting rid of these variants, we don't use minimal anywhere and design doesn't recognize it

packages/core/src/lib/collapse.svelte
```diff
@@ -0,0 +1,95 @@
...
+      )}
+    >
+      <Icon name="chevron-down" />
+    </div>
+    </div>
+  </div>
+
+  {#if open}
+    <div
+      class={cx('text-black transition-all duration-500', {
```
micheal-parks, line 88: Would recommend just rendering the slot and not this wrapping div so that the slot content can remain flexible

packages/core/src/routes/+page.svelte
```diff
@@ -87,6 +88,40 @@ let buttonClickedTimes = 0;
     />
   </div>
 
+
+  <div class="flex gap-4">
+    <!-- Collapse -->
+    <Collapse
+      title='Motor 1'
+    >
+      <div style='font-size: 12px; padding: 1rem; border: 1px solid #e4e4e6; border-top: none;'>Motor one was concieved and executed at Bell Labs in 1972 under the guidance of lead director Dennis Richie and Superviser Wallace Breen.</div>
```
micheal-parks, line 97: Can we use tailwind here?

packages/storybook/src/stories/collapse.spec.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+  import { Collapse } from '@viamrobotics/prime-core';
+  export let title = '';
+  export let content = '';
+</script>
+
+
+
+<Collapse title={title}>
+  <div slot="content" style="font-size: 12px; padding: 1rem; border: 1px solid #e4e4e6; border-top: none;">
```
micheal-parks, line 10: Same tailwind question as above

packages/core/src/lib/collapse.svelte
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Whether the collapse is in the open position.
+ */
+export let open = false;
+/**
+ * The variant of the collapse component.
+ */
+export let variant: Variants = 'default';
+
+const dispatch = createEventDispatcher();
```
micheal-parks, line 39:
```suggestion
type Events = {
  /** Fires when the collapse is toggled */
  toggle: boolean
}

const dispatch = createEventDispatcher<Events>();

Comment (ethanlook) [reply packages/core/src/lib/collapse.svelte]: We should update all components to this event dispatcher style.

File: packages/core/src/lib/collapse.spec.svelte [hunk 8]
```diff
@@ -0,0 +1,47 @@
...
+    <Collapse
+      title='Number 3'
+    >
+      <Breadcrumbs slot='title' crumbs={['Hi', 'Hello']}></Breadcrumbs>
+      <Badge slot='header' label='Active'></Badge>
+      <div class="text-sm p-4 border border-t-0 border-light">Three is company.</div>
+    </Collapse>
+
+    <Collapse
+      title='Nummber 4'
```
ethanlook, line 41: Misspelled number :)

packages/core/src/lib/collapse.svelte
```diff
@@ -0,0 +1,82 @@
...
+        }
+      )}
+    >
+      <Icon name="chevron-down" />
+    </div>
+    </div>
+  </div>
+
+  {#if open}
+      <slot />
```
ethanlook, line 80: I think if we have multiple slots, like this component does, all of the slots should have names. What do you think about adding the slot name "content" here?

ethanlook (approved):
Clean and simple! I like it!

ethanlook:
> I'm finding it difficult to properly show collapse's slot usage in storybook and given the point of storybook is to demonstrate how to use a component, my stories should be better.  Do you happen to have any advice?

I agree, but we can leave it as-is for now. I filed [APP-2404](https://viam.atlassian.net/browse/APP-2404) to fix this exact problem.

PR #326, opened 2023-08-15
Add navigation map to prime-blocks

micheal-parks:
This PR adds all necessary shared navigation-ui for the navigation service card in app.viam.com / navigation card in RC.

Two new APIs were introduced. One for a generic maplibre component:
```svelte
<MapLibre>
  <MapLibreMarker lngLat={{ lng: 0, lat: 0 }} />
</MapLibre>
```

nav-1

And one for the navigation service map:
```svelte
<NavigationMap
  mode={'debug' | 'configure'}
  waypoints={Waypoint[]}
  obstacles={Obstacle[]}
  on:click
  on:delete-waypoint
  on:create-obstacle
  on:delete-obstacle
  on:move-obstacle
  on:change-obstacle-geometry
/>
```

nav-2


packages/blocks/src/lib/maplibre/index.svelte
```diff
@@ -0,0 +1,129 @@
...
+
+  Children will mount once the map is fully loaded.
+
+  ```svelte
+    <MapLibre>
+      <MapLibreMarker lngLat={{ lng: 0, lat: 0 }} />
+    </MapLibre>
+  ```
+-->
+<script lang='ts'>
```
DTCurrie, line 14: Should we be adding?

```svelte
<svelte:options immutable />
```

micheal-parks: Done!

packages/blocks/.eslintrc.cjs
```diff
@@ -13,6 +13,7 @@ module.exports = {
   },
   rules: {
     // @todo(mp APP-2482): move to js-config
+    'sonarjs/no-empty-collection': 'off',
```
micheal-parks, line 16: Not compatible with svelte binding elements to an array (used in vectorInput)

DTCurrie (approved):
✅

packages/blocks/src/lib/maplibre/hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+  MapLayerMouseEvent,
+  MapLayerTouchEvent,
+} from 'maplibre-gl';
+import { type Writable, get } from 'svelte/store';
+
+/**
+ * Provides context for a <MapLibre> instance. Must be called within a child of this component.
+ */
+export const useMapLibre = () => {
+  const mapContext = getContext<Writable<Map> | undefined>('map');
```
mcous, line 15: We've been using [`Symbol`s](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) with descriptive names for context keys to avoid potential collisions. e.g.

```ts
const MapContextKey = Symbol('prime-blocks-map')

// ...

const mapContext = getContext<....>(MapContextKey)
```

packages/blocks/src/lib/maplibre/hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+    throw new Error(
+      'useMapLibre is a context sensitive hook that must be used inside a <MapLibre> component.'
+    );
+  }
+
+  const map = get(mapContext);
+  const mapCenter = getContext<Writable<LngLat>>('center');
+  const mapSize =
+    getContext<Writable<{ width: number; height: number }>>('size');
+  const mapZoom = getContext<number>('zoom');
```
mcous, line 27: Can we combine these three contexts (map, size, and zoom) into a single context with multiple stores?

packages/blocks/src/lib/maplibre/index.svelte
mcous: Are we sure? `map` looks to be a mutable value that we use in reactive statements

packages/blocks/src/lib/maplibre/marker.svelte
```diff
@@ -0,0 +1,41 @@
...
+import { useMapLibre } from './hooks';
+
+export let lngLat: LngLatLike;
+export let scale = 1;
+export let visible = true;
+export let color = '#ff0047';
+
+const { map } = useMapLibre()
+
+const marker = new Marker({ scale, color });
```
mcous, line 28: What happens if the `scale` or `color` props change?

packages/blocks/src/lib/navigation-map/components/center-inputs.svelte
```diff
@@ -0,0 +1,20 @@
...
+
+const { map, mapCenter } = useMapLibre();
+
+const handleInput = (event: CustomEvent<LngLat>) => {
+  map.jumpTo({ center: event.detail })
+}
+
+</script>
+
+<div class='absolute right-10 flex flex-wrap gap-y-2 items-end justify-between py-2.5 w-60 px-4 z-[1000]'>
```
mcous, line 14: Should `z-[1000]` be added to the tailwind config as something more descriptive? `z-max` or similar so it's intent is a little more clear?

packages/core/src/lib/input/slider-input.svelte
```diff
@@ -94,6 +99,7 @@ const handlePointerMove = (event: PointerEvent) => {
   if (value !== next) {
     value = next;
     dispatch('input', value);
+    dispatch('slide', value);
```
mcous, line 102: Do we need this non-standard `slide` event if it's the same value as the standard `input`?

packages/core/package.json
```diff
@@ -17,6 +17,7 @@
     "test": "svelte-kit sync && vitest run",
     "test:watch": "vitest"
   },
+  "module": "./dist/index.js",
```
mcous, line 20: This seems suspicious, since `module` is a non-standard field and has been supplanted by `exports`. Why was this needed? It seems wrong, because `dist/index.js` (to my knowledge) cannot be imported in a non-Svelte project

packages/blocks/src/routes/+page.svelte
```diff
@@ -12,8 +13,23 @@ const fetchPointcloud = async () => {
...
+      {/await}
+    </div>
+  </div>
+
+  <div class='px-12'>
+    <div class="relative w-full pt-0 border border-gray-200 aspect-video">
+      <MapLibre />
+    </div>
+  </div>
+</div>
```
mcous, line 35:
```suggestion
</div>

```

packages/blocks/src/routes/+page.svelte
```diff
@@ -12,8 +13,23 @@ const fetchPointcloud = async () => {
...
-  {#await fetchPointcloud() then pointcloud}
-    <SlamMap2D {pointcloud} {motionPath} />
-  {/await}
-</div>
+<div class='m-auto max-w-6xl flex flex-col gap-6 py-6'>
+  <NavigationMap />
+
+  <div class='px-12'>
+    <div class='relative w-full border border-gray-200 aspect-video'>
+      {#await fetchPointcloud() then pointcloud}
```
mcous, line 21: Do we need error handling for this promise?

packages/blocks/src/routes/+page.svelte
```diff
@@ -1,8 +1,9 @@
 <script lang='ts'>
 
-import { SlamMap2D } from '$lib';
 import pointcloudUrl from '$lib/assets/pointcloud.pcd?url'
 import motionPath from '$lib/assets/cbirrtsmooth800.txt?raw'
+import { MapLibre, SlamMap2D } from '$lib';
+import NavigationMap from './navigation-map.svelte';
 
 const fetchPointcloud = async () => {
   const response = await fetch(pointcloudUrl)
```
mcous, line 9: Do we need non-2xx response checking for this request?

packages/blocks/src/lib/navigation-map/components/input/geometry.svelte
```diff
@@ -0,0 +1,85 @@
...
+        geometry.radius = x;
+      }
+      if (y) {
+        geometry.length = y;
+      }
+      break;
+    }
+  }
+
+  dispatch('input', geometry);
```
mcous, line 51: - Is this going to miss changes when `x === 0`, `y === 0`, or `z === 0`?
- Given that we `dispatch` the updated object, we should not be mutating a passed-in prop like this. It could pull the rug out from under the parent component or another sibling that received the same `Geometry` object

```suggestion
  const [x, y, z] = event.detail;
  let nextGeometry: Geometry
  
  switch (geometry.type) {
    case 'box': {
      nextGeometry = {
        ...geometry,
        length: x ?? geometry.length,
        width: y ?? geometry.width,
        height: z ?? geometry.height,
      }
      break;
    }
    case 'capsule': {
      nextGeometry = {
        ...geometry, 
        radius: x ?? geometry.radius,
        length: y ?? geometry.length,
      }
      break;
    }
    case 'sphere': {
      nextGeometry = {
        ...geometry,
        radius: x ?? geometry.radius,
      }
      break;
    }
  }

  dispatch('input', nextGeometry);
```

packages/blocks/src/lib/maplibre/index.svelte
```diff
@@ -0,0 +1,142 @@
...
+const dispatch = createEventDispatcher<Events>();
+const mapStore = writable<Map | undefined>(undefined);
+const centerStore = writable<LngLat>(center);
+const sizeStore = writable({ width: 0, height: 0 });
+const zoomStore = writable(zoom);
+
+setContext('map', mapStore);
+setContext('center', centerStore);
+setContext('size', sizeStore);
+setContext('zoom', zoomStore);
```
mcous, line 62: Follow up on other comment: I think we should create all these stores and call `setContext` in a `provideMapLibre` function that is exported alongside `useMapLibre`

packages/blocks/src/lib/navigation-map/components/obstacle.svelte
```diff
@@ -0,0 +1,59 @@
...
+import { T } from '@threlte/core';
+import type { Obstacle } from '$lib';
+import { view, hovered } from '../stores';
+
+export let obstacle: Obstacle;
+
+let material: THREE.MeshPhongMaterial;
+
+$: name = obstacle.name;
+$: material?.color.set($hovered === obstacle.name ? '#FFD400' : '#FF7D80');
```
mcous, line 14: Can these colors get human readable names as variables?

packages/blocks/src/lib/navigation-map/components/vector-input.svelte
```diff
@@ -0,0 +1,54 @@
...
+import { createEventDispatcher } from 'svelte';
+
+export let readonly = false;
+export let placeholders = ['0', '0', '0'];
+export let labels = ['x', 'y', 'z'];
+export let type: 'integer' | 'number' = 'number';
+export let values: number[] = [];
+export let step = 1;
+
+const inputs: HTMLInputElement[] = [];
```
mcous, line 13: Should this be an object keyed by `label` rather than an array? If the `labels` prop changes, this feels like it will break

packages/blocks/src/lib/navigation-map/components/vector-input.svelte
```diff
@@ -0,0 +1,54 @@
+<script lang='ts'>
+
+import { Label, SliderInput } from '@viamrobotics/prime-core';
+import { createEventDispatcher } from 'svelte';
+
+export let readonly = false;
+export let placeholders = ['0', '0', '0'];
```
mcous, line 7: Also feels like it should be an object keyed by label

packages/blocks/src/lib/navigation-map/components/vector-input.svelte
```diff
@@ -0,0 +1,54 @@
...
+export let labels = ['x', 'y', 'z'];
+export let type: 'integer' | 'number' = 'number';
+export let values: number[] = [];
```
mcous, line 10: Object keyed by label?

packages/blocks/src/lib/navigation-map/stores.ts
```diff
@@ -0,0 +1,27 @@
...
+export const tab = writable<'Obstacles' | 'Waypoints'>('Waypoints');
+
+/** The visible set of tabs. */
+export const tabs = writable<string[]>(['Obstacles', 'Waypoints']);
+
+/** If we're looking at obstacles in a 2d top-down or 3d orbiting view */
+export const view = currentWritable<'2D' | '3D'>('2D');
+
+/** Whether or not we can create obstacles */
+export const write = currentWritable(false);
```
mcous, line 21: Can this be a more descriptive variable than `write`? e.g. `canCreateObstacles`, from the comment

mcous (commented):
Assorted thoughts and questions. Looks like we also might have a bug where vector values of `0` will be missed due to a `if` that looks like it's intending to only guard against `number | undefined`

packages/blocks/src/lib/navigation-map/types.ts
```diff
@@ -0,0 +1,37 @@
+import type { ViamObject3D } from '@viamrobotics/three';
+
+type BaseGeometry = {
+  pose: ViamObject3D;
+};
```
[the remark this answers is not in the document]
micheal-parks: It is - what specific rule is that? I can investigate

packages/blocks/src/lib/navigation-map/stores.ts
```diff
@@ -0,0 +1,27 @@
...
+import { currentWritable } from '@threlte/core';
+import type { Waypoint, Obstacle } from '$lib';
+
+export const hovered = currentWritable<string | null>(null);
+
+export const waypoints = currentWritable<Waypoint[]>([]);
+export const obstacles = currentWritable<Obstacle[]>([]);
+
+/** The currently selected tab. */
+export const tab = writable<'Obstacles' | 'Waypoints'>('Waypoints');
```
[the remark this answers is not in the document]
micheal-parks: I reeaaallly dislike enums in TS 😭 and very much prefer string unions. I don't see an advantage to enums, and do see a lot of unnecessary added verbosity and noise. With string unions type checking is just as strict, and you have an equal amount of issues with having to find/replace if you decide your naming was poor. But I might have not yet heard the right argument for enums?

packages/blocks/src/lib/maplibre/index.svelte
micheal-parks: Map isn't a prop though?

packages/blocks/src/lib/maplibre/hooks.ts
micheal-parks: Ah great suggestion, will do

packages/blocks/src/lib/maplibre/hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+  MapLayerEventType,
+  MapLayerMouseEvent,
+  MapLayerTouchEvent,
+} from 'maplibre-gl';
+import { type Writable, get } from 'svelte/store';
+
+/**
+ * Provides context for a <MapLibre> instance. Must be called within a child of this component.
+ */
+export const useMapLibre = () => {
```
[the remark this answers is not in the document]
micheal-parks: could you link me an example?

packages/blocks/src/lib/maplibre/hooks.ts
micheal-parks: yep, good idea 👍

packages/blocks/src/lib/maplibre/marker.svelte
micheal-parks: Didn't make them reactive - will do 👍

packages/blocks/src/routes/+page.svelte
micheal-parks: This is just playground :P this request will never see a prod env

packages/blocks/src/routes/+page.svelte
micheal-parks: just playground stuff 🛝🛝🛝

packages/blocks/src/routes/+page.svelte
micheal-parks: hmmmm maybe check is not working in CI for the blocks package

packages/blocks/src/lib/navigation-map/stores.ts
micheal-parks: yes, was not a good name - replaced!

packages/blocks/src/lib/maplibre/hooks.ts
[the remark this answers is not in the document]
mcous: https://github.com/viamrobotics/app/blob/4a5891a08540d9e7de4efbd77c2b4c67878ae114/ui/src/lib/auth.ts

packages/core/package.json
micheal-parks: I think something was breaking in CI but agreed, replaced with the standard `{ exports: { import: "..." } }` field

packages/blocks/src/lib/maplibre/hooks.ts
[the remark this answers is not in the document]
micheal-parks: Updated! Thanks for the suggestion, much cleaner

packages/blocks/src/lib/maplibre/index.svelte
micheal-parks: Updated!

packages/blocks/src/lib/navigation-map/components/obstacle.svelte
micheal-parks: Yep, updated!

packages/blocks/src/routes/+page.svelte
mcous: Yeah I agree, I'm getting some weirdness locally, too

packages/blocks/.eslintrc.cjs
mcous: Oh man this rule is straight up crashing (as opposed to flagging an error) for me locally. Is that what you're seeing, too?

micheal-parks: uh oh - nope this was working for me - but I did encounter eslint rules (TS plugin related) crashing last week in this repo specifically

packages/blocks/src/lib/navigation-map/components/vector-input.svelte
micheal-parks: Good call - fixed!

packages/blocks/src/lib/navigation-map/components/vector-input.svelte
micheal-parks: agreed, fixed!

packages/blocks/src/lib/navigation-map/components/vector-input.svelte
micheal-parks: Fixed!

packages/blocks/src/lib/navigation-map/components/center-inputs.svelte
micheal-parks: Agreed, added to theme

packages/core/src/lib/input/slider-input.svelte
micheal-parks: The problem we're trying to solve here is this:

We want the visualization to update instantaneously when the user drags the slider because this tool exists for fine tuning / instantaneous feedback, but we don't want the visualization to update if the user is entering data within the input until an enter keypress or blur occurs. It's basically how [tweakpane handles events](https://cocopon.github.io/tweakpane/) for numeric inputs. I think this may be the best way to go for real-time feedback, and there are some alternatives that I've tried that seem to have more problems, like updating on every input event but having exceptions for trailing decimals or for zeros after decimals so that they don't get sheared off in controlled inputs. Curious if you have thoughts or ideas!

I also want to probably have a guild meeting in the near future to voice some of these problems and begin to formalize solutions

packages/blocks/src/lib/navigation-map/stores.ts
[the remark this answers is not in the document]
micheal-parks: That's fair, in that case, does `Record<string, string> as const` fit needs? I find that they're more flexible than enums:

```ts
export const TabType = {
  Obstacles: 'obstacles',
  Waypoints: 'waypoints',
} as const

// This way we get a string union automatically from the object 
export type TabTypes = keyof typeof TabType
```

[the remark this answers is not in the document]
mcous: > ```ts
> export const TabType = {
>   Obstacles: 'Obstacles',
>   Waypoints: 'Waypoints',
> } as const
> ```

I'm down to use const objects instead of enums, I've just been finding enums a little more convenient, since we don't need to both defined the object and and the type. The `TabType` (object) vs `TabTypes` (union type) feels a little duplicative and confusing

> I find that they're more flexible than enums:

Since I've only just started using enums, I'd be curious to hear about what sort of inflexibility you've encountered

> ```ts
> // This way we get a string union automatically from the object 
> export type TabTypes = keyof typeof TabType
> ```

This forces us to write an object that has matching keys and values, which could lead to annoyances. I think this would need to be a union on the values, instead:

```ts
export type TabTypes = typeof TabType[keyof typeof TabType]
```

packages/blocks/.eslintrc.cjs
mcous: I've disabled this rule in an overrides, so you might be able to remove this line once `main` is merged in!

packages/core/src/lib/input/slider-input.svelte
mcous: That sounds like the difference between `input` (value changes instantaneously) and `change` (user finished entering a value). My gut instinct would be to treat this more like a vanilla `input[type="range"]`

Playing around with Tweakpane, this seems to match their behavior, in that the position of the slider - the instantaneous value - is not moved until the text box value is confirmed via Enter or blur.

- Fire `input` whenever the slider is moved
- Fire `change` whenever the slider is released
- **Do not fire `input`** on textbox `input`, keep it internal
- Fire `input` + `change` on textbox `change`

packages/blocks/src/lib/navigation-map/components/input/geometry.svelte
```diff
@@ -0,0 +1,96 @@
...
+
+const handleDimensionsInput = (event: CustomEvent<Record<string, number>>) => {
+  switch (geometry.type) {
+    case 'box': {
+      const length = event.detail['Length (m)'];
+      const width = event.detail['Width (m)'];
+      const height = event.detail['Height (m)'];
+
+      if (length !== undefined) {
+        geometry.length = length;
```
mcous, line 26: Re-commenting due to diff so this problem doesn't get lost: this function modifies the `geometry` object passed in as a prop, which _will_ change its value in the parent component, which is not good

We should probably start marking prop objects as `Readonly`

micheal-parks: In this case I'm intentionally mutation because this block will run alongside the frameloop whenever users are resizing geometry - I think I might have a solution that I like much better than this though that I think you'll be more comfortable with, stay tuned!

packages/blocks/src/lib/maplibre/index.svelte
micheal-parks: Pretty sure the immutable flag is related to props only, check this out:

https://svelte.dev/repl/6b600f4495b14e2f92555cd6b062de5c?version=4.2.0

Toggling the flag in the "Greeting" component will determine whether assignments to nested properties will reflect in reactive updates in children. If you look at the compiler output it's just swapping a `safe_not_equal` | `not_equal` function being piped into that component

packages/blocks/src/lib/navigation-map/stores.ts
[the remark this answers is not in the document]
micheal-parks: Question: how would you solve this problem using enums:

```
let store = writable<'Waypoints' | 'Obstacles'>('Waypoints')
```

Would it just be?

```
const enum Tabs {
  Waypoints = 'Waypoints',
  Obstacles = 'Obstacles',
}

export const tab = writable<Tabs.Waypoints | Tabs.Obstacles>(Tabs.Waypoints);
```

Or is there a better way? Don't love the above but happy to update this to be consistent if the rest of our code is moving in this direction

packages/core/src/lib/input/slider-input.svelte
micheal-parks: ^ I like this approach a lot. I'm going to turn this description into a ticket, since I think this is something we'll need to ensure that we're doing across the board

.github/workflows/ci.yml
```diff
@@ -15,6 +15,8 @@ jobs:
       - uses: actions/checkout@v3
       - name: Install 🔧
         uses: ./.github/actions/setup
+      - name: Build dependencies 🏗️
+        run: pnpm --dir packages/core build
```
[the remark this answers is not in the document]
ethanlook: Why is `build` required for lint?

packages/blocks/.eslintrc.cjs
micheal-parks: worked, restored this!

packages/blocks/src/lib/maplibre/index.svelte
```diff
@@ -0,0 +1,132 @@
...
+  map!.off('style.load', handleCreate);
+})
+
+$: map?.setMinPitch(minPitch);
+$: map?.setMaxPitch(maxPitch);
+
+</script>
+
+{#if created}
+  <slot />
```
ethanlook, line 120: Did we land on naming all slots if there are multiple slots?

packages/blocks/src/lib/maplibre/marker.svelte
```diff
@@ -0,0 +1,39 @@
...
+<script lang='ts'>
+
+import { onDestroy } from 'svelte';
+import { Marker } from 'maplibre-gl';
+import { useMapLibre } from './hooks';
+import type { LngLat } from '$lib';
+
+export let lngLat: LngLat;
+export let scale = 1;
+export let color: string = '';
```
ethanlook, line 23: nit: docstrings please!

packages/blocks/src/lib/navigation-map/components/input/geometry.svelte
```diff
@@ -0,0 +1,96 @@
...
+
+import { Radio } from '@viamrobotics/prime-core'
+import { createEventDispatcher } from 'svelte';
+import type { Geometry, Shapes } from '$lib';
+import VectorInput from '../vector-input.svelte';
+import { createGeometry } from '../../lib/geometry';
+
+export let geometry: Geometry;
+
+const dispatch = createEventDispatcher<{ input: Geometry }>();
```
ethanlook, line 11: nit: docstrings please

packages/blocks/src/lib/navigation-map/components/input/lnglat.svelte
```diff
@@ -0,0 +1,45 @@
...
+import { SliderInput } from '@viamrobotics/prime-core';
+import { createEventDispatcher } from 'svelte';
+import { LngLat } from 'maplibre-gl';
+
+export let label: string | undefined = undefined;
+export let readonly: boolean | undefined = undefined;
+export let lng: number | undefined = undefined;
+export let lat: number | undefined = undefined;
+
+const dispatch = createEventDispatcher<{ input: LngLat }>();
```
ethanlook, line 12: nit: docstrings please

packages/blocks/src/lib/navigation-map/components/input/orientation.svelte
```diff
@@ -0,0 +1,34 @@
+<script lang='ts'>
+
+import { createEventDispatcher } from 'svelte';
+import { SliderInput, Label } from '@viamrobotics/prime-core';
+
+export let th: number;
+
+const dispatch = createEventDispatcher<{ input: number }>();
```
ethanlook, line 8: nit: docstrings please

packages/blocks/src/lib/navigation-map/components/map.svelte
```diff
@@ -0,0 +1,71 @@
...
+import type { Map } from 'maplibre-gl';
+import { MapLibre } from '$lib';
+import { view, cameraMatrix } from '../stores';
+import ObstacleLayer from './obstacle-layer.svelte';
+import RobotMarker from './robot-marker.svelte';
+import CenterInputs from './center-inputs.svelte';
+import Nav from './nav/index.svelte';
+import Waypoints from './waypoints.svelte';
+
+export let baseGeoPose: { lng: number; lat: number } | undefined = undefined;
```
ethanlook, line 13: nit: docstrings please

packages/blocks/src/lib/navigation-map/components/nav/index.svelte
```diff
@@ -0,0 +1,44 @@
+<script lang='ts'>
+
+import { createEventDispatcher } from 'svelte';
+import { Tabs } from '@viamrobotics/prime-core';
+import type { Obstacle } from '$lib';
+import { tab, tabs, hovered } from '../../stores';
+import ObstaclesTab from './obstacles.svelte';
+import WaypointsTab from './waypoints.svelte';
+
+const dispatch = createEventDispatcher<{ 'update-obstacles': Obstacle[] }>();
```
ethanlook, line 10: nit: docstrings please

packages/blocks/src/lib/navigation-map/components/nav/obstacles.svelte
```diff
@@ -0,0 +1,152 @@
...
+import OrientationInput from '../input/orientation.svelte';
+import { environment, hovered, boundingRadius, obstacles } from '../../stores';
+import { calculateBoundingBox } from '../../lib/bounding-box';
+import { createEventDispatcher } from 'svelte';
+import { createObstacle } from '$lib/navigation-map/lib/create-obstacle';
+import { createName } from '$lib/navigation-map/lib/create-name';
+
+type Events = {
+  update: Obstacle[]
+};
```
ethanlook, line 16: nit: docstrings please

packages/blocks/src/lib/navigation-map/components/nav/obstacles.svelte
```diff
@@ -0,0 +1,152 @@
...
+const { map } = useMapLibre();
+
+const handleSelect = (selection: { name: string; location: LngLat }) => {
+  const zoom = boundingRadius[selection.name]!;
+  const bb = calculateBoundingBox(zoom, selection.location);
+  map.fitBounds(bb, { duration: 800, curve: 0.1 });
+};
+
+const handleLngLatInput = (name: string) => (event: CustomEvent<LngLat>) => {
+  const index = $obstacles.findIndex((obstacle) => obstacle.name === name);
```
ethanlook, line 29: Here and in the other handlers: is it safe to assume that the obstacle index will be found? You assert that the obstacles exist immediately after finding index.

packages/blocks/src/lib/navigation-map/components/obstacle.svelte
```diff
@@ -0,0 +1,63 @@
+
+<script lang='ts'>
+
+import { theme } from '@viamrobotics/prime-core/theme';
+import * as THREE from 'three';
+import { T } from '@threlte/core';
+import type { Obstacle } from '$lib';
+import { view, hovered } from '../stores';
+
+export let obstacle: Obstacle;
```
ethanlook, line 10: nit: docstring please

packages/blocks/src/lib/navigation-map/components/obstacle.svelte
```diff
@@ -0,0 +1,63 @@
...
+      <T.CapsuleGeometry
+        computeBounding={name}
+        args={[geometry.radius, geometry.length, 16, 32]}
+      />
+    {/if}
+    <T.MeshPhongMaterial
+      bind:ref={material}
+    />
+  </T.Mesh>
+{/each}
```
ethanlook, line 63: This makes me really want a generic renderer API for geometries like these. We shouldn't have to add a file like this every time we want to add a new visualization type. Fine for now but let's continue the conversation.

packages/blocks/src/lib/navigation-map/components/robot-marker.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang='ts'>
+
+import { theme } from '@viamrobotics/prime-core/theme';
+import { MapLibreMarker, type LngLat } from '$lib';
+
+export let lngLat: LngLat | undefined = undefined;
```
ethanlook, line 6: nit: docstring please

packages/blocks/src/lib/navigation-map/index.svelte
```diff
@@ -0,0 +1,60 @@
...
+/** The obstacles to render on the map. */
+export let obstacles: Obstacle[] = [];
+
+/** The initial tab to show */
+export let tab: 'Waypoints' | 'Obstacles' = 'Waypoints';
+
+/** The visible set of tabs */
+export let tabs = ['Waypoints', 'Obstacles'];
+
+export let baseGeoPose: { lng: number; lat: number } | undefined = undefined;
```
ethanlook, line 44: nit: docstring please

packages/blocks/tailwind.config.ts
```diff
@@ -2,7 +2,7 @@ import type { Config } from 'tailwindcss';
 import { theme } from '@viamrobotics/prime-core/theme';
 
 export default {
-  content: ['./src/**/*.{ts,svelte}'],
+  content: ['./src/**/*.{ts,svelte}', '../core/dist/**/*.{ts,js,svelte}'],
   theme,
```
ethanlook, line 6: Why not add the core `node_modules` here?

packages/core/theme.ts
```diff
@@ -54,6 +58,18 @@ export const theme = {
...
+      'power-wire': '#FF0047',
+      'pcb': '#00501F',
+      'power-button': '#00EF83',
+      'solar-power': '#FFD400',
+      'mars-rover': '#FF7D80',
+      'raspberry': '#9F0049',
+      'hologram': '#00E8E8',
+      'yoshimi': '#FF79FF',
+      'cyberpunk': '#A51AFF',
+      'hyperlink': '#0000EA',
```
ethanlook: Just double checking - are these consistent in figma?

ethanlook: nit: the other colors use lowercase letters in the hex code

ethanlook (commented):
Sorry for being the docstring enforcer! Do you mind doing a pass over all of your changes to be sure components, props, and custom events have docstrings? I tried to catch as much as I could. Thanks!

Could you please bump the `prime-core` and `prime-blocks` versions? Thanks!

Lastly, I'm a little confused by the code organization under `packages/blocks/src/lib/navigation-map`. It is weird to me to have a `lib` nested under a `lib` and such - it feels like its own package. I don't have a good suggestion here but wondering how it makes you feel?

packages/blocks/src/lib/navigation-map/stores.ts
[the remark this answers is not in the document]
mcous: > Question: how would you write this block using enums:
> 
> ```ts
> let store = writable<'Waypoints' | 'Obstacles'>('Waypoints')
> ```

You can use the enum name itself the same way you would use a union type:

```ts
const enum Tab {
  Waypoints = 'Waypoints',
  Obstacles = 'Obstacles',
}

export const tab = writable<Tab>(Tab.Waypoints);
```

> One last thing: if we go with enums I'd prefer it to remain internal, this to me looks like an overcomplicated public api:
> 
> ```ts
> import { NavigationMap, Tabs } from '@viamrobotics/prime-blocks'
> 
> ...
> <NavigationMap tab={Tabs.Obstacles} />
> ```
> 
> I'd rather
> 
> ```ts
> import { NavigationMap, Tabs } from '@viamrobotics/prime-blocks'
> 
> ...
> <NavigationMap tab='Obstacles' />
> ```

Matter of personal preference; I prefer the first one 😝. In order to accept a string-based public API  - which has its benefits! - then we should go with the object-based approach above. That would give us the best of both worlds, though naming things becomes ever-so-slightly annoying

```ts
export const NavigationTab = {
  Waypoints: 'Waypoints',
  Obstacles: 'Obstacles',
} as const

export type NavigationTabType = typeof NavigationTab[keyof typeof NavigationTab]

export const tab = writable<NavigationTabType>(NavigationTab.Waypoints);
```

Consumers could then chose to use:

```ts
import { NavigationMap, NavigationTab } from '@viamrobotics/prime-blocks'

<NavigationMap tab={NavigationTab.Obstacles} />
// or
<NavigationMap tab='Obstacles' />
```

packages/core/theme.ts
micheal-parks: They do not exist in figma ✨ These are our brand colors. They should exist in figma though, I'll ask design to do this.

.github/workflows/ci.yml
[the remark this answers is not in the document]
micheal-parks: It's because blocks uses the workspace version of core during development / linting / typechecking, which requires a build of core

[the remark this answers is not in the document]
ethanlook: Gotcha! Then I agree about combining. Thanks!

packages/blocks/src/lib/maplibre/index.svelte
micheal-parks: Not sure - I think we're all going back and forth on it. I'm going to be conservative and not name this one, especially since it's the default or catch-all slot 🗑️

packages/blocks/src/lib/maplibre/marker.svelte
micheal-parks: woops, thanks for catching!

packages/blocks/src/lib/navigation-map/components/obstacle.svelte
micheal-parks: For sure, curious about your thoughts!

packages/core/theme.ts
micheal-parks: Updated! Wish prettier would do this...

micheal-parks:
@ethanlook No prob about the docstrings, added!

> Lastly, I'm a little confused by the code organization...

It very much also feels like its own package to me too - part of this was due to just how complicated rendering was at a geospatial scale / making the UX for geospatial viz not horrible by including conveniences like zooming obstacle size / etc. Google maps even [has their own package](https://www.npmjs.com/package/@googlemaps/three) just to handle a lot of quirks and boilerplate that arise in this space, unfortunately in the free space (or even in the mapbox space) the options are pretty terrible.

I'm ok with the nested lib folder (unless you can think of a better name), it is what is says it is: a collection of library functions that are only gonna be used here. Actually, there are some things that might be better off in the `@viamrobotics/three` package, but I'd like to leave that for a follow up step

packages/blocks/tailwind.config.ts
micheal-parks: Is there something from `node_modules` we should be watching?

packages/blocks/src/lib/navigation-map/stores.ts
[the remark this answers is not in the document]
micheal-parks: Gotcha! Updated!

packages/blocks/src/lib/navigation-map/components/nav/obstacles.svelte
micheal-parks: Yep, name is a required key for this data structure, so an error being thrown if an index isn't found makes sense to me

.github/workflows/ci.yml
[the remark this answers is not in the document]
micheal-parks: Going to keep this for follow up

PR #327, opened 2023-08-15
Add fonts to core tailwind theme

mcous:
It is still the consumer's responsibility to make sure the fonts are available in its build, this just adds the classes to tailwind

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,10 +1,11 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.8",
+  "version": "0.0.9",
   "publishConfig": {
     "access": "public"
   },
   "scripts": {
+    "prepare": "svelte-kit sync",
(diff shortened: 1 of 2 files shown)


micheal-parks (approved):

packages/core/package.json
```diff
@@ -1,15 +1,15 @@
...
   "publishConfig": {
     "access": "public"
   },
   "scripts": {
+    "prepare": "svelte-kit sync",
     "dev": "vite dev",
     "build": "vite build && pnpm run package",
     "preview": "vite preview",
     "package": "svelte-kit sync && svelte-package && publint",
-    "prepublishOnly": "pnpm run package",
```
mcous, line 12: Note: `prepublishOnly` implicitly runs before `publish`, which means we were implicitly rebuilding the project during publish, even though we were pulling down the built artifacts. Fun stuff

PR #328, opened 2023-08-15
DATA-1828 - Add left right arrows

alexis-wei:
add left and right arrow icons to our library


packages/core/src/lib/icon/icons.ts
```diff
@@ -16,6 +16,8 @@ export const paths: Record<string, string> = {
   'arrow-expand-left': MDI.mdiArrowExpandLeft,
   'arrow-expand-right': MDI.mdiArrowExpandRight,
   'arrow-top-right-bottom-left': MDI.mdiArrowTopRightBottomLeft,
+  'arrow-left': MDI.mdiArrowLeft,
```
micheal-parks, line 19: can you sort these alphabetically?

alexis-wei: done!

micheal-parks (approved):

PR #329, opened 2023-08-15
Improve numeric inputs and icon

micheal-parks:
This PR improves the icon component:
* Removes the title attribute, which is causing internal icon labels to show up on hover
* Makes icons aria-hidden, we should be using aria-labels on buttons and other elements (see link in component comment)

...and improves the input component:
* Exposes the input element as a bindable variable, which helps with `input.focus()`, etc.
* Allows numbers to be passed to inputs
* Switches to using native HTML keyboard events for numeric input incrementing / decrementing
* Fixes a few small dragging issues with the slider input
* Remove decimal stripping from type=integer. This behavior has been error prone from the start, and we should instead just report invalid input states and let the component consumer handle invalid states


packages/core/src/lib/input/numeric-input.spec.ts
```diff
@@ -11,43 +11,19 @@ describe('Numeric Input', () => {
...
-  });
-
-  it('Should decrement by two on arrowdown keydown', async () => {
-    render(Input, { placeholder: 'Enter a number', step: 2 });
-
-    const input: HTMLInputElement =
-      screen.getByPlaceholderText('Enter a number');
-
-    expect(input.value).toBe('');
+  it('Should set the input value', () => {
```
micheal-parks, line 14: We shouldn't test arrow up / down behavior since it is native

packages/core/src/lib/input/numeric-input.svelte
```diff
@@ -7,19 +7,11 @@ For numeric user inputs.
 <NumericInput type="integer"  on:input={onInput} />
 ```
 -->
-<svelte:options
-  immutable
-  accessors
-/>
+<svelte:options immutable />
```
micheal-parks, line 10: Accessors aren't really valuable since these aren't web components

packages/core/src/lib/input/slider-input.svelte
```diff
@@ -84,7 +85,7 @@ const handlePointerMove = (event: PointerEvent) => {
     return;
   }
 
-  if (next > startValue) {
+  if (next >= startValue) {
```
micheal-parks, line 88: This fixes the slider not animating over the initial drag position

packages/core/src/lib/input/slider-input.svelte
```diff
@@ -97,8 +98,10 @@ const handlePointerMove = (event: PointerEvent) => {
     `;
   }
 
-  value = next;
-  dispatch('input', value);
+  if (value !== next) {
+    value = next;
+    dispatch('input', value);
```
micheal-parks, line 103: This will reduce thrashing the consumer of this event

packages/core/.eslintrc.cjs
```diff
@@ -13,5 +13,8 @@ module.exports = {
   },
   rules: {
     'sonarjs/no-duplicate-string': 'off',
+    // @todo(mp): move to js-config
```
ethanlook, line 16: Is there a ticket to track this?

packages/core/src/lib/input/numeric-input.svelte
```diff
@@ -31,61 +23,25 @@ export let type: NumericInputTypes = 'number';
...
 
 /** The amount to increment/decrement when using the up/down arrows. */
 export let step = 1;
 
-const dispatch = createEventDispatcher<{
-  input: number;
-  keydown: number;
-}>();
+/** The HTML input element. */
+export let input: HTMLInputElement = undefined!
```
ethanlook, line 32: `undefined!` again

ethanlook (approved):

packages/core/src/lib/switch.spec.ts
```diff
@@ -10,28 +10,17 @@ describe('Switch', () => {
...
-
-    await fireEvent.mouseEnter(tooltipIcon);
-
-    await waitFor(() => {
-      expect(screen.getByText('Tooltip Message')).toBeVisible();
-    });
-
-    const icon = screen.getByRole('img', { name: /information-outline/u });
-    expect(icon).toBeVisible();
+    const icon = container.querySelector('svg') as Element;
```
anjinai, line 22: [q] - is the reason we are no longer testing the tooltip's behavior on hover in the switch tests because that should already be tested in the tooltip specific tests?

packages/core/src/lib/input/input.svelte
```diff
@@ -35,31 +35,20 @@ export let disabled = false;
 /** The state of the input (info, warn, error, success), if any. */
 export let state: InputState = 'none';
 
+/** The HTML input element. */
+export let input: HTMLInputElement = undefined!;
```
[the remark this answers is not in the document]
micheal-parks: 😂 It's not as scary as it looks. Here we're just asserting that `input` will be defined by the time that it's used by the parent component. It's really just the explicit type definition of how binding is typed within components, and only necessary because it's a prop:

```svelte
<script lang='ts'>

let input: HTMLElement

onMount(() => {
  input.value = 3
})

</script>

<input bind:this={input} />
```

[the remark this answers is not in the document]
ethanlook: Gotcha! Could you leave a comment explaining this? To the untrained eye it's definitely scary!

packages/core/src/lib/switch.spec.ts
micheal-parks: yeah, two reasons:
* IMO this test was overstepping its boundary
* Mocking hover behavior was being flakey once I changed the selector, and I don't think unit tests should really attempt to simulate hover states, I think if we want to test this we should instead test that props are passed correctly to the tooltip

packages/core/src/lib/input/input.svelte
micheal-parks, line 39:
```suggestion 
export let input: HTMLInputElement = undefined!;
// Assert this element will be defined by the time it is used by the parent.
```

mcous:
@micheal-parks @ethanlook sorry I'm late to this review! I want to follow up on this:

```ts
/** The HTML input element. */
export let input: HTMLInputElement = undefined!;
```

> 😂 It's not as scary as it looks. Here we're just asserting that input will be defined by the time that it's used by the parent component.

I don't quite agree with this statement. The `input` variable will be defined **if an only if it's used after the component is mounted.** Otherwise, `input` will be `undefined`, and the `!` hides this.

[Playground link](https://svelte.dev/repl/c8b5bc060ed64d5f92c54db5bec0e8c9?version=4.2.0)

```svelte
<script>
	import {onMount} from 'svelte'
	
	let element;

	console.log("Element in script:", typeof element)

	onMount(() => {
		console.log("Element in mount:", typeof element)
	})
</script>

<p bind:this={element}>Hello world</p>
```

Logs:

```
"Element in script:" "undefined"
"Element in mount:" "object"
```

PR #330, opened 2023-08-16
Add SLAM 2d map to prime-blocks

micheal-parks:
This PR transplants the 2d Slam map (currently in the RC) to prime-blocks with small tweaks that makes the API less opinionated:

```svelte
<SlamMap2d
  pointcloud={new Uint8Array()} // An arraybuffer representing a PCD file.
  basePose={{ x: number, y: number, z: number }} // An optional pose of a robot.
  destination={{ x: number, y: number }} // An optional user-specificed robot destination.
  helpers={true | false} // Whether or not scene helpers should be rendered. Default true.
  motionPath={'0.1,0.2\n0.3,0.4\n'} // An optional motion path.
/>
``` 

slammy

The map was mostly moved as-is to help unblock Steven who will need it in app soon.

It also bundles in the following fixes / improvements for expediency:
* Improve types for bound input HTML elements
* Remove flex class from icon elements


packages/blocks/.eslintrc.cjs
```diff
@@ -12,6 +12,15 @@ module.exports = {
...
+    // @todo(mp APP-2482): move to js-config
     'sonarjs/no-duplicate-string': 'off',
+    'no-undef-init': 'off',
+    'unicorn/no-useless-undefined': 'off',
+
+    // @todo(mp APP-2482): diagnose why these aren't working correctly
+    '@typescript-eslint/no-unsafe-assignment': 'off',
+    '@typescript-eslint/no-unsafe-return': 'off',
+    '@typescript-eslint/no-unsafe-member-access': 'off',
+    '@typescript-eslint/no-unsafe-call': 'off',
```
micheal-parks, line 24: Our TS eslint plugin is not able to type check svelte components. Added a ticket to investigate [here](https://viam.atlassian.net/browse/APP-2482) and turned off for now since svelte-check will still catch these since we're in TS strict mode.

packages/blocks/src/lib/assets/images/base-marker.txt
mcous, line 1: Why is this file needed? I thought Vite is capable of importing `png` as a data URL

packages/blocks/src/lib/assets/images/converter.js
mcous, line 1: Same comment about committing the image data url; it seems a little prone to desync. I think it'd be better to import `png` directly, or - if we're not comfortable with the whims of `assetInlineLimit` - add this logic as a [Vite import plugin](https://stackoverflow.com/questions/73847316/import-raw-image-data-into-a-script-with-vite), instead

mcous (commented):
Quick thought on data URLs

packages/blocks/.eslintrc.cjs
mcous: > out TS eslint plugin is not able to type check svelte components

Weird that it only seems to be happening with the `on:create` arguments of various `T.[]` components. If this was a general problem I'd expect to see a bunch more errors, both here and in app, which uses the same ESLint config.

Perhaps it's safer to use ignore comments rather than disable the rules wholesale, since it seems to be limited to `@threlte/core`?

packages/blocks/src/lib/slam-map-2d/points.svelte
```diff
@@ -0,0 +1,154 @@
...
+
+extend({ MeshDiscardMaterial });
+
+/** A buffer representing a .pcd file */
+export let pointcloud: Uint8Array | undefined;
+
+/** The size of each individual point */
+export let size: number;
+
+type $$Events = {
```
ethanlook, line 25: Why `$$Events` here and `Events` in `packages/blocks/src/lib/slam-map-2d/index.svelte`?

Does declaring the type separately show up in IDE suggestions as expected?

packages/blocks/tsconfig.json
```diff
@@ -5,6 +5,7 @@
   ],
   "compilerOptions": {
     "types": ["svelte", "vite/client", "vitest"],
+    "noPropertyAccessFromIndexSignature": false,
```
ethanlook, line 8: Do we have a ticket to track moving this into `js-config`? Do we want that in `js-config`?

ethanlook (commented):
There are several unrelated things in this PR - would it be possible to split this into smaller PRs? In general, I strongly prefer small PRs and don't love the pattern of slipping things into unrelated PRs. Happy to chat more offline.

micheal-parks:
> would it be possible to split this into smaller PRs?

@ethanlook I'd normally agree - the only reason I'm bundling like this is because of deliverable pressure coming from the team awareness side. Once obstacle avoidance is wrapped up I'm happy to take the time to split things - and this is my normal modus operandi - up but I'm honestly too exhausted right now to take the time 😞

ethanlook:
> > would it be possible to split this into smaller PRs?
> 
> @ethanlook I'd normally agree...

I hear you and I totally get it! I appreciate the acknowledgment! Carry on 🚀

ethanlook (approved):

packages/blocks/tsconfig.json
micheal-parks: We do have one! Ticket here: https://viam.atlassian.net/browse/APP-2482.

I think the best course of action is to disable until we can fix at the source

packages/blocks/src/lib/slam-map-2d/points.svelte
micheal-parks: `$$Events`` has a different effect: it singles complete changes to event dispatcher payload rather than the contents of a custom event's detail object. It's used within threlte applications because of the usage of `createRawEventDispatcher`, which bypasses using CustomEvent objects since we're working in webgl rather than HTML-land. All intellisense is preserved 👍

packages/blocks/src/lib/assets/images/base-marker.txt
micheal-parks: Yeah I'm also very not happy with this setup, I'm only preserving it at the moment because direct png imports failed with Vite when we initially set this up in RC + imported into app. I'd rather preserve the thing that we know works while the deadlines are tight and do the good thing afterwards

packages/blocks/.eslintrc.cjs
micheal-parks: Found a workaround! I think the issue is related to how there are currently type-checking issues in Svelte's template sections. I was able to restore all rules by just passing references to functions that were typed in the script section, and that cleared everything up.

packages/blocks/src/lib/slam-map-2d/motion-path.svelte
```diff
@@ -0,0 +1,57 @@
...
+      const y = Number.parseFloat(yString) / 1000;
+      points.push(x, y, 0);
+    }
+  }
+
+  const vertices = new Float32Array(points);
+  geometry.setPositions(vertices);
+};
+
+$: updatePath(path);
```
mcous, line 45: Given that we create a new `LineGeometry` every time, would this make the intent more clear as a reactive derived variable?

```svelte
const pathStringToGeometry = (pathstr?: string): LineGeometry => {
  if (pathstr === undefined) {
    return new LineGeometry();
  }

  const points: number[]

  // ...

  const vertices = new Float32Array(points);
  const geometry = new LintGeometry()

  geometry.setPositions(vertices);

  return geometry
}

$: geometry = pathStringToGeometry(path)
```

packages/blocks/src/lib/slam-map-2d/motion-path.svelte
```diff
@@ -0,0 +1,57 @@
...
+import { LineGeometry } from 'three/examples/jsm/lines/LineGeometry.js';
+import { renderOrder } from './constants';
+
+extend({ Line2, LineMaterial });
+
+export let path: string | undefined;
+
+let geometry: LineGeometry = new LineGeometry();
+
+const updatePath = (pathstr?: string) => {
```
mcous, line 23: Should this function be extracted and unit tested?

packages/blocks/src/lib/slam-map-2d/motion-path.svelte
```diff
@@ -0,0 +1,57 @@
...
+
+  const vertices = new Float32Array(points);
+  geometry.setPositions(vertices);
+};
+
+$: updatePath(path);
+
+</script>
+
+{#if path}
```
mcous, line 49: Should this check `geometry` rather than `path`, given `geometry` is the variable that's actually used?

packages/blocks/src/lib/slam-map-2d/points.svelte
```diff
@@ -0,0 +1,158 @@
...
+  material.size = size;
+}
+$: if (pointcloud) {
+  update(pointcloud);
+}
+$: raycaster.camera = camera.current as THREE.OrthographicCamera;
+
+onMount(() => dispatch('update', { center, radius }));
+
+raycaster.addEventListener('click', (event: THREE.Event) => {
```
mcous, line 135: Should this event listener be removed in `onDestroy`?

packages/blocks/src/lib/slam-map-2d/points.svelte
mcous, line 1: This component looks like a good candidate to split up into a relatively bare collaborator whose dependencies can be unit tested, given the amount of processing done in the `<script>` tag and the sparse template

packages/legacy/package.json
mcous, line 1: Are the changes in `packages/legacy` necessary for this PR, or did they get added in the course of investigating the ESLint thing? If they are unrelated and can be dropped from this PR, that would be better given the number of `test.skips` that seem to have happened

mcous (approved):
LGTM! Some minor organizational comments (plus a `NaN` recommendation) and requests for whenever is convenient in terms of feature urgency

packages/blocks/src/lib/slam-map-2d/constants.ts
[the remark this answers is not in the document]
micheal-parks: Yeah totally with you concerning constants, helpers, utils. Renamed!

packages/blocks/src/lib/slam-map-2d/helpers.svelte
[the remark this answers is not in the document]
micheal-parks: In this - and only this - case I'd argue helpers is meaningful because of three's naming conventions here: https://threejs.org/docs/index.html#api/en/helpers/ArrowHelper

packages/blocks/src/lib/slam-map-2d/motion-path.svelte
```diff
@@ -0,0 +1,57 @@
...
+  geometry = new LineGeometry();
+
+  const points: number[] = [];
+
+  for (const xy of pathstr.split('\n')) {
+    const [xString, yString] = xy.split(',');
+    if (xString !== undefined && yString !== undefined) {
+      const x = Number.parseFloat(xString) / 1000;
+      const y = Number.parseFloat(yString) / 1000;
+      points.push(x, y, 0);
```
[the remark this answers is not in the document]
micheal-parks: Done!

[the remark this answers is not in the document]
micheal-parks: Done!

packages/blocks/src/lib/slam-map-2d/motion-path.svelte
micheal-parks: Yes! Just not yet - this is an advanced debugging feature that really only has internal use. There's a project in the works to hook the feature up to an API, at which point the incoming path data and therefore the function body is very likely to change. We should do it then if necessary (might just end up being plugging the response into the renderer rather than any parsing).

packages/blocks/src/lib/slam-map-2d/points.svelte
micheal-parks: Good catch! Done.

packages/blocks/src/lib/slam-map-2d/points.svelte
micheal-parks: Good point, moved a majority of the logic into a separate .ts file

packages/blocks/package.json
```diff
@@ -30,42 +30,54 @@
     "prime.css"
   ],
   "peerDependencies": {
```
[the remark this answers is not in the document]
micheal-parks: I had the same thought - everything here is required for the navigation and slam blocks. Maybe something to re-assess as more blocks are added?

packages/blocks/src/lib/slam-map-2d/motion-path.svelte
micheal-parks: I like this, updated!

packages/legacy/package.json
micheal-parks: Yeah, unfortunately I wasn't able to get CI to work until I upgraded every package. I'm not sure why and think my lack of experience with pnpm monorepos is probably to blame 😅 

All but two of the skipped tests are in a component we've completely stopped using due to event issues: VectorInput. The remaining two are testing v-code-editor, which we never ended up adopting in app.

packages/blocks/src/lib/slam-map-2d/helpers.svelte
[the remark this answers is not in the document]
mcous: Damn ok I'll raise my concerns with Three.js 🙃

PR #332, opened 2023-08-18
Add select, migrate to lint-staged

DTCurrie:
Add's an implementation of the native select with our styles, searchable selects and multi-selects are up next.

Screenshot 2023-08-18 at 12 10 48 PM


.lintstagedrc.yaml
```diff
@@ -0,0 +1,2 @@
+"*.{js,ts,svelte,md,html,css}": "eslint"
+"*.{js,ts,svelte,md,mdx,html,css}": "prettier --check"
```
DTCurrie, line 2: Let me know if I missed anything here.

DTCurrie, line 2: Let me know if I missed anything here.

packages/core/src/lib/input/input.svelte
```diff
@@ -80,7 +79,7 @@ $: icon = {
 
   {#if icon !== ''}
     <span
-      class={cx('absolute right-2 top-1.5', {
+      class={cx('absolute right-2', {
```
DTCurrie, line 82: Fixes the icon issue seen here: 
image

DTCurrie:
Adding docs comments now.

packages/core/src/lib/select/select.svelte
```diff
@@ -0,0 +1,65 @@
...
+        'border-info-dark focus:outline-[1.5px] focus:-outline-offset-1 focus:outline-info-dark':
+          isInfo,
+        'border-warning-bright focus:outline-[1.5px] focus:-outline-offset-1 focus:outline-warning-bright':
+          isWarn,
+        'border-danger-dark focus:outline-[1.5px] focus:-outline-offset-1 focus:outline-danger-dark':
+          isError,
+      }
+    )}
+    {...$$restProps}
+    on:select
```
micheal-parks, line 58: Should we go with `on:input` like the other inputs? Also I don't think `<select>` fires this event:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#see_also

packages/core/src/lib/select/select.spec.svelte
```diff
@@ -0,0 +1,30 @@
...
+  - https://sveltesociety.dev/recipes/testing-and-debugging/unit-testing-svelte-component
+  - https://github.com/testing-library/svelte-testing-library/issues/48
+ -->
+<script lang="ts">
+import type { InputState } from '$lib/input/input.svelte';
+import Select from './select.svelte';
+
+export let value: string | undefined = undefined;
+export let disabled = false;
+export let state: InputState = 'none';
```
micheal-parks, line 18:
```suggestion
export let state: InputState | undefined = 'none';
```
Otherwise TS/ESLint will complain if an explicit property isn't passed, I fixed the other inputs in https://github.com/viamrobotics/prime/pull/326

micheal-parks (approved):

packages/core/src/lib/select/select.spec.svelte
DTCurrie: This is just for a test element but sure.

DTCurrie: Tried it on both and am getting this linting error:

`'any' overrides all other types in this union type  @typescript-eslint/no-redundant-type-constituents`

micheal-parks: that's... odd, there aren't any `any`s in that union?

packages/core/src/lib/input/input.svelte
ethanlook: Does this have any impact on other icon use?

ethanlook (approved):

packages/core/src/lib/input/input.svelte
DTCurrie: @ethanlook These are the only three icons allowed on the input, they are based on the `state` prop.

PR #334, opened 2023-08-22
Move tests to __tests__, button refactoring/fixes, additional cleanup

DTCurrie:


packages/core/src/lib/button/button.svelte
```diff
@@ -51,26 +48,26 @@ export let width: 'full' | 'default' = 'default';
   class={cx(
     'mx-auto inline-flex items-center justify-center gap-1.5 whitespace-nowrap border py-1.5 text-xs',
     {
-      'pointer-events-none select-none !border-disabled-light !bg-disabled-light text-disabled-dark':
+      '!border-disabled-light !bg-disabled-light text-disabled-dark pointer-events-none select-none':
```
mcous, line 51: While we're cleaning up button: in the last component library I worked in, `pointer-events: none` on disabled buttons messed with tooltips, because `mouseenter` et. al. are pointer events. This was a problem because we wanted the tooltip to explain _why_ the button was disabled.

In general, I think `pointer-events: none` should be used really sparingly. In this specific case, since this is a native `<button>` and we are setting the disabled attribute, I don't think we need to disable pointer events to prevent the click handler from firing; it's built-in browser behavior.

(This comment probably applies to `select-none`, too, but I haven't used it before)

mcous (approved):
LGTM

packages/core/src/lib/button/button.svelte
DTCurrie: I agree, and actually this current version makes our buttons pretty inaccessible. We probably shouldn't be using `disabled` since that will make buttons imperceivable to screen readers. `select-none` prevents users from selecting the button text, IMO we may just want that always on unless we feel it is useful to let users select/copy button text.

packages/core/src/lib/class-names.ts
DTCurrie, line 1: @mcous not sure I love this but wanted to document usage for these and not rely on the ambiguous `classnames` import, but very open to suggestions here.

packages/core/src/lib/button/button.svelte
```diff
@@ -41,37 +42,50 @@ export let icon = '';
 
 /** The width of the button. */
 export let width: 'full' | 'default' = 'default';
+
+/** Additional CSS classes to pass to the button. Can be a `classnames.Arguement`. */
+export let classes: ClassNames = '';
+$: classList = classNames(
```
DTCurrie, line 48: @mcous and here is how we are using the `class-names` helper. There may be a cleaner way to do this but not sure.

packages/core/src/__tests__/button.spec.ts
```diff
@@ -62,9 +62,12 @@ describe('Button', () => {
...
 
-  it('Renders a disabled button', () => {
-    render(Button, { disabled: true });
-    expect(screen.getByRole('button')).toBeDisabled();
+  it('Renders a disabled button', async () => {
+    const { component } = render(Button, { disabled: true });
+    const onClick = vi.fn();
+    component.$on('click', onClick);
+    await fireEvent.click(screen.getByRole('button'));
+    expect(onClick).not.toHaveBeenCalledOnce();
```
mcous, line 70: [nit] if `onClick` was called twice, this assertion would pass

```suggestion
    expect(onClick).not.toHaveBeenCalled();
```

packages/core/src/__tests__/icon-button.spec.ts
```diff
@@ -1,47 +1,51 @@
...
 
-  it('Renders a disabled button', () => {
-    render(ButtonIcon, { disabled: true });
-    expect(screen.getByRole('button')).toBeDisabled();
+  it('Renders a disabled button', async () => {
+    const { component } = render(IconButton, { ...common, disabled: true });
+    const onClick = vi.fn();
+    component.$on('click', onClick);
+    await fireEvent.click(screen.getByRole('button'));
+    expect(onClick).not.toHaveBeenCalledOnce();
```
mcous, line 49:
```suggestion
    expect(onClick).not.toHaveBeenCalled();
```

packages/core/src/lib/button/icon-button.svelte
```diff
@@ -0,0 +1,73 @@
...
+export let variant: 'primary' | 'danger' = 'primary';
+
+/**
+ * The text that appears in a native popup box on hoveClassListArguements to the value
+ * of `label`.
+ */
+export let title = label;
+
+/** Additional CSS classes to pass to the button. */
+export let classes: cx.Argument = '';
```
mcous, line 39: If we wanted to use `class` as the prop name, I think we can, according to the [Svelte docs on exporting props](https://svelte.dev/docs/svelte-components#script-1-export-creates-a-component-prop)!

To nitpick: `classes` feels a little non-standard and I'd rather use something conventional, like `class`. I even wonder if `cx` would make sense, to make it clear that it's gonna get passed to `classnames`

```ts
$: classList = cx(..., extraClasses)

...

export { extraClasses as cx }
```

packages/core/src/lib/button/button.svelte
```diff
@@ -85,6 +98,6 @@ export let width: 'full' | 'default' = 'default';
   {/if}
 
   <span>
-    {label}
+    <slot />
```
mcous, line 101: 🎉

packages/core/vite.config.ts
```diff
@@ -9,7 +9,7 @@ export default defineConfig({
     },
   },
   test: {
-    include: ['src/**/*.spec.ts'],
+    include: ['src/__tests__/*.spec.ts'],
```
mcous, line 12: This misses nested tests, which is probably going to confuse someone in the future. I think this change should be reverted or modified to add an extra `**`:

```suggestion
    include: ['src/**/__tests__/*.spec.ts'],
```

packages/core/src/__tests__/badge.spec.ts
mcous, line 1: The way I'm used to working with a `__tests__` directory - and the way I intended in the proposal is Slack - is to place multiple test directories as siblings to the subjects under test. This PR, however, creates a single, flat `src/__tests__` directory, which breaks [test file symmetry](https://github.com/testdouble/contributing-tests/wiki/Symmetrical-Unit-Test).

Can we move these spec directories and files to be symmetrical? e.g.

```
- src/
    - lib/
        - badge.svelte
        - breadcrumbs.svelte
        - __tests__/
            - badge.spec.ts
            - breadcrumbs.spec.ts
        - button/
            - button.svelte
            - __tests__/
                - button.spec.svelte
```

mcous (changes requested):
Main change request is a tweak to the `__tests__` directory structure. Would also prefer something more "standard" than `classes` for the class name prop

packages/core/src/__tests__/button.spec.ts
DTCurrie: Whoops that is an autocomplete miss, I meant to just use `toHaveBeenCalled` good catch.

packages/core/src/lib/button/icon-button.svelte
DTCurrie: Oh yeah I think we actually do that elsewhere so that would be great. I am not sure about `cx`, it kind of comes from the react world of `classSet` right? Maybe we just use `class` and at least the typing should show it can of `classnames` type? We can also call it out in documentation, but also just plain class strings work.

packages/core/src/__tests__/badge.spec.ts
DTCurrie: Sure.

mcous (approved):
:tada:

PR #337, opened 2023-08-23
Update table to use cx instead of style

DTCurrie:
Updated the `table` elements to accept `cx` as a class to support extending styles using tailwind classes. Also just some general cleanup around stories for passing `cx`. This will keep extending styles consistent with how we do it in the buttons. Also updated them to accept `$$restProps` which I think we may want to do for every component so they can support all the various `aria-` attributes that may come up, plus anything else we don't want to specifically handle.


mcous (approved):

ethanlook (approved):

ehhong (approved):
this is so much better! thanks so much for updating this

PR #339, opened 2023-08-24
Add searchable select

DTCurrie:
Adds a searchable version of the select using an input and a custom menu:

Screenshot 2023-08-29 at 10 37 23 AM

Also includes a bundle of small fixes and adjustments that came up during review/testing.


packages/core/src/lib/select/select-menu.svelte
```diff
@@ -0,0 +1,72 @@
...
+  <ul
+    role="menu"
+    class="select-menu flex max-h-36 flex-col overflow-y-auto"
+    bind:this={element}
+    on:mouseleave
+  >
+    <slot />
+  </ul>
+  {#if button !== undefined}
+    <button
```
mcous, line 38: What's this button for? I think we'll need an explicit `type=button` to prevent it becoming an implicit submit button if it's used inside a `<form>`

packages/core/src/lib/select/select.svelte
```diff
@@ -25,9 +34,35 @@ export let value: string | undefined = undefined;
...
+const onMouseDown = (event: MouseEvent) => {
+  if (disabled) {
+    event.preventDefault();
+    return;
+  }
+
+  dispatch('mousedown', event);
+};
+
+const onKeyDown = (event: KeyboardEvent) => {
```
mcous, line 57: What are these `...Down` events used for? They'll differ from a `click` or a key press, in that both of those won't fire until the mouse/key has been released (`...Up`)

mcous (commented):
Initial thoughts

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -0,0 +1,273 @@
...
+    bind:button
+    on:buttonclick={handleButtonClick}
+    on:mouseleave={handleMouseLeave}
+    {open}
+  >
+    {#if sortedOptions.length > 0}
+      {#each searchedOptions as { search, option }, index (option)}
+        <li role="presentation">
+          <button
+            role="menuitem"
```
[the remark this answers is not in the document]
DTCurrie: Makes tabbing and focus logic work as expected, follow the pattern shown [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role#examples) which is similar to our requirements.  We could just skip the `<ul/>` entirely and just use a series of `<buttons />` but this seems to be a common pattern based on some googling. Definitely open to a better layout especially if we can reduce all of the nesting.

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -0,0 +1,273 @@
...
+    bind:heading
+    bind:button
+    on:buttonclick={handleButtonClick}
+    on:mouseleave={handleMouseLeave}
+    {open}
+  >
+    {#if sortedOptions.length > 0}
+      {#each searchedOptions as { search, option }, index (option)}
+        <li role="presentation">
+          <button
```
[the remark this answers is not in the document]
DTCurrie: The button is focusable by tab natively with this layout, so it shouldn't require a `tabindex`. If we wanted to skip the `button` and just use the `li` as a `menuitem` we would need a `tabindex` there and would need rely on `click` and `keydown` handlers instead of just the `button`'s click.

packages/core/src/lib/select/select-menu.svelte
DTCurrie: This is a bit of a holdover from the previous `select` but it is also shown in this layout here:

Screenshot 2023-08-24 at 2 53 34 PM

I don't love this setup really, would probably prefer to do it with slots. Actually maybe I will just do that.

packages/core/src/lib/select/select-menu.svelte
```diff
@@ -0,0 +1,72 @@
...
+.select-menu::-webkit-scrollbar-thumb {
+  background-color: black;
+  border-radius: 0;
+  border: 0 solid transparent;
+}
+
+.select-menu {
+  scrollbar-width: thin;
+  scrollbar-color: black transparent;
+}
```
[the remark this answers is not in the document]
DTCurrie: I wanted to add this as a tailwind thing because yeah I don't love adding these non-tailwind classes here. I'll look into the firefox setup, this was copied from the previous select so glad to know that never worked.

packages/core/src/lib/select/utils.ts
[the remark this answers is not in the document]
DTCurrie: We have added `utils` in a few places but I am happy to change it, don't really have a preference there. We have no test coverage yet for this particular one. I agree a lot of these are confusing, I just pulled them directly from the legacy code and cleaned them up. This is actually after some complexity reduction if you can believe it.

I am wondering if we would prefer to use a library for these fuzzy searches but I was having trouble finding a good one that works with the highlighting we are doing. Definitely open to suggestions there if you know any.

packages/core/src/lib/select/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+        index = i + 1;
+      }
+    }
+
+    if (results[index] === undefined) {
+      results[index] = [datum];
+    } else {
+      results[index]!.push(datum);
+    }
+  }
```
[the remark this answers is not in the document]
DTCurrie: Yeah this was just pulled over from the legacy select. I am still trying to simplify these/pull them apart as it is going to get worse with the multi-select. I'll give that a go.

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -0,0 +1,273 @@
...
+}
+</script>
+
+<svelte:body on:click={handleClickOutside} />
+<div
+  bind:this={wrapper}
+  class="relative flex w-full"
+>
+  <div class="flex w-full">
+    <input
```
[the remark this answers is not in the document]
DTCurrie: If I do this we get back into the case of needing `id`s to add the appropriate `aria-controls` attribute to point to the menu. So we need to make a call there.

packages/core/src/lib/select/select-menu.svelte
DTCurrie: Hm actually can't use slots all the way down because of rendering svelte fragments inside of conditionals, I will continue to explore some options but until then I can make sure we get the `type` on there.

packages/core/src/lib/select/select.svelte
DTCurrie: These are here to allow disabled selects to continue to be focusable/discoverable while behaving disabled. If we use the disabled attribute the select will not be discoverable, but selects don't have a readonly attribute like inputs. So what we want to do is capture events when the select is disabled and prevent them from doing anything if so. In the case of the `keydown` event we still want to allow the `tab` key to behave natively since we want to be able to tab away from disabled inputs. the `mousedown` one is to make sure you can't open the select menu. Otherwise if the select isn't disable we just dispatch the native event as usual.

packages/core/src/lib/select/select-menu.svelte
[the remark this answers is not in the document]
DTCurrie: Actually firefox was already covered, at least under the limited capacity allowed with `scrollbar-color`. It unfortunately appears rounded but it does at least render black:

Screenshot 2023-08-24 at 3 43 57 PM

packages/core/src/lib/select/utils.ts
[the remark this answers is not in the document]
DTCurrie: Ok so I moved this into its own file and left the other two functions in a `dom-utils` file, not sure how you feel about that but at least it is a bit more clear.

I went through the logic of this search logic, applied typing, and added a whole bunch of comments. We are still having to do the matching twice to manage prioritizing based on match for each word of an option and then match to determine how to highlight the text. It shouldn't be horrible and not a huge issue now, so if there isn't a super clear way to clean that up happy to put that in a TODO for later.

I think with the cleanup and renaming it is a little bit easier to understand. If we want, we can export all of the functions so they should be easy to test units.

packages/core/src/lib/select/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+  option: string;
+}
+
+const addSpecialCharacterEscapes = (value: string) =>
+  // eslint-disable-next-line no-useless-escape
+  value.replaceAll(/[-[\]{}()*+?.,\\^$|#]/gu, '\\$&').replaceAll(/\s/gu, '\\s');
+
+const matchEntries = (entries: string[], searchTerm: string) => {
+  const results: Record<string, string[]> = {};
+  const toMatch = addSpecialCharacterEscapes(searchTerm);
```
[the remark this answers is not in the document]
DTCurrie: `lodash`'s approach worked so going to stick with that to have less regex to maintain 🥳

packages/core/src/lib/select/utils.ts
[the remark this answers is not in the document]
DTCurrie: Ok so this actually has a very specific logic to it I had to untangle. So we prioritize results where the search term occurred at the beginning of any word in the option as the highest, then we prioritize results based on the index of the word where the match occurred, then any non-matches get the lowest priority. I cleaned it up a bit and added a bunch of docs to explain.

packages/core/src/lib/unique-id.ts
```diff
@@ -0,0 +1,41 @@
+import { setContext, getContext } from 'svelte';
+import { readable, type Readable, get } from 'svelte/store';
+
+type GetCount = () => number;
+type UniqueIdStore = Readable<GetCount>;
+
+const CONTEXT_KEY = 'prime-core-uid';
```
mcous, line 7: A `Symbol` is ever so slightly safer, since it cannot collide

```suggestion
const CONTEXT_KEY = Symbol('prime-core-uid');
```

packages/core/src/lib/unique-id.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Initialize a store to  track ID counts to ensure uniqueness
+ *
+ * ```ts
+ * // +layout.svelte
+ * import { initializeUniqueIdStore } from '$lib/unique-id';
+ * initializeUniqueIdStore();
+ * ```
+ */
+export const initializeUniqueIdStore = () => {
```
mcous, line 18: In the SvelteKit app, we've been naming things that provide a context with `provide`. So here:

```suggestion
export const provideUniqueID = () => {
```

packages/core/src/lib/unique-id.ts
```diff
@@ -0,0 +1,41 @@
...
+ * const first = uniqueId('first');     // returns first_1
+ * const second = uniqueId('second');   // returns second_2
+ * ```
+ */
+export const useUniqueId = (): ((prefix: string) => string) => {
+  const uidStore = getContext<UniqueIdStore>(CONTEXT_KEY);
+  const getCount = get(uidStore);
+
+  return (prefix: string) => `${prefix}_${getCount()}`;
+};
```
mcous, line 41: Does this need to return a function? I feel like in usage, this might be easier:

```suggestion
export const useUniqueId = (prefix: string) => string => {
  const uidStore = getContext<UniqueIdStore>(CONTEXT_KEY);
  
  return `${prefix}_${get(uidStore)}`
};
```


```svelte
<script lang="ts">
  import { useUniqueId } from '@viamrobotics/prime-core'

  const menuID = useUniqueID('my-menu')
</script>
```

packages/core/src/lib/unique-id.ts
```diff
@@ -0,0 +1,41 @@
...
+ * // some-component.svelte
+ * import { useUniqueId } from './unique-id';
+ *
+ * const uniqueId = useUniqueId();
+ * const first = uniqueId('first');     // returns first_1
+ * const second = uniqueId('second');   // returns second_2
+ * ```
+ */
+export const useUniqueId = (): ((prefix: string) => string) => {
+  const uidStore = getContext<UniqueIdStore>(CONTEXT_KEY);
```
mcous, line 37: Should this have a default fallback behavior if the context is not present? Either way, it adds a gotcha to implementation that needs to be documented

mcous (commented):
I see how the store-based solution works, but I don't love how it requires to add an extra context provider to use the some components in the library. Given that this implementation is not SSR stable, would it be better to simply go with `nanoid/non-secure` and forgo adding in a required context?

packages/core/src/lib/unique-id.ts
DTCurrie: I think so, this is basically pulled straight from that thread on the svelte issue. I am not sure with the context/store if there is any value in having it return a function vs just exposing one.

DTCurrie:
> I see how the store-based solution for unique IDs works, but I don't love how it requires to add an extra context provider to use the some components in the library. Given that this implementation is not SSR stable, would it be better to simply go with `nanoid/non-secure` and forgo adding in a required context?

We could also just have it track the count internally without a store, but then we will need to require it be imported via a `context="module"` script which could add a level confusion.

I am cool with just using `nanoid` or just a string generator function in JS if we aren't super worried about it.

packages/core/src/lib/select/searchable-select.svelte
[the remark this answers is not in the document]
mcous: Per [MDN's docs on `menu`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role):

> [tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#tabindex) attribute
The `menu` container has `tabindex` set to `-1` or `0` and each item in the menu has tabindex set to `-1`.

Inside a vertical `menu`, it seems like the recommended keyboard interaction (of MDN) is to use up and down arrows to navigate menu options, not tab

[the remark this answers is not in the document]
DTCurrie: You can navigate this with the up and down arrow keys and you cannot use tab. The menu is a `ul` which is non-interactive which is equal to `tabindex` of `-1` so I can add it but shouldn't change the behavior. You can use to tab to exit it along with escape, but because we capture and handle the tab event it is impossible to navigate to the buttons with it. I can set a `tabindex` of `-1` on the buttons but similarly it won't change the behavior.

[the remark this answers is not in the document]
DTCurrie: Interesting, I just tried removing the buttons as well and moving everything to the `li` and it all behaves as expected but setting the `height` on the `li` doesn't work:

Screenshot 2023-08-25 at 3 55 56 PM

packages/core/src/lib/select/__tests__/searchable-select.spec.ts
ethanlook, line 1: nit: consider using [`userEvent`](https://testing-library.com/docs/user-event/intro) for this testing to simulate interactions

ethanlook (approved):

packages/core/src/lib/select/search.ts
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * Returns the passed option broken down into segments to apply highlighting
+ * to the portion of the option that matches the passed search term.
+ */
+export const getSearchHighlight = (
+  option: string,
+  searchTerm: string,
+  { index }: RegExpExecArray
+): SearchHighlight | undefined => {
```
mcous, line 81: This doesn't look like it can return `undefined`

mcous (approved):
LGTM, agree with Ethan's comments

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -0,0 +1,250 @@
...
+export let state: SelectState = 'none';
+export let sort: SortOptions = 'default';
+export let button: { text: string; icon: string } | undefined = undefined;
+export let heading = '';
+
+const dispatch = createEventDispatcher<{
+  input: string | undefined;
+  change: string | undefined;
+  search: string;
+  buttonclick: undefined;
```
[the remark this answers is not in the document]
ethanlook: In the cases that I've wanted this, I've explicitly typed the return as `Record<string, never>` to signal to the consumer the lack of associated `detail`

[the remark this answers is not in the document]
DTCurrie: I read through a bunch on this and landed on this PR where the feature was added, specifically this comment: https://github.com/sveltejs/svelte/pull/7224#issuecomment-1620430992

Also with this, we signal to the user the detail is `undefined`: 
Screenshot 2023-08-29 at 1 10 28 PM

If I use `Record<string, never>` this is what we show the user, which is way less clear IMO:
Screenshot 2023-08-29 at 1 11 25 PM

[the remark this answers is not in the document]
DTCurrie: Though looking at the migration docs for `v4` it seems `null` is preferred so we can swap to that. Generally I prefer `undefined` but I don't mind sticking to their recommendation: https://svelte.dev/docs/v4-migration-guide#stricter-types-for-svelte-functions

packages/core/src/lib/select/select.svelte
```diff
@@ -25,9 +43,35 @@ export let value: string | undefined = undefined;
 export let disabled = false;
 
 /** The state of the select (info, warn, error, success), if any. */
-export let state: InputState = 'none';
+export let state: SelectState = 'none';
+
+const onInput = (event: Event) => {
```
micheal-parks, line 48: Why are the disabled checks necessary here? Shouldn't native selects already handle disabling input when disabled?

I'm primarily worried about how this degrades the experience of event forwarding, since these event objects are being wrapped in the custom event detail attribute.

If we end up finding this pattern necessary - ie intercepting events before forwarding, I think it would be valuable to replicate Threlte's [createRawEventDispatcher](https://threlte.xyz/docs/reference/core/utilities#createraweventdispatcher) in PRIME so that we can forward raw event objects.

DTCurrie: Because disabled inputs and selects are not focusable/keyboard navigable and completely hidden to screen readers.

micheal-parks: 🤦 Oh yes, keep forgetting about this problem. Then based on this and other discussions I've observed I'd recommend we replicate [createRawEventDispatcher](https://threlte.xyz/docs/reference/core/utilities#createraweventdispatcher) and maybe [forwardEventHandlers](https://threlte.xyz/docs/reference/core/utilities#forwardeventhandlers), I'd hate for us to end up back in the react space of having things like `nativeEvent` properties :(

DTCurrie: So I looked at how Threlte did it and they rely on `svelte/internal` which was hidden form the public API and is likely to change in v5, so I wonder if there is a better solution.

DTCurrie: Updated per our conversation.

micheal-parks (approved):

PR #340, opened 2023-08-24
Fix IconButton size

ethanlook:
The `IconButton` used to have size `30px x 30px` in legacy, but picked up `32px x 32px` in the migration. This keeps it consistent for the migration.


ehhong (approved):
pixel pushing!

ethanlook:
> pixel pushing!

taylor

PR #344, opened 2023-08-25
Slider core

Kschappacher:


packages/core/src/lib/__tests__/slider.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    const slider = screen.getByRole('slider');
+    expect(slider).toBeVisible();
+
+    // TODO Drag isnt working as expected
+    // Drag documentation https://testing-library.com/docs/example-drag/
+    // await fireEvent.mouseDown(slider);
+    // await fireEvent.mouseMove(document, { clientX: -5, y: 0 });
+    // await fireEvent.mouseUp(document);
+    await fireEvent.drag(slider, { delta: { x: -5, y: 0 } });
```
anjinai, line 188: @mcous - I was hoping you could give me some guidance on how to best mock dragging the slider for these tests.  Both Kurt and I were having difficulty.  

Note: I'll update the rest of the tests to match the guidelines/comments you gave me in my modal PR.

packages/core/src/lib/__tests__/slider.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+const customStart = -20;
+const customEnd = 30;
+const value = 15;
+const step = 25;
+
+export const delay = async (time: number) => {
+  await new Promise((resolve) => {
+    setTimeout(resolve, time);
+  });
+};
```
mcous, line 23: As of Node v18 (which we're on) you can use `node:timers/promises` for this instead: https://nodejs.org/dist/latest-v18.x/docs/api/timers.html#timerspromisessettimeoutdelay-value-options

packages/core/src/lib/__tests__/slider.spec.ts
mcous: My first reaction would definitely be to use [`userEvent`](https://testing-library.com/docs/user-event/intro/) rather than `fireEvent`; I think I see `userEvent` commented out above. It does a much better job emulating user interactions that trying to `fireEvent` all the different browser events manually

I'd hope something [like this](https://testing-library.com/docs/user-event/pointer#moving-a-pointer) could just work™️

anjinai:
@mcous - I tried to take your suggestion to use pointer from testing library.  Moving this ticket back to todo for now given it's listed as low priority, but if you have some time - it would be great to pair with you on this.

mcous:
Finally had the time to play around with this, and I agree that testing mouse interactions with the slider using `testing-library` isn't going to work. JSDOM has no layout engine, so the width of the slider in `vitest` is `0`.

There's a chance we can test mouse interactions, which we should do if it doesn't exhibit the same issues. Otherwise, I think there's consensus that Playwright needs to come into `prime-core` specifically for testing things like this that require a real browser and/or layout engine

PR #345, opened 2023-08-25
Bump vite in legacy

DTCurrie:
Resolves https://github.com/viamrobotics/prime/pull/342


mcous (approved):
Does vite (and/or vitest) need to be bumped in any other projects, too?

DTCurrie:
> Does vite (and/or vitest) need to be bumped in any other projects, too?

This puts everything inline to use the same version of `vite` which was ahead of the dependabot alert. I didn't want to push anything else with this.

PR #346, opened 2023-08-28
Add use:clickOutside action to prime-core

mcous:
## Overview

@DTCurrie and I have been thinking:

- "do something on click outside" is a common thing across our UIs
- [Svelte actions](https://svelte.dev/docs/svelte-action) (custom `use:...` directives) are the most reliable way to implement this, given reactivity and cleanup concerns
- The [classic click outside action example](https://svelte.dev/repl/0ace7a508bd843b798ae599940a91783?version=3.16.7) does not play nicely with TypeScript 

This PR takes as stab at adding a `clickOutside` action to `@viamrobotics/prime-core` in a type-checking-friendly manner.

```svelte
<script lang="ts">
import { writable } from 'svelte/store'
import { clickOutside } from '@viamrobotics/prime-core'

const isOpen = writable(false)
</script>

<button on:click={() => isOpen.set(true)}>
  Open Menu
</button>
{#if isOpen}
  <article use:clickOutside={() => isOpen.set(false)}>
    <h3>Cool menu</h3>
  </article>
{/if}
```

## Implementation notes

I elected _not_ to use a custom event - as in the example linked above - because it doesn't type-check without adding the custom event to _all_ DOM elements, which feels inaccurate because the event is only emittable if you add the action.

Until working on this, I didn't realize that Svelte actions can take arguments. This resolves the type-checking issue quite nicely and reduces boilerplate. I like how `use:clickOutside` subtly mirrors `on:click`, while making it clear that it's a custom thing that we added:

```svelte
<div
  on:click={() => console.log('hello')}
  use:clickOutside={() => console.log('goodbye')}
/>
```

### DIFF
--- a/packages/core/src/lib/click-outside.ts
+++ b/packages/core/src/lib/click-outside.ts
@@ -1,0 +1,55 @@
+import type { Action } from 'svelte/action';
+
+/**
+ * Trigger a callback when the user clicks outside the element.
+ *
+ * Use a [Svelte action](https://svelte.dev/docs/svelte-action):
+ * ```svelte
+ * <script lang="ts">
+ * import { writable } from 'svelte/store'
+ * import { clickOutside } from '@viamrobotics/prime-core'
+ *
+ * const isOpen = writable(false)
+ * </script>
+ *
+ * <button on:click={() => isOpen.set(true)}>
+ *   Open Modal
+ * </button>
+ * {#if isOpen}
+ *   <article use:clickOutside={() => isOpen.set(false)}>
+ *     <h3>Cool modal</h3>
+ *   </article>
+ * {/if}
+ * ```
+ *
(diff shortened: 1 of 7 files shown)


micheal-parks (approved):
Great design!

packages/core/src/lib/click-outside.ts
ethanlook, line 1: Nice! We should get rid of clickOutside in RDK: https://github.com/viamrobotics/rdk/blob/main/web/frontend/src/lib/click-outside.ts

DTCurrie: Definitely, I can put that on my list of changes I am doing.

mcous: Lol did not even know that was there 🫠

PR #347, opened 2023-08-29
Do not trigger click outside if element is removed from DOM

mcous:
Found an interesting edge case with the click-outside action: if you click and element that triggers a state change that causes the element to be removed from the DOM by Svelte, this DOM removal will race with the click event handler.

In the browser (or at least, my browser), Svelte will win and remove the element before the click handler runs. This causes the `node.contains(event.target)` condition to evaluate to false, even though the event target **used to be** in the parent node.

This PR adds an extra check to ensure the clicked event target is in the DOM. I had to get a little creative with the test, since in Node.js / JSDOM, the click handler wins the race. The test I wrote failed before the fix and passed afterwards. It will be covered in-browser by E2E tests in the app.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.19",
+  "version": "0.0.20",
   "publishConfig": {
     "access": "public"
   },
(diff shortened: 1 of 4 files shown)


ethanlook (approved):
Very interesting! Good find!

PR #348, opened 2023-08-30
Navigation map tweaks

micheal-parks:
A small collection of `<NavigationMap>` tweaks based on updated designs:

* Add a slot for a custom tab
* Update tab styles
* Fix a bug where axes would render with no obstacles
* Bump up lighting of obstacles a bit

### DIFF
--- a/packages/blocks/src/lib/navigation-map/components/nav/index.svelte
+++ b/packages/blocks/src/lib/navigation-map/components/nav/index.svelte
@@ -1,7 +1,6 @@
 <script lang='ts'>
 
 import { createEventDispatcher } from 'svelte';
-import { Tabs } from '@viamrobotics/prime-core';
 import { NavigationTab, type Obstacle } from '$lib';
 import { tab, tabs, hovered } from '../../stores';
 import ObstaclesTab from './obstacles.svelte';
--- a/packages/blocks/src/lib/navigation-map/components/nav/index.svelte
+++ b/packages/blocks/src/lib/navigation-map/components/nav/index.svelte
@@ -2,7 +2,7 @@
 
 import { createEventDispatcher } from 'svelte';
 import { Tabs } from '@viamrobotics/prime-core';
-import { NavigationTab, type NavigationTabType, type Obstacle } from '$lib';
+import { NavigationTab, type Obstacle } from '$lib';
 import { tab, tabs, hovered } from '../../stores';
 import ObstaclesTab from './obstacles.svelte';
 import WaypointsTab from './waypoints.svelte';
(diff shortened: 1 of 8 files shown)


packages/blocks/src/lib/navigation-map/components/nav/index.svelte
```diff
@@ -12,36 +11,52 @@ const dispatch = createEventDispatcher<{
...
+  </ol>
 
-  <ul
-    on:mouseleave={() => ($hovered = null)}
-    class='px-4 py-2 sm:h-[calc(100%-30px)] overflow-y-scroll'
-  >
-    {#if $tab === NavigationTab.Waypoints}
+  {#if $tab === NavigationTab.Waypoints}
+    <ul
+      on:mouseleave={() => ($hovered = null)}
```
mcous, line 40: Is this another way of doing `hovered.set(null)`? If so, I kinda prefer using `.set` explicitly. It's more portable to vanilla TS, and I think it does a better job of reminding future readers that `hovered` is a box, not a value

packages/blocks/src/lib/navigation-map/components/scene.svelte
```diff
@@ -28,6 +28,8 @@ useMapLibreEvent('move', () => {
 useFrame(() => {
   renderer.clear();
 
+  axes.visible = scenes.length > 0
```
mcous, line 31: [probably for another PR] To ensure this plays nicely with SvelteKit eventually, should these mutable global singletons be moved to a context?

packages/blocks/src/lib/navigation-map/index.svelte
```diff
@@ -36,10 +36,10 @@ export let waypoints: Waypoint[] = [];
 export let obstacles: Obstacle[] = [];
 
 /** The initial tab to show */
-export let tab: NavigationTabType = NavigationTab.Waypoints;
+export let tab: string = NavigationTab.Waypoints;
 
 /** The visible set of tabs */
-export let tabs = [NavigationTab.Waypoints, NavigationTab.Obstacles];
+export let tabs: string[] = [NavigationTab.Waypoints, NavigationTab.Obstacles];
```
mcous, line 42: What was the reasoning for loosening this to `string` rather than adding `Attributes` to the union?

mcous (approved):
LGTM

packages/blocks/src/lib/navigation-map/components/scene.svelte
micheal-parks: Yeah, this code is still making environment assumptions - and it should only ever run in the browser. Would like to pick your brain for following up since you've done some recent work with isomorphic sveltekit rendering

packages/blocks/src/lib/navigation-map/components/nav/index.svelte
micheal-parks: I'm fine with both `$` and the vanilla syntax, although to me `$hovered = null` is a very transparent sugar. The reason I was sticking with `$` is consistency with the read ops above, which I wouldn't want to change since I think it's much more syntactically elegant than `.subscribe()`. Can change to `.set()` though if you or other reviewers are more comfortable with it

PR #349, opened 2023-08-30
Add multiselect

DTCurrie:
Screenshot 2023-08-30 at 1 16 53 PM

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.21",
+  "version": "0.0.22",
   "publishConfig": {
     "access": "public"
   },
--- a/packages/core/src/lib/__tests__/pill.spec.ts
+++ b/packages/core/src/lib/__tests__/pill.spec.ts
@@ -25,9 +25,8 @@ describe('Pill', () => {
 
   it('Renders a normal pill if no disabled attribute has been specified', () => {
     render(Pill, { value: 'test' });
-    expect(screen.getByText('test').parentElement).toHaveAttribute(
-      'aria-disabled',
-      'false'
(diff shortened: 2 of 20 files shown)


micheal-parks:
I'd recommend moving the dropdown to above / covering the pills. Had a few discussions with design during building the original multiselect about whether or not to do this and we agreed that having both the pills and the dropdown visible was a worse UX because we're showing two element sections at once that have redundant functionality (you can remove a pill both in the dropdown and by clicking x)

DTCurrie:
> I'd recommend moving the dropdown to above / covering the pills. Had a few discussions with design during building the original multiselect about whether or not to do this and we agreed that having both the pills and the dropdown visible was a worse UX because we're showing two element sections at once that have redundant functionality (you can remove a pill both in the dropdown and by clicking x)

Ah ok, I was going off the old designs since new ones haven't been provided for multiselects yet. Kind of hesitant to stray too far from those until we have proper figmas, but I can update this assuming there isn't any other big design changes expected.

DTCurrie:
> > I'd recommend moving the dropdown to above / covering the pills. Had a few discussions with design during building the original multiselect about whether or not to do this and we agreed that having both the pills and the dropdown visible was a worse UX because we're showing two element sections at once that have redundant functionality (you can remove a pill both in the dropdown and by clicking x)
> 
> Ah ok, I was going off the old layout since new ones haven't been provided for multiselects yet. Kind of hesitant to stray too far from those until we have proper figmas, but I can update this assuming there isn't any other big design changes expected.
> 
> We stopped using the dropdown, so that may be why it is different.

Updated:

Screenshot 2023-08-30 at 1 51 12 PM

ethanlook (approved):
Looks great! Nice code sharing.

Don't forget to bump the version!

ethanlook (approved):

PR #350, opened 2023-08-30
Add NotificationContainer and notify

DTCurrie:
Adds the `<NotificationCenter />` which when added to the DOM, will allow users to show `<Banner />`s as notifications using the `notify` API.

Major changes:
- `<Notify />` -> `<Banner />` because having a `Notify` component and a `notify` API was confusing, also feels more accurate.
- `<Banner />` now accepts `title` and `message` slots instead of props to allow "~r~ _i_ **c** `h`" content and components to be used.

- [x] Add docs


ethanlook:
> * `<Notify />` -> `<Banner />` because having a `Notify` component and a `notify` API was confusing, also feels more accurate.

Works for me! The component that is now called `Banner` is already being used in App, so we'll need to update some uses when we upgrade.

packages/core/src/lib/banner.svelte
```diff
@@ -0,0 +1,146 @@
...
+
+<script lang="ts">
+import cx from 'classnames';
+import { createEventDispatcher } from 'svelte';
+import { Button, Icon } from '$lib';
+
+/** The severity of the notification you want to show users*/
+export let variant: BannerVariant;
+
+/** The scaling applied on the y access of the page*/
```
ethanlook, line 18: s/access/axis

ethanlook (approved):
Looks awesome! Thanks! Will it be possible to replace all existing uses of notify with this in app, even in the `frontend` directory?

Reminder to update the package version!

DTCurrie:
> > * `<Notify />` -> `<Banner />` because having a `Notify` component and a `notify` API was confusing, also feels more accurate.
> 
> Works for me! The component that is now called `Banner` is already being used in App, so we'll need to update some uses when we upgrade.

Yeah I considered that but I figured it is better to pull the band-aid on these changes now before we wrap the migration. I considered keeping an marking the `Notify` as `@deprecated` but I didn't want it to linger.

packages/core/src/lib/banner.svelte
DTCurrie: Didn't notice that when I was refactoring, good catch!

DTCurrie:
> Looks awesome! Thanks! Will it be possible to replace all existing uses of notify with this in app, even in the `frontend` directory?

That's a good question. I don't see why not but we may have to test that out.

ethanlook:
> pull the band-aid on these changes now

I totally agree, we should definitely make these changes now, just wanted to raise awareness so it's not a surprise when we inevitably upgrade app.

PR #351, opened 2023-08-31
Add missing text input bindings

micheal-parks:


ethanlook (approved):
Is there a test we can add to validate this?

micheal-parks:
@ethanlook yep, good idea, will add in a follow up tomorrow 👍

DTCurrie:
Interesting, since the `Input` had these declared I figured they would get passed through by `$$restProps`.

PR #352, opened 2023-09-05
Fix breadcrumb padding and select event handling

DTCurrie:
I noticed after bumping `prime-core` in app we had some errors popping up around the new event handling. After some experimentation the pattern I used for the `<Select />` seems to be the best way to intercept native event handling without overriding it or forcing custom events. We capture and prevent default and propagation of the event on the capture phase before letting whatever native event was passed to the component do its thing on the bubble phase. Added some tests around that as well.

Also updated the padding around the breadcrumb text to match the padding in designs. I noticed they were wider than expected when using them on the module details page in the viam app.

### DIFF
--- a/packages/core/src/lib/prevent-handler.ts
+++ b/packages/core/src/lib/prevent-handler.ts
@@ -1,0 +1,44 @@
+/**
+ * Returns an event handler that will prevent default behavior and stop
+ * immediate propagation of an event if `prevent` is true. Useful to layer
+ * disabled behavior onto native event forwarding.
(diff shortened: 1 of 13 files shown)


ethanlook (approved):
Nice simplification with the prevent handlers!

packages/core/src/lib/prevent-handler.ts
ethanlook, line 1: These could be helpful in app as well. Do you think they are worth exporting from the package?

PR #354, opened 2023-09-06
APP-2206 Add rounded-none to select

ethanlook:
Safari adds a `5px` border radius to selects (because of course it does.) Sharp corners for all!


micheal-parks (approved):
🔲 

Huh, surprised tailwind resets don't cover that

PR #355, opened 2023-09-07
Fix theme z-indexes

micheal-parks:
I accidently added a `z-index` prop to the theme's root object, which obliterates all other z-index classes. It's still in `extend`.


ethanlook (approved):

DTCurrie (approved):
✅

PR #359, opened 2023-09-08
Add cx prop to all components to apply additional css classes

DTCurrie:
It keeps coming up as I work in the RC and with the MRR work, so going ahead and adding it.


packages/core/src/lib/collapse.svelte
```diff
@@ -47,7 +47,7 @@ const handleClick = () => {
     role="button"
     aria-label="Toggle Content"
     tabindex="0"
-    class="flex w-full cursor-pointer items-center justify-between border border-light bg-white px-4 py-2 text-default"
+    class="border-light text-default flex w-full cursor-pointer items-center justify-between border bg-white px-4 py-2"
```
mcous, line 50: [not for this PR] Oof, something is messed up with out prettier / eslint tailwind plugins. The fact that `border-light` and `text-default` got shunted to the front mean those plugins aren't reading our tailwind config properly.

Something to deal with sooner rather than later I think

DTCurrie: I have been seeing a lot of churn around class ordering so that makes sense. I can try to look into it.

PR #362, opened 2023-09-11
Configure prettier and eslint to check .svelte files

mcous:
## Overview

**.svelte files have not been running through ESLint nor Prettier via `scripts` and CI.**

This PR fixes the misconfigurations and addresses all the exposed formatting / lint errors.

- Updates shared configs to latest
- Aligns the `prettier` and `eslint` scripts with our other FE projects to use explicit globs rather than tool defaults
- Fixes format, lint, and typecheck issues
    - Or, in the case of `legacy`, ignores them

## Review requests

Since this PR touches pretty much every file in this repository, I split out each project into its own diff to ease reviewing. This PR will be the one that gets merged, but feel free to leave comments anywhere

- Config files: #367
- Core: #368 
- Blocks: #369
- Legacy + Storybook: #370


DTCurrie:
Did a quick scan and things look good. Lots of `type` -> `interface`, tailwind class reordering, single to double quotes, and adding missing semi-colons. I will feel less crazy with unexpected diffs now:  [](https://y.yarn.co/409b1182-1bc8-4013-a9ad-a23dfdcda39d_text.gif)

appreciates-you

DTCurrie (approved):

packages/core/src/lib/input/input.svelte
```diff
@@ -28,10 +28,10 @@ import cx from 'classnames';
 export let value: string | number | undefined = '';
 
 /** Whether or not the input should be rendered as readonly and be operable. */
-export let readonly: boolean | undefined = false;
+export let readonly = false as boolean | undefined;
```
micheal-parks, line 31: this is interesting, why is casting preferred?

micheal-parks (approved):
Just read through all of it, everything looks good 🚀 🚀 

Can we get this merged quickly? I'm making pretty heavy edits in the navigation card right now and would like to get probably a large amount of conflicts out of the way.

ethanlook (approved):
:shipit: 🚀 

Let's do it! Thank you!!

packages/core/src/lib/input/input.svelte
micheal-parks: ahhhh got it, thanks for the context!

PR #363, opened 2023-09-11
Make preventHandler use reactive

ethanlook:
The handlers that used `preventDefault` would be stuck in their original state because the use of `preventHandler` wasn't reactive.


DTCurrie (approved):
✅

DTCurrie:
Sorry just merged a PR that also bumped to the same version.

mcous (approved):
Good catch. I think this issue probably affects other functions in the codebase; I'm definitely guilty of writing a non-reactive handler for click-outside logic before we had the action-based approach

DTCurrie:
> Good catch. I think this issue probably affects other functions in the codebase; I'm definitely guilty of writing a non-reactive handler for click-outside logic before we had the action-based approach

Yeah I will keep an eye out for these as I go through the RC but we may want to make a pass of all the current stuff.

ethanlook:
> Sorry just merged a PR that also bumped to the same version.

No worries!

PR #364, opened 2023-09-11
[APP 2658] - Navigation card pass 2

micheal-parks:
This PR is primarily a second pass on the obstacle renderer to make it more performant and flexible in preparation coming features like motion path visualization, transient obstacles, and users drawing obstacles. 

Obstacles are now rendered within a single THREE.Scene - instead of each obstacle within its own scene - with the origin of the scene at a lng,lat that represents the center of the visible map.

It also captures a few cleanup items for the navigation card:
* Introduces obstacle interactivity (users can click / hover on obstacles and get feedback)
* Fixes an issue where the map and obstacles were not rendering in lockstep
* Removes the feature flag for 3d view, it's ready to use at this point


packages/blocks/src/lib/navigation-map/components/axes-helper.svelte
```diff
@@ -0,0 +1,45 @@
+<!--
```
micheal-parks, line 1: I'm going to probably submit a PR for this to be moved into `@threlte/extras`, but building it here for now just because this repo moves faster

packages/blocks/src/lib/navigation-map/components/obstacle.svelte
```diff
@@ -22,39 +23,78 @@ $: (material as THREE.MeshPhongMaterial | undefined)?.color.set(
   <T.Mesh
     {name}
     obstacle={name}
-    lnglat={obstacle.location}
+    userData.lngLat={obstacle.location}
```
DTCurrie, line 30: Why the dot notation on the attribute?

DTCurrie (approved):
✅

mcous (approved):

PR #365, opened 2023-09-11
cx tests and fixes

DTCurrie:
Noticed some bugs with the `cx` props I added so added some tests and fixes.


packages/core/src/lib/__tests__/cx-test.ts
DTCurrie, line 1: Willing to toss these out if we feel they aren't helpful, made them so it was easier to quickly write up all the tests but we can add these test-by-test if we don't want to keep them.

mcous (approved):

PR #366, opened 2023-09-11
Remove collapse (moving to RC)

DTCurrie:
After a few discussions I am removing the RC:

1. It is more complicated than what make sense for `core`
2. It is only used in `remote-control` with the upcoming robot detail design pages in `app` so doesn't make sense to make it a `block`
3. With svelte, a generic collapse that doesn't care about what the target and content look like is just as simple as creating a boolean store and toggling it, so not really worth a component.

If we decide we do want to introduce a `collapse` API, we have discussed using a hook or something similar.


ethanlook (approved):
👋🏼

PR #371, opened 2023-09-12
Fix some cx imports

ethanlook:
Some incorrect imports slipped through - I checked them all and every story seems to render correctly now!


packages/storybook/src/stories/icon/icon.stories.mdx
ethanlook, line 1: The `WithStyles` component was removed in favor of the `cx` control

DTCurrie (approved):
✅

PR #374, opened 2023-09-13
APP-2293: Migrate modal to @viamrobotics/prime-core

anjinai:
Migrates modal component to `@viamrobotics/prime-core`.

## Change log

- Copied `modal.svelte` to `packages/core.`
- Updated unit tests.
- Created storybook stories to `packages/storybook.`
- Bumped prime core version.

### Notes

- N/A

## Testing

- Added unit tests

## Review requests

- None

[APP-2293]: https://viam.atlassian.net/browse/APP-2293


packages/core/src/lib/modal.svelte
```diff
@@ -0,0 +1,108 @@
...
+    handleCloseModal();
+  }
+};
+
+onMount(() => {
+  document.body.classList.toggle('overflow-hidden', open);
+});
+</script>
+
+<div
```
anjinai, line 64: I'm getting some A11y errors.  For example, A11y: Non-interactive element <div> should not be assigned mouse or keyboard event listeners.(a11y-no-noninteractive-element-interactions).  Should I make this div a button?  If I do that, the cursor makes the background look clickable, which looks a bit strange.  Anyone have any suggestions?

ethanlook: For the mouse handler, you could use [`clickOutside`](https://github.com/viamrobotics/prime/blob/main/packages/core/src/lib/click-outside.ts) on the modal element.

For the keyboard handler, you could add the handler `onMount` to the `window`.

ethanlook (approved):

packages/core/src/lib/modal.svelte
anjinai: Ty ty!  Will try those!

PR #376, opened 2023-09-14
Harden types for icon names

ethanlook:
This is a minor improvement that will hopefully improve quality of life when using icons. Icons now must be one of the available icon names instead of any `string`. Autocomplete will be more helpful and we won't ever ship code that has missing icons!

image


DTCurrie (approved):
🚀

PR #377, opened 2023-09-15
APP-2703: rewrite core notifications to be context based

mcous:
## Overview

So we can use the core notifications in the SvelteKit app, they needed to change from a global-state based solution to a context-based solution. This PR makes that change and fixes up some a11y issues with the existing `Notification` and `Banner` components.

## Change log

See comments below

- Switch `core`'s `notify` API to a context-based `provideNotify` and `useNotify` approach for SvelteKit compatibility
- Tweak roles of the presentational `Banner` component for better a11y

## Review requests

Unit testing these components was prohibitively difficult due to their animations and Svelte's tendency to not animate in a Node.js environment (because it thinks it's SSR).

I think we might want to consider bringing Playwright back into the mix for these sorts of concerns, while keeping most tests in Vitest.


packages/core/src/lib/banner/__tests__/banner.spec.svelte
```diff
@@ -12,10 +12,9 @@
 
 <script lang="ts">
 import type cx from 'classnames';
-import { Banner } from '$lib';
-import type { BannerVariant } from '$lib/banner.svelte';
+import { Banner, type BannerVariantType } from '$lib';
 
-export let variant: BannerVariant;
+export let variant: BannerVariantType;
```
mcous, line 17: This would be a little breaking change, if we imported this type anywhere. I don't think we do, though

packages/core/src/lib/banner/banner.svelte
```diff
@@ -110,10 +104,7 @@ $: {
         />
       {/if}
 
-      <figure
-        class="flex flex-col"
-        role="alert"
```
mcous, line 115: From what I can tell from reading the specs, for `aria-live` regions like `alert` and `status`, screen readers announce dynamic changes to the elements, which means the `role=alert` element needs to be on the page first, _then_ the alert contents need to appear.

This, combined with the highly presentational nature of `Banner` lead me to make the following changes:

- I removed `role="alert"` from `Banner`
- I added `role="alert"` to `NotificationContainer`

This marks the `NotificationContainer` as a live element, and any time a new `Banner` (via `NotificationItem`) appears in the container, it should be announced as an alert. 

One remaining question: why is `Banner` a `<figure>` + `<figcaption>`? I haven't seen them used like this before for text

packages/core/src/lib/banner/variants.ts
mcous, line 1: I would have preferred to leave these in `banner.svelte`, but ESLint did not like it and couldn't recognize the exports, complaining at import sites that they were undefined. There isn't much harm in splitting them out to a pure TS file, and our toolchain was a lot happier with it

packages/core/src/lib/notification/context.ts
```diff
@@ -0,0 +1,150 @@
...
+
+/** Internal notification context. */
+export interface NotifyContext {
+  state: NotifyState;
+  notify: Notify;
+}
+
+/**
+ * Notification sender.
+ */
```
mcous, line 23:
```suggestion
/** Notification sender. */
```

packages/core/src/lib/banner/banner.svelte
DTCurrie: > One remaining question: why is Banner a <figure> + <figcaption>? I haven't seen them used like this before for text

I actually meant to ask about this when I did the `Notify -> Banner` change and totally forgot. It looks like this is a carry over from `legacy` but I also am not sure why they are being used. Since they are not being used to caption any specific piece of media, feels like they could be replaced.

DTCurrie (approved):
✅

packages/storybook/src/stories/notify.stories.mdx
```diff
@@ -1,11 +1,65 @@
...
+```ts
+import { useNotify } from '@viamrobotics/prime-core';
+
+const doSomethingCrazy = () => {
+  const notify = useNotify();
+
+  notify.info('Plain slice please!');
+  notify.warn('Hello, Brooklyn!');
+  notify.danger("I'm walkin' here!");
+  notify.success('Fuhgeddaboudit!');
```
ethanlook, line 60: I'm dying 🤣

ethanlook (approved):

micheal-parks (approved):

PR #378, opened 2023-09-18
Remove font overrides and use defaults

ethanlook:
This means monospace will be used for code blocks as well.

image


mcous (approved):

ethanlook:
> What's the reason behind doing this? This was something Nat Hartman and I intentionally set up.

Code samples were not being rendered in monospace.

micheal-parks:
There should be a way to target only the code samples. This current change I'd consider more of a step back than forward, since storybook is a good staging ground for ensuring design fidelity, and design fidelity is tied closely to our font choices.

ethanlook:
> There should be a way to target only the code samples. This current change I'd consider more of a step back than forward, since storybook is a good staging ground for ensuring design fidelity, and design fidelity is tied closely to our font choices.

Changing the font here only affects the docs themselves, not the font used in the components themselves. I would not consider the fonts in the documentation to be necessary for ensuring design fidelity.

Targeting the code samples does not work because of these overrides.

micheal-parks:
> Changing the font here only affects the docs themselves, not the font used in the components themselves

Nice! Ok, this wasn't clear to me since this used to not be the case. Thanks for explaining, all fears assuaged 🎉

micheal-parks (approved):

ethanlook:
> Random thought: we should really get a `deploy` tag on Prime so that I don't bombard PRs with pointless comments 😂

Agreed! Great DX improvement

PR #380, opened 2023-09-20
[APP 2769] - Refactor renderer / fix rendering precision issues

micheal-parks:
This PR primarily focuses on fixing precision issues that would become obvious when obstacles would have a smaller than 10 meter radius. The list of fixes or improvements are:

* Switch from using my back-of-the-envelope calculations for converting between latitude, longitude and cartesian coordinates to maplibre's mercator unit math methods.
*  Render the obstacle layer only after receiving a new camera projection from maplibre, instead of using Threlte's render loop.
* Use a raw `THREE.Camera` for rendering the scene so that the camera's projection matrix isn't altered by internal `THREE.PerspectiveCamera` behavior.

All in all this makes our obstacle positions very precise based on my testing! 🥳 🎉


packages/blocks/src/lib/navigation-map/components/axes-helper.svelte
```diff
@@ -29,7 +29,7 @@ export let length = 10;
   scale.z={thickness}
   position.y={length / 2}
 >
-  <T is={new THREE.MeshBasicMaterial({ color: 'green' })} />
+  <T is={new THREE.MeshBasicMaterial({ color: 'blue' })} />
```
micheal-parks, line 32: These were incorrect. x,y,z should correspond to r,g,b, so I quickly fixed it up!

mcous (approved):
Code LGTM

PR #381, opened 2023-09-20
APP-2596: fix tooltip positioning and functionality

mcous:
## Overview

This PR addresses APP-2596 moving forward by ensuring prime-core tooltips cannot overflow the page and addresses some a11y issues with the previous component.

(The overflow issue appears to have been worked-around in Vue during design syncs by adding various `overflow-x-hidden` classes to various containers).

## Change log

- Fix: ensure position is reactively computed when components mount instead of on first hover
- Fix: ensure tooltip content is rendered even when hidden for proper positioning calculation
- Fix: change the target container element from `button` to `span` so tooltips can wrap `<a>`
- Fix: allow `placement` to switch from left/right to top/bottom if the tooltip doesn't fit (and vice versa)
- Fix: simplify arrow element to avoid a thin gap between the tooltip and the arrow
- Refactor: move `floating-ui` usage into its own store
- BREAKING: do not set `aria-describedby` on the target container
- BREAKING: the tooltip wrapper element is no longer focusable
- BREAKING: rename "text" slot to `description` to better align with `aria-describedby`

### Usage

This new usage is slightly more verbose, and requires a little more care during usage to ensure proper a11y. That being said, the previous version often made it impossible to do correct a11y by forcing everything that has a tooltip to be a `button`.

- You must now specify `aria-describedby`
- You must now add `tabindex` yourself if the wrapped element is not normally focusable

I think using [slot props](https://svelte.dev/docs/special-elements#slot-slot-key-value) to pass the tooltip ID to the slot children is the best path forward here, because:

- It allows the component author to decide what's being described
- It's more terse than a "hooks"-based approach
- Unlike a `use:`-directive approach, it can be used with non-native elements 

New usage:

```svelte
<!-- Svelte -->
<Tooltip location="left" let:tooltipID>
  <p tabindex="0" aria-describedby={tooltipID}>
    This element needs some additional context.
  </p>
  <p slot="description">This is the tooltip text!</p>
</Tooltip>

<!-- HTML (paraphrased) -->
<span>
  <p tabindex="0" aria-describedby="tooltip-xyz">
    This element needs some additional context.
  </p>
</span>

<div role="tooltip" id="tooltip-xyz">
  <p>This is the tooltip text!</p>
</div>
```

Old usage:

```svelte
<!-- Svelte -->
<Tooltip location="left">
  <p>This element needs some additional context.</p>
  <p slot="text">This is the tooltip text!</p>
</Tooltip>

<!-- HTML (paraphrased) -->
<button aria-describedby="tooltip-xyz">
  <p>This element needs some additional context.</p>
</button>

<div role="tooltip" id="tooltip-xyz">
  <p>This is the tooltip text!</p>
</div>
```

### Future changes

- Consider adding [`autoUpdate()`](https://floating-ui.com/docs/autoUpdate) into the mix
- Consider adding a `target` slot to assign the _visual_ target of the tooltip
    - e.g. the tooltip describes an input, but is visually rooted on an info icon 

## Review requests

- Do the a11y changes make sense?
- Does the dev page look good?
- Does the storybook look good?


ehhong:
from playing around on the dev playground:

I changed the tooltip section to be `flex flex-col` and saw the following behavior:
- if left tooltip overflows, cannot scroll to see the full contents of the tooltip
image

- on Chrome, the right tooltip appears on the left if it overflows. on Safari, the right tooltip appears on the right if it overflows and you can scroll to see the full contents

**Chrome**
image

**Safari**
image

[design nit] this is so small but on the left and right tooltips, it looks like either the box and triangle have a outline color on them, or the triangle is slightly too far from the box. i don't see this on the top or bottom tooltips.
image
image

ehhong:
realized the previous behavior was from code off of main 😅 i think they still apply but i found a couple more things now that i'm actually testing code from this PR
- tooltip doesn't seem to be working at all on Safari (in playground, no tooltip appears on hover)
- on Chrome, the size of the element seems to change on hover, causing a lil wiggly dance on hover:
[vertical wiggle] https://github.com/viamrobotics/prime/assets/31592018/432acecf-e788-42f0-afdf-b14e222670a9
[horizontal wiggle] https://github.com/viamrobotics/prime/assets/31592018/077aaa7b-6805-4629-9dc0-304ef3a9894d

mcous:
Thanks @ehhong! Drafting PR while issues are addressed

mcous:
CC @ehhong, thanks again for the testing! Here's what I've got:

> tooltip doesn't seem to be working at all on Safari (in playground, no tooltip appears on hover)

I can't reproduce this on macOS 13.5 / Safari 16.6; Safari seems to be working for me. What version are you on?

> on Chrome, the size of the element seems to change on hover, causing a lil wiggly dance on hover:

Oops! This was me trying to get a little too clever. I had an extra element wrapping the tooltip to try to additionally prevent overflow. I removed the extra element and instead made sure that the `floating-ui` placement runs even if the tooltip is hidden.

> I changed the tooltip section to be flex flex-col and saw the following behavior:
>
> - if left tooltip overflows, cannot scroll to see the full contents of the tooltip
image
> - on Chrome, the right tooltip appears on the left if it overflows. on Safari, the right tooltip appears on the right if it overflows and you can scroll to see the full contents

The best solution I could figure out for this problem (which also exists on `main`) is to enable the [`fallbackAxisSideDirection` option for `flip`](https://floating-ui.com/docs/flip#fallbackaxissidedirection) in `floating-ui`. This tells `floating-ui` it's ok to move the tooltip to the `top` or `bottom` if the placement is `right` or `left` but there's no room on either (and vice versa).

> [design nit] this is so small but on the left and right tooltips, it looks like either the box and triangle have a outline color on them, or the triangle is slightly too far from the box. i don't see this on the top or bottom tooltips.

Also present in `main`! Resolved by changing the arrow from a triangle to a diamond (full rotated square). This also (happily) simplifies the styling code, because a diamond is valid for all directions and doesn't need additional rotation. (Idea stolen from the [floating-ui tutorial](https://floating-ui.com/docs/tutorial#arrow-middleware))

ethanlook (approved):

packages/core/src/lib/tooltip/tooltip.svelte
```diff
@@ -0,0 +1,97 @@
...
+  <slot {tooltipID} />
+</span>
+
+<div
+  bind:this={tooltip}
+  id={tooltipID}
+  role="tooltip"
+  class:invisible={!isVisible}
+  class="absolute left-0 top-0 z-max w-max max-w-[250px] border border-gray-9"
+  style:top={$styles.tooltip.top}
```
micheal-parks, line 80: I was not aware of this style syntax, very cool!

micheal-parks (approved):

PR #382, opened 2023-09-21
Add obstacle draw tools

micheal-parks:
Let there be drawing!

obstacles


raybjork (approved):

nfranczak (approved):
looks superrrrrrr awesome!
question: if I create a geometry, how is that translated into the navigation service config?

micheal-parks:
> question: if I create a geometry, how is that translated into the navigation service config?

This component will emit an `update` event with the modified geometries. That information is merged into the robot config when this component is embedded in app.

packages/blocks/src/lib/navigation-map/components/draw-tool.svelte
mcous, line 1: I see a fair amount of geometry math in the component. Should that math be moved into a sibling module and unit tested?

packages/blocks/src/lib/navigation-map/components/obstacle.svelte
```diff
@@ -1,26 +1,145 @@
...
+    // Scale
+  } else if (event.originalEvent.altKey) {
+    pointermove.set(event.point.x, event.point.y);
+    pointermove.sub(pointerdown);
+
+    const { y } = pointermove;
+
+    switch (obstacle.geometries[0]!.type) {
+      case 'sphere': {
+        obstacle.geometries[0]!.radius = Math.max(0, pointerdownRadius - y);
```
mcous, line 70: This is mutating an object (kind deep in its graph) that is passed in as a prop. This seems like a relatively easy place for a bug to sneak into in the future.

As we discussed in the UI guild a couple weeks ago, what could we do here to avoid mutating objects that are passed around between components?

An immutable approach here feels a bit heavy-handed, given the depth of changes. Given that we want to notify the parent when this changes, would changing this to some sort of reactive store be better? Mutating inside a `set` or `update` isn't an issue, it could give us a helpful place to put some of this math, too, which would be easier to test, and I think it makes it more explicit that "this child will be changing the value inside this box"

mcous (commented):
How can we

packages/blocks/src/lib/navigation-map/components/obstacle.svelte
micheal-parks: >  would changing this to some sort of reactive store be better

Yeah, I had considered this, I think your arguments here have finally convinced me, will update this!

packages/blocks/src/lib/navigation-map/components/draw-tool.svelte
micheal-parks: Yeah I also think that I could provide a little more insight into what's happening here with descriptive function names if they were moved, it's not obvious that most of this is translations from LngLat <-> Mercator units <-> Meters. Will update!

packages/blocks/src/lib/navigation-map/components/draw-tool.svelte
```diff
@@ -0,0 +1,105 @@
...
+        args={[width, height, 10]}
+        on:create={handleGeometryCreate}
+      />
+    {:else}
+      <T.PlaneGeometry
+        args={[width, height]}
+        on:create={handleGeometryCreate}
+      />
+    {/if}
+    <T is={new THREE.MeshPhongMaterial({ color: 'red' })} />
```
ethanlook, line 103: Nit: red means error and generally sounds alarm bells for me when it is used in a UI. This doesn't feel like danger. Is there another color we can use?

ethanlook:
This change has sufficiently convinced me that we should have a test environment for the PRIME storybook. I made a ticket: https://viam.atlassian.net/browse/APP-2847

micheal-parks:
@ethanlook Best ticket ever ❤️

packages/blocks/src/lib/navigation-map/components/draw-tool.svelte
micheal-parks: Agreed with this principle. The main reason why we chose red was to easily contrast with the map color palette. I'll revisit this though with Peter and Nat to see if we can come up with something better.

packages/blocks/src/lib/navigation-map/components/obstacle.svelte
micheal-parks: Hmmm switching to a store is causing more complexity / problems than it seems to be solving. Since prop editing is occuring only in this component I'm going to merge as is and schedule a session with you to revisit since these comments keep popping up in PRs and I'd like to find a good pattern that we both like :)

mcous: Could the change event simply be the change data and then the parent can do the mutation?

micheal-parks: That makes sense, thanks for the breakdown! I update the component to read and write from the store value, obstacles are no longer passed from parent to child and then mutated here.

PR #384, opened 2023-09-22
APP-2206: break apart tooltip component for custom usage

mcous:
## Overview

Following up on #381, as I went to integrate the updated Tooltip, I saw that several Svelte usage sites of this tooltip would benefit from the "separate target from described element" work I mentioned in that ticket.

This PR breaks the `<Tooltip>` into its constituent parts:

- `<TooltipContainer>` - a stateful wrapper
- `<TooltipTarget>` - a wrapper that attaches mouse/focus event listeners
- `<TooltipText>` - a presentational tooltip text component
 
You can continue to use `<Tooltip>`, or if that doesn't work for the UI in question, assemble the three parts however you need. In practice, I think this will mostly be "the visual target is different that the element on which we need to set `aria-describedby`.

I also upgraded the style calculation logic to automatically adjust scrolls/resizes/moves via floating-ui's [`autoUpdate`](https://floating-ui.com/docs/autoUpdate).

## Change log

- Add `TooltipContainer`, `TooltipTarget`, and `TooltipText` components that can be composed to create custom tooltips
- Wire in `autoUpdate` so the `recalculateStyle` component export isn't needed

## Review requests

I have updated the playground and the Storybook, which are good places to start with this PR. The actual positioning logic is hard to test without Playwright, but if you do see anything that could use unit testing that isn't currently tested, please call it out!

The best way to test the `autoUpdate` functionality:

- Test the `<SliderInput>`
- Scroll up and down on the dev playground, where I've added a permanently visible tooltip

### DIFF
--- a/packages/core/src/lib/input/slider-input.svelte
+++ b/packages/core/src/lib/input/slider-input.svelte
@@ -57,7 +57,6 @@ const dispatch = createEventDispatcher<{
   change: number | undefined;
 }>();
 
-let numberDragTooltip: Tooltip;
 let numberDragCord: HTMLDivElement;
 let numberDragHead: HTMLDivElement;
 let isDragging = false;
@@ -114,13 +113,6 @@ $: handlePointerMove = (event: PointerEvent) => {
     value = next;
     dispatch('input', value);
   }
-
-  /**
-   * TODO: Determine why this is being interpreted as an `any` type by the
-   * linter when it is of `() => Promise<void>`.
-   */
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-  numberDragTooltip.recalculateStyle();
 };
 
 const handlePointerUp = () => {
@@ -139,13 +131,6 @@ const handlePointerDown = async (event: PointerEvent) => {
   await tick();
 
   numberDragHead.style.transform = 'translate(0px, 0px)';
-
-  /**
-   * TODO: Determine why this is being interpreted as an `any` type by the
-   * linter when it is of `() => Promise<void>`.
-   */
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-  numberDragTooltip.recalculateStyle();
 };
 
 const handleInput = () => {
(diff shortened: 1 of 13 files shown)


packages/core/src/lib/tooltip/tooltip-styles.ts
```diff
@@ -27,48 +49,92 @@ export interface TooltipStyles {
...
+    state,
+    ($state, set) => {
+      const { target, tooltip } = $state;
+
+      return target && tooltip
+        ? autoUpdate(target, tooltip, () => updateStyle($state, set))
+        : noop;
+    },
+    INITIAL_STYLE
+  );
```
mcous, line 99: Gonna write this up in the wiki, but this was my first async derived store. It was the right tool for this particular job, but it had enough quirks that I'd be careful about using it in the future

packages/core/src/lib/input/slider-input.svelte
```diff
@@ -114,13 +113,6 @@ $: handlePointerMove = (event: PointerEvent) => {
     value = next;
     dispatch('input', value);
   }
-
-  /**
-   * TODO: Determine why this is being interpreted as an `any` type by the
-   * linter when it is of `() => Promise<void>`.
```
mcous, line 120: This appears to be a limitation of the Svelte ESLint plugin; it's a little buggy when it comes to correctly manipulating the parse tree when TS is involved (e.g. https://github.com/sveltejs/eslint-plugin-svelte/issues/583).

I've been moving type exports to separate pure TS files to work around the issue. It's a little irritating to move the interfaces farther away from the component that uses them; but it's worth it to avoid losing type safety

DTCurrie (approved):
✅

micheal-parks (approved):

mcous:
Once last fix caught by @ehhong: if `state="visible"`, the tooltip would flash in the top left of its container before moving into position. Resolved by ensuring that `invisible` is applied if the position hasn't yet been calculated. I think this _might_ have been a regression introduced by my recent work, but now its fixed and there's a test in place

PR #385, opened 2023-09-25
Remove extra flex from label

ehhong:
I think this flex is causing issues with the label positioning -- it's causing the label to be on the left even with position has value `top`

### DIFF
--- a/packages/core/src/lib/label.svelte
+++ b/packages/core/src/lib/label.svelte
@@ -50,7 +50,7 @@ export { extraClasses as cx };
   )}
 >
   <span
-    class={cx('flex text-xs', {
+    class={cx('text-xs', {
       'inline whitespace-nowrap': position === 'left',
       'text-subtle-1': !disabled,
       'cursor-not-allowed text-disabled-dark': disabled,
--- a/packages/core/src/routes/+page.svelte
+++ b/packages/core/src/routes/+page.svelte
@@ -492,6 +492,13 @@ const notify = useNotify();
       />
     </Label>
   </div>
+  <Label position="left">
+    Left
+    <Input
+      slot="input"
+      name="name"
+    />
+  </Label>
 
   <!-- Notify -->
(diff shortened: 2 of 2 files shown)


packages/core/src/routes/+page.svelte
```diff
@@ -492,6 +492,13 @@ const notify = useNotify();
       />
     </Label>
   </div>
+  <Label position="left">
+    Left
+    <Input
+      slot="input"
+      name="name"
+    />
+  </Label>
```
ehhong, line 501: just adding an extra example for reference

packages/core/src/lib/label.svelte
```diff
@@ -50,7 +50,7 @@ export { extraClasses as cx };
   )}
 >
   <span
-    class={cx('flex text-xs', {
+    class={cx('text-xs', {
       'inline whitespace-nowrap': position === 'left',
```
mcous, line 54: Oh yeah, flex was definitely a conflicting class with inline, and given the parent is flex, I don't really see why this inline is here either

packages/core/src/lib/label.svelte
```diff
@@ -50,7 +50,7 @@ export { extraClasses as cx };
   )}
 >
   <span
-    class={cx('flex text-xs', {
```
mcous, line 53: I agree with the removal, but I'm curious how this would do anything to the input slot, since it's a sibling rather than a child of this span.

I've definitely had this label render "incorrectly" horizontally on me, but every time it's been because I forgot to include `slot="input"` on my input element

mcous (approved):
I agree, I don't see why flex is on there, and it definitely conflicts with `inline` which is also sometimes applied

packages/core/src/lib/label.svelte
ehhong: oh, i didn't realize there was an input slot 🙈 that did fix my issue. i'll close this PR then?

PR #386, opened 2023-09-26
Vector input fixes

micheal-parks:
Fixes a few minor issues with vector input that emerged due to recent changes:
* Vector input cannot be set as an immutable component, because it depends on object props that in some cases must be mutable
* Vector input should fire `input` and `change` events like the underlying `SliderInput` that it wraps.


ethanlook (approved):
Nice! Could you please add `on:change` tests?

micheal-parks:
@ethanlook Definitely, done!

PR #387, opened 2023-09-27
RSDK-4395 - Update RC UI to show compass heading of robot

nfranczak:
https://viam.atlassian.net/browse/RSDK-4395
We change the robot-marker icon.
To do this, we add the following properties to the `MapMarkerLibre` class:
1. `(property) icon: HTMLElement`. With this we can define what icon we want the marker to use with an SVG url.
2. `(property) rotation?: number`. With this we can define the rotation of the marker. Here 0 is north and 90 is east.

Screenshot 2023-09-27 at 11 29 10 AM

### DIFF
--- a/packages/blocks/src/lib/maplibre/marker.svelte
+++ b/packages/blocks/src/lib/maplibre/marker.svelte
@@ -26,15 +26,22 @@ export let scale = 1;
 /** The marker color. */
 export let color = '';
 
+/** The marker icon */
+export let icon: HTMLElement | undefined;
+
+/** The marker rotation, where 0 is north */
+export let rotation = 0;
+
 const { map } = useMapLibre();
 
 let marker: Marker | undefined;
 
 $: {
   marker?.remove();
-  marker = new Marker({ scale, color });
+  marker = new Marker({ element: icon!, scale, color });
   marker.getElement().style.zIndex = '1';
   marker.setLngLat(lngLat);
+  marker.setRotation(rotation);
   marker.addTo(map);
 }
 
--- a/packages/blocks/src/lib/maplibre/marker.svelte
+++ b/packages/blocks/src/lib/maplibre/marker.svelte
@@ -27,7 +27,7 @@ export let scale = 1;
 export let color = '';
 
 /** The marker icon */
-export let icon: HTMLElement | undefined;
+export let element: HTMLElement | undefined = undefined;
 
 /** The marker rotation, where 0 is north */
 export let rotation = 0;
@@ -38,7 +38,7 @@ let marker: Marker | undefined;
 
 $: {
   marker?.remove();
-  marker = new Marker({ element: icon!, scale, color });
+  marker = new Marker(element ? { element, scale, color } : { scale, color });
   marker.getElement().style.zIndex = '1';
   marker.setLngLat(lngLat);
   marker.setRotation(rotation);
--- a/packages/blocks/src/lib/navigation-map/components/robot-marker.svelte
+++ b/packages/blocks/src/lib/navigation-map/components/robot-marker.svelte
@@ -1,14 +1,23 @@
 <script lang="ts">
-import { theme } from '@viamrobotics/prime-core/theme';
 import { MapLibreMarker, type LngLat } from '$lib';
+
+export let robotIcon = document.createElement('div');
+robotIcon.className = 'robot-marker';
+robotIcon.style.backgroundImage = `url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24' fill='green'%3E%3Cpath d='M12,2L4.5,20.29L5.21,21L12,18L18.79,21L19.5,20.29L12,2Z' /%3E%3C/svg%3E")`;
+robotIcon.style.width = `20px`;
+robotIcon.style.height = `20px`;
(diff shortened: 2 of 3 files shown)


nfranczak:
I believe before this ticket can be complete, there will be to be an accompanying PR in RDK which passes the desired rotation in.

nfranczak:
@raybjork

packages/blocks/src/lib/navigation-map/components/robot-marker.svelte
```diff
@@ -1,14 +1,23 @@
...
+robotIcon.className = 'robot-marker';
+robotIcon.style.backgroundImage = `url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24' fill='green'%3E%3Cpath d='M12,2L4.5,20.29L5.21,21L12,18L18.79,21L19.5,20.29L12,2Z' /%3E%3C/svg%3E")`;
+robotIcon.style.width = `20px`;
+robotIcon.style.height = `20px`;
+
 /** The Lng,Lat of the marker. */
 export let lngLat: LngLat | undefined = undefined;
+
+/** The rotation of the marker. */
+export let heading = 0;
```
micheal-parks, line 14:
```suggestion
export let rotation = 0;

File: packages/blocks/src/lib/maplibre/marker.svelte [hunk 2]
```diff
@@ -26,15 +26,22 @@ export let scale = 1;
 /** The marker color. */
 export let color = '';
 
+/** The marker icon */
+export let icon: HTMLElement | undefined;
```
micheal-parks, line 30:
```suggestion
export let element: HTMLElement | undefined;

File: packages/blocks/src/lib/maplibre/marker.svelte [hunk 3]
```diff
@@ -26,15 +26,22 @@ export let scale = 1;
...
+export let rotation = 0;
+
 const { map } = useMapLibre();
 
 let marker: Marker | undefined;
 
 $: {
   marker?.remove();
-  marker = new Marker({ scale, color });
+  marker = new Marker({ element: icon!, scale, color });
```
micheal-parks, line 41: To avoid a non null assertion I'd recommend:
```suggestion
  marker = new Marker(element ? { element, scale, color } : { scale, color });

File: packages/blocks/src/lib/navigation-map/components/robot-marker.svelte [hunk 4]
```diff
@@ -1,14 +1,23 @@
 <script lang="ts">
-import { theme } from '@viamrobotics/prime-core/theme';
 import { MapLibreMarker, type LngLat } from '$lib';
 
+export let robotIcon = document.createElement('div');
```
micheal-parks, line 4: Why make this a div with a background image and not just an SVG?

packages/blocks/src/lib/navigation-map/components/robot-marker.svelte
```diff
@@ -1,14 +1,23 @@
...
+robotIcon.className = 'robot-marker';
```
micheal-parks, line 5: Is this class used anywhere?

packages/blocks/src/lib/navigation-map/components/robot-marker.svelte
```diff
@@ -1,14 +1,23 @@
...
+robotIcon.style.backgroundImage = `url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24' fill='green'%3E%3Cpath d='M12,2L4.5,20.29L5.21,21L12,18L18.79,21L19.5,20.29L12,2Z' /%3E%3C/svg%3E")`;
```
micheal-parks, line 6: I'm a little worried about the lack of contrast of this icon on green backgrounds. Could we get design to produce us one that would work a bit better? Maybe something as simple as green but with a white outline would be better

packages/blocks/src/lib/navigation-map/components/waypoints.svelte
```diff
@@ -7,5 +7,6 @@ import { waypoints } from '../stores';
   <MapLibreMarker
     scale={0.7}
     lngLat={waypoint}
+    icon={undefined}
```
micheal-parks, line 10: Instead of `undefined` here we should add a default to the export in `marker.svelte`:

```ts
export let icon: HTMLElement | undefined = undefined;
```

packages/blocks/src/lib/navigation-map/components/robot-marker.svelte
nfranczak: I found this [marker.ts](https://github.com/maplibre/maplibre-gl-js/blob/2fa582c90/src/ui/marker.ts#L152) file.
From it I learned that the icon of a marker is from its element. If when a marker is created and its element is undefined the default element is same as the one used by waypoints.

My reasoning is that using a div with background image allows us to supplant the default element.

packages/blocks/src/lib/navigation-map/components/robot-marker.svelte
nfranczak: nope, I can remove

micheal-parks (approved):
Looks great!

PR #388, opened 2023-09-27
Add anatomy of a component to README

DTCurrie:
Starting to bring in things from https://github.com/viamrobotics/prime/pull/357 starting with this doc change. Feel free to leave edits as suggestions.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -80,3 +80,84 @@ pnpm -C packages/core test:watch  # watch tests
 ```
 
 [vitest]: https://vitest.dev/
+
+## Anatomy of a Component
+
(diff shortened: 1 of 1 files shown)


packages/core/README.md
```diff
@@ -80,3 +80,84 @@ pnpm -C packages/core test:watch  # watch tests
...
+  $: {
+    someString = prop ? 'whoa' : 'no';
+  }
+
+  // lifecycle hooks
+  onMount(() => ...);
+</script>
+
+<!-- Your layout -->
+<div class="border-black"><slot /></div>
```
ethanlook, line 158: Maybe something about using default slot when there's only one slot, but named slots when there are multiple?

ethanlook (approved):
💪🏼

packages/core/README.md
DTCurrie: Good callout

micheal-parks (approved):
🚀

PR #390, opened 2023-09-28
Add h-7.5 and w-7.5 classes

DTCurrie:
Continuing with changes from https://github.com/viamrobotics/prime/pull/357 this replaces the usage of `h-[30px]` and `w-[30px]` arbitrary classes with new utility classes.

### DIFF
--- a/packages/core/theme.ts
+++ b/packages/core/theme.ts
@@ -70,8 +70,14 @@ export const theme = {
       cyberpunk: '#a51aff',
       hyperlink: '#0000ea',
     },
+    height: {
+      '7.5': '1.875rem',
+    },
     transitionProperty: {
       'expand-vertical': 'max-height,visibility',
+    },
+    width: {
+      '7.5': '1.875rem',
     },
     zIndex: {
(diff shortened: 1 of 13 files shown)


packages/core/theme.ts
```diff
@@ -70,9 +70,15 @@ export const theme = {
       cyberpunk: '#a51aff',
       hyperlink: '#0000ea',
     },
+    height: {
```
mcous, line 73: Should we be extending [`spacing`](https://tailwindcss.com/docs/customizing-spacing#extending-the-default-spacing-scale) rather than height and width directly?

mcous (changes requested):
LGTM but I think we want to extend `spacing`, not `height` and `width`

packages/core/theme.ts
DTCurrie: Ok yes thank you, I though the extend was “sizing” for some reason and couldn’t find it. Thought it was weird I had to declare both.

mcous (approved):

PR #391, opened 2023-09-29
Fix modal

anjinai:


packages/core/src/lib/__tests__/modal.spec.ts
```diff
@@ -24,15 +24,41 @@ describe('Modal', () => {
     const onClose = vi.fn();
     component.$on('close', onClose);
 
-    const background = container.querySelector('div[role="button"]');
-    if (!background) {
-      throw new Error('Background not found');
+    const modal = container.querySelector('.relative.max-w-lg');
+    if (!modal) {
+      return;
     }
```
mcous, line 30: If your tests couldn't find an element for the modal, the test should fail. In this case, it'd be fine to move right on the the `click`. This is a side-effect of using `querySelector` in a test instead of `screen.getByRole` or similar, which _will_ error if the element can't be found

Things you should almost never do in unit tests, in descending order of "don't do this":

- Early returns
- Throwing your own errors
- `if` statements
- `for` loops

packages/core/src/lib/modal.svelte
```diff
@@ -26,31 +26,41 @@ Creates a modal overlay.
...
 /** Whether the modal is open. */
 export let open = false;
 /** The variant of the modal. */
 export let variant: 'small' | '' = '';
 
+let clickOutsideActive = false;
+
 const dispatch = createEventDispatcher<{
   /** Fires when the modal is requested to close. */
   close: boolean;
```
mcous, line 42: Does this event really need a payload?

packages/core/src/lib/modal.svelte
```diff
@@ -26,31 +26,41 @@ Creates a modal overlay.
...
+    return;
   }
+  dispatch('close', true);
 };
 
-const handleBackgroundKey = (event: KeyboardEvent) => {
-  if (event.key === 'Enter' || event.key === 'Space') {
+const handleEscapePress = (event: KeyboardEvent) => {
+  if (event.key === 'Escape') {
     handleCloseModal();
```
mcous, line 64: I would add an `event.preventDefault()` for good measure

packages/core/src/lib/modal.svelte
```diff
@@ -60,19 +70,21 @@ onMount(() => {
...
     {
       invisible: !open,
     }
   )}
-  role="button"
-  tabindex="0"
-  on:click={handleBackgroundClick}
-  on:keydown={handleBackgroundKey}
+  role="dialog"
+  tabindex="-1"
```
mcous, line 83: I'm not 100% sure this `tabindex` is necessary; a `div` shouldn't be in the document's tab list to begin with. Will double check MDN on this, though, since maybe it has an effect on children?

packages/core/src/lib/modal.svelte
```diff
@@ -60,19 +70,21 @@ onMount(() => {
 });
 </script>
 
+<svelte:window on:keydown={handleEscapePress} />
+
 <div
   class={cx(
     'fixed left-0 top-0 z-50 flex h-full w-full items-center justify-center bg-black bg-opacity-40',
     {
       invisible: !open,
```
mcous, line 79: I know this is existing code, but IMO a closed modal should simply not be rendered. Can this `invisible` be swapped with an `#if`?

```svelte
{#if open}
  <svelte:window on:keydown={...} />
  <!-- ... -->
{/if}
```

As I'm writing this out, I bet there's a decent change the class="invisible" (as opposed to "do not render" is why `use:clickOutside` isn't working

packages/core/src/lib/modal.svelte
```diff
@@ -26,31 +26,41 @@ Creates a modal overlay.
...
+});
 
-const handleBackgroundClick = (event: MouseEvent) => {
-  if (event.target === event.currentTarget) {
-    handleCloseModal();
+const handleCloseModal = () => {
+  if (!clickOutsideActive) {
+    return;
   }
+  dispatch('close', true);
```
mcous, line 59: Should this check `open` before firing the event?

mcous (commented):
Initial thoughts! Gonna test this out locally to see about the click outside issue

packages/core/src/lib/modal.svelte
```diff
@@ -60,19 +70,21 @@ onMount(() => {
 });
```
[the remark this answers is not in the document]
mcous: Update: we may need to guard against SSR on this one by verifying that `document` is defined before doing anything:

```ts
$: if (typeof document !== 'undefined') {
  document.body.classList.toggle('overflow-hidden', open);
}
```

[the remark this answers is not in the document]
anjinai: I get `ReferenceError: document is not defined` when I try to move this out of onMount and change it to a reactive statement.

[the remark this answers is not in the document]
anjinai: Ohhhh, that's why I was getting a reference error.  Thank you!

packages/core/src/lib/modal.svelte
```diff
@@ -26,31 +26,41 @@ Creates a modal overlay.
...
+
 const dispatch = createEventDispatcher<{
   /** Fires when the modal is requested to close. */
   close: boolean;
 }>();
 
-const handleCloseModal = () => {
-  dispatch('close', true);
-};
+afterUpdate(() => {
```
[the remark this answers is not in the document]
anjinai: That totally worked, thank you!!  

The only issue I have is that I'm failing one of the clickOutside edge case tests for if a clicked event gets removed from the DOM:

```
  it('should not trigger if clicked element gets removed from the DOM', async () => {
    render(Subject, { onClickOutside });

    const user = userEvent.setup();
    const insideButton = screen.getByTestId('inside');

    insideButton.addEventListener('click', () => {
      insideButton.remove();
    });

    await user.click(insideButton);

    expect(onClickOutside).not.toHaveBeenCalled();
  });
```

ehhong:
could you also add a max-height to the modal? i'm coming across a situation where if the content overflows the modal it looks like this (instead of having a vertical scroll)
image

packages/core/src/lib/modal.svelte
```diff
@@ -26,81 +26,84 @@ Creates a modal overlay.
...
+  document.body.classList.toggle('overflow-hidden', open);
+}
+
 /** The variant of the modal. */
 export let variant: 'small' | '' = '';
 
 const dispatch = createEventDispatcher<{
   /** Fires when the modal is requested to close. */
-  close: boolean;
+  close: void;
```
mcous, line 45: `void` is only a valid type annotation for function returns. I think this should be `undefined` or `never`. If TS complains about either of those, then I think `close: true` and not worrying about it any more is also fine

packages/core/src/lib/modal.svelte
```diff
@@ -26,81 +26,84 @@ Creates a modal overlay.
...
-  dispatch('close', true);
-};
-
-const handleBackgroundClick = (event: MouseEvent) => {
-  if (event.target === event.currentTarget) {
-    handleCloseModal();
+  if (!open) {
+    return;
   }
+  dispatch('close');
```
mcous, line 52: [Nit] Generally I'm a fan of early returns in complex chains, but given we've got a single condition and a single action here, I think this is easier to read if we flip the condition:

```suggestion
  if (!open) {
    dispatch('close');
  }
  ```

File: packages/core/src/lib/__tests__/modal.spec.ts [hunk 11]
```diff
@@ -1,38 +1,59 @@
...
-    await fireEvent.click(background);
+    await fireEvent.click(modal.parentElement!);
 
-    expect(onClose).toHaveBeenLastCalledWith(
-      expect.objectContaining({ detail: true })
-    );
+    expect(onClose).toBeCalled();
+  });
+
+  it('if open is true, modal should be visible', () => {
```
mcous, line 34: Can we add an expectation to this test to check that `aria-modal` is set properly?

packages/core/src/lib/__tests__/modal.spec.ts
```diff
@@ -1,38 +1,59 @@
...
 
 describe('Modal', () => {
   it('should emit close event when close icon button is clicked', async () => {
-    const { getByTitle, component } = render(Modal, {
+    const { component } = render(Modal, {
       open: true,
     });
 
-    const closeButton = getByTitle('Close modal');
+    const closeButton = screen.getByTitle('Close modal');
```
mcous, line 11: [Not new code, improvement opportunity] `getByTitle` is not particularly accessible. It think this should be:

```suggestion
    const modal = screen.getByRole('dialog')
    const closeButton = within(modal).getByRole('button', { name: /close/iu });
```

You'll need to `import { ..., within } from '@testing-library/svelte'` for this to work

packages/core/src/lib/modal.svelte
```diff
@@ -26,81 +26,84 @@ Creates a modal overlay.
...
+
+{#if open}
   <div
-    class={cx('relative max-w-lg border border-medium bg-white p-6 shadow-sm', {
-      'max-h-[600px] min-h-[400px] w-[480px]': variant === '',
-      'max-h-[320px] w-[400px]': variant === 'small',
-    })}
+    class="fixed left-0 top-0 z-50 flex h-full w-full items-center justify-center bg-black bg-opacity-40"
+    role="dialog"
+    tabindex="-1"
```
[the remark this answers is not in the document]
ehhong: oops, this may have been my fault, i was reading the guidelines here https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

do we still need to provide tabindex to prevent tabbing to the modal if it's not open? or maybe the aria role is enough to accommodate that?

packages/core/src/lib/modal.svelte
ehhong: i was able to do `null` here, not sure if that is any better though

packages/core/src/lib/modal.svelte
anjinai: Thanks for the suggestion!  Updated to accept an isOpen writable store

packages/core/src/lib/modal.svelte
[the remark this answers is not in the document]
anjinai: Thank you!  Updated accordingly, my test for this isn't cooperating, do you happen to have any suggestions?

```
  it('should focus on heading element on mount', () => {
    render(Modal, { isOpen });
    const heading = document.querySelector('#modal-heading');
    const elementInFocus = document.activeElement;
    expect(elementInFocus).toEqual(heading);
  });
```

packages/core/src/lib/modal.svelte
```diff
@@ -26,81 +26,91 @@ Creates a modal overlay.
...
 import cx from 'classnames';
-import { createEventDispatcher, onMount } from 'svelte';
 import IconButton from './button/icon-button.svelte';
+import { clickOutside } from '$lib';
+import { writable } from 'svelte/store';
+import { onMount } from 'svelte';
 
 /** Whether the modal is open. */
-export let open = false;
+export let isOpen = writable(false);
```
mcous, line 35: I don't think this should have a default value, because the component really isn't usable if you don't pass something in

```suggestion
export let isOpen: Writable<boolean>;
```

packages/core/src/lib/__tests__/modal.spec.ts
```diff
@@ -1,38 +1,71 @@
...
+  });
 
-    expect(onClose).toHaveBeenLastCalledWith(
-      expect.objectContaining({ detail: true })
-    );
+  it('should focus on heading element on mount', () => {
+    render(Modal, { isOpen });
+    const heading = document.querySelector('#modal-heading');
+    const elementInFocus = document.activeElement;
+    expect(elementInFocus).toEqual(heading);
```
mcous, line 69: - Don't use `querySelector`, use `getByRole`
- Unless it's semantically important (e.g. it's the ID of a label element that is passed to `aria-labeledby`) IDs should not be used as test hooks
- Use [`expect(...).toHaveFocus()`](https://github.com/testing-library/jest-dom#tohavefocus) to check focus

```suggestion
    render(Modal, { isOpen });
    
    const modal = screen.getByRole('dialog');
    const heading = within(modal).getByRole('heading')

    expect(heading).toHaveFocus()
```

packages/core/src/lib/modal.svelte
```diff
@@ -26,81 +26,91 @@ Creates a modal overlay.
...
+        class="absolute right-4 top-4 text-gray-6"
+        icon="close"
+        label="Close modal"
+        on:click={handleCloseModal}
+      />
+      <div
+        class={cx('flex flex-col gap-2', { 'min-h-[400px]': variant === '' })}
+      >
+        <h2
+          id="modal-heading"
```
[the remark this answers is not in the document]
anjinai: Thank you, removed id from that element

packages/core/src/lib/__tests__/modal.spec.ts
```diff
@@ -1,38 +1,71 @@
...
 describe('Modal', () => {
-  it('should emit close event when close icon button is clicked', async () => {
-    const { getByTitle, component } = render(Modal, {
-      open: true,
+  let isOpen = writable(true);
+  let currentValue = true;
+
+  beforeEach(() => {
+    isOpen = writable(true);
+    currentValue = true;
```
mcous, line 13: You can use `get` from `svelte/store` to access the current store value

```ts
import { get, writable } from 'svelte/store';

// ...

isOpen = writable(true);
console.log(get(isOpen)); // logs `true`
```

packages/core/src/lib/modal.svelte
```diff
@@ -26,81 +26,90 @@ Creates a modal overlay.
...
+const handleEscapePress = (event: KeyboardEvent) => {
+  if (event.key === 'Escape') {
+    event.preventDefault();
     handleCloseModal();
   }
 };
 
 onMount(() => {
-  document.body.classList.toggle('overflow-hidden', open);
+  if ($isOpen) {
```
mcous, line 58: - Attaching this login to `onMount` won't get us the behavior we want; we need it to run whenever the modal opens, which will likely occur more than once sometime after mount
- There's very rarely a need to use `querySelector` in Svelte, instead we can [`bind`]()  the `<h2>`'s DOM element to a `let` variable: https://svelte.dev/docs/element-directives#bind-this

We can use a reactive statement to achieve this behavior:

```ts
let headingElement: HTMLElement | undefined

$: headingElement?.focus()
```

The `$:` will cause the code that follows - `headingElement?.focus()` - to run whenever the `let headingElement` variable changes. The variable will change value in reaction to two events:

1. The modal appears and the `<h2>` element we're bound to mounts, so `headingElement` is set to the `HTMLElement`
2. The modal disappears and the `<h2>` unmounts, so `headingElement` is set to `undefined`

In (1), our statement will trigger focus. In (2), the statement will short-circuit.

Be sure to test this out in a browser! There's a chance this will call `.focus()` too early, and the element won't actually focus. If that's the case, then we may need to tweak this ever so slightly:

```ts
$: if (headingElement) {
  requestAnimationFrame(() => headingElement?.focus())
}
```

packages/core/src/routes/+page.svelte
```diff
@@ -526,7 +528,7 @@ const notify = useNotify();
   <div>
     <Button on:click={handleOpenModal}>Open Modal</Button>
     <Modal
-      open={modalOpen}
+      isOpen={modalOpen}
       on:close={handleCloseModal}
```
mcous, line 532: Looks like this `on:close` can be removed

packages/core/src/lib/modal.svelte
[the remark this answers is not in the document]
mcous: Not quite, there's almost never any need to use `querySelector` in Svelte. Will add a comment on the latest diff so it doesn't get buried

mcous (commented):
Almost there! I think the simplification of the component is really good, just need a few more tweaks to get the heading stuff sorted

packages/core/src/routes/+page.svelte
anjinai: whoops, thanks!

packages/core/src/lib/__tests__/modal.spec.ts
anjinai: That's so much simpler, tyty!

packages/core/src/lib/modal.svelte
anjinai: Ahhh, I see - thanks so much for the explanation!

packages/core/src/lib/__tests__/modal.spec.ts
```diff
@@ -1,38 +1,61 @@
...
+  });
+
+  it('should close modal when close icon button is clicked', async () => {
+    render(Modal, { isOpen });
+
+    const modal = screen.getByRole('dialog');
+    const closeButton = within(modal).getByRole('button', { name: /close/iu });
 
-    expect(onClose).toBeCalledWith(expect.objectContaining({ detail: true }));
+    await userEvent.click(closeButton);
```
mcous, line 20: Rather than using `userEvent` directly, you should create a user object from `userEvent.setup()`, as [recommended in the docs](https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent)

packages/core/src/lib/__tests__/modal.spec.ts
```diff
@@ -1,38 +1,61 @@
-import { describe, it, expect, vi } from 'vitest';
-import { render, fireEvent } from '@testing-library/svelte';
+import { describe, it, expect, beforeEach } from 'vitest';
+import { render, fireEvent, screen, within } from '@testing-library/svelte';
```
mcous, line 2: Should `fireEvent` be removed?

packages/core/src/lib/__tests__/modal.spec.ts
```diff
@@ -1,38 +1,61 @@
...
+    isOpen.set(false);
+    render(Modal, { isOpen });
+    const modal = screen.queryByRole('dialog');
+    expect(modal).not.toBeInTheDocument();
+  });
+
+  it('should close modal when escape key is pressed', async () => {
+    render(Modal, { isOpen });
+
+    await fireEvent.keyDown(window, { key: 'Escape' });
```
mcous, line 51: Prefer `await user.keyboard('{Escape}')`

packages/core/src/lib/modal.svelte
```diff
@@ -26,81 +26,89 @@ Creates a modal overlay.
...
+  document.body.classList.toggle('overflow-hidden', $isOpen);
+}
+
+let headingElement: HTMLElement | undefined;
+
+$: headingElement?.focus();
+
+$: if (headingElement) {
+  requestAnimationFrame(() => headingElement?.focus());
+}
```
mcous, line 46: You should only need one of these, not both

packages/core/src/lib/__tests__/modal.spec.ts
anjinai: Sorry missed removing this fireEvent.  Updated!

PR #392, opened 2023-10-03
APP-2292 - ListBox core migration

Kschappacher:
Migrating List box to Prime Core, updated the svelte files to have native dispatcher and booleans as well as updated story books, tests etc...

### DIFF
--- a/packages/core/src/lib/list-box.svelte
+++ b/packages/core/src/lib/list-box.svelte
@@ -28,7 +28,7 @@ import { Icon } from '$lib';
 const dispatch = createEventDispatcher();
 
 // Toggle element on or off
-export let disabled: boolean;
+export let disabled = false;
 // Elements in the left list
 export let left = '';
 // Elements in the right list
@@ -39,8 +39,8 @@ export let leftlabel = '';
 export let rightlabel = '';
 // Height of the two boxes
 export let height = '200px';
-// ?????
-export let suffix: boolean;
+// Enable suffix that can be put in front of elements in the box
+export let suffix = false;
 
 type ListBoxSide = 'left' | 'right';
 
--- a/packages/core/src/lib/list-box.svelte
+++ b/packages/core/src/lib/list-box.svelte
@@ -1,0 +1,249 @@
+<!--
+    @component
+    Two lists of elements that can be moved between the two lists
+
+    ```svelete
+     <ListBox
+        disabled = {false}
+        left ='leftText1,leftText2,leftText3',
+        right = 'rightText1,rightText2,rightText3'
+        leftlabel = 'Left'
+        rightlabel = 'Right'
+        height = '200px'
+        suffix= {false}
+        />
+    ```
+    
+-->
+<svelte:options immutable />
+
+<script lang="ts">
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+// disabled to pass untyped object to cx library in template
+
+import cx from 'classnames';
+import { afterUpdate, createEventDispatcher } from 'svelte';
+import { Icon } from '$lib';
+
+const dispatch = createEventDispatcher();
+
+// Toggle element on or off
+export let disabled: boolean;
+// Elements in the left list
+export let left = '';
+// Elements in the right list
+export let right = '';
+// Label for the left list
+export let leftlabel = '';
+// Label for the right list
+export let rightlabel = '';
+// Height of the two boxes
+export let height = '200px';
+// ?????
+export let suffix: boolean;
+
+type ListBoxSide = 'left' | 'right';
+
+interface ListBoxOption {
+  value: string;
+  selected: boolean;
+  suffix?: string;
+}
+
+interface Options {
+  left: ListBoxOption[];
+  right: ListBoxOption[];
+}
+
+const LEFT = 'left';
+const RIGHT = 'right';
+
+const generateOption = (initial: string): ListBoxOption => {
+  if (suffix) {
+    const split = initial.split(' ');
+    return { value: split[0] ?? '', suffix: split[1] ?? '', selected: false };
+  }
+
+  return { value: initial, selected: false };
+};
+
+let options = {
+  left: left ? left.split(',').map((val) => generateOption(val)) : [],
+  right: right ? right.split(',').map((val) => generateOption(val)) : [],
+};
+
+// when new values are added, we want to update the state of options to include those values
+const addNewData = () => {
+  const allValues = new Set([
+    ...options.left.map((opt) => opt.value),
+    ...options.right.map((opt) => opt.value),
+  ]);
+  const newLeftOptions = left
+    ? left
+        .split(',')
+        .map((val) => generateOption(val))
+        .filter((opt) => !allValues.has(opt.value))
+    : [];
+  const newRightOptions = right
+    ? right
+        .split(',')
+        .map((val) => generateOption(val))
+        .filter((opt) => !allValues.has(opt.value))
+    : [];
+
+  const newOptions = {
+    left: [...options.left, ...newLeftOptions],
+    right: [...options.right, ...newRightOptions],
+  };
+
+  options = newOptions;
+};
+
+afterUpdate(addNewData);
+
+const handleOptionClick = (option: ListBoxOption, side: ListBoxSide) => {
+  if (disabled) {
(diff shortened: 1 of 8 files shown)


anjinai:
This is still in draft mode, just wanted to check that this is good to review?

packages/core/src/lib/list-box/__tests__/list-box.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+  rightlabel: 'Right',
+  height: '200px',
+  suffix: false,
+};
+
+describe('ListBox', () => {
+  it('Renders left options and can move them to the right', async () => {
+    render(ListBoxSlot, { ...defaultProps });
+
+    await fireEvent.click(screen.getByText('leftText1'));
```
anjinai, line 20: Cousins recently gave me some feedback in one of my other prime core migration PRs saying that it's better to use user event instead of fireEvent based on the testing library documentation.  

https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent

Kschappacher:
> This is still in draft mode, just wanted to check that this is good to review?

Yupp, good to review!

packages/core/src/lib/list-box/__tests__/list-box.spec.ts
Kschappacher: 👍 swapped it all over

anjinai (approved):
LGTM

packages/core/src/lib/list-box.svelte
mcous, line 1: Should this have been moved into the `list-box` directory?

packages/core/src/lib/list-box.svelte
```diff
@@ -0,0 +1,249 @@
...
+        height = '200px'
+        suffix= {false}
+        />
+    ```
+    
+-->
+<svelte:options immutable />
+
+<script lang="ts">
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
mcous, line 21: To my knowledge, this isn't required elsewhere, why is it required here? Also, prefer per-line ignores to per-file ignores

packages/core/src/lib/list-box.svelte
```diff
@@ -0,0 +1,249 @@
...
+        name="arrow-up"
+      />
+    </button>
+    <button
+      class={cx(
+        'flex h-8 w-8 -rotate-90 cursor-pointer items-center justify-center border border-black bg-white hover:scale-105',
+        { 'border-black/50': disabled }
+      )}
+      on:click={() => handleMoveClick(LEFT)}
+      data-testid="move-left"
```
mcous, line 207: Why is this `testid` needed? Is there a way to avoid it with a different component structure or better queries?

packages/core/src/lib/list-box.svelte
```diff
@@ -0,0 +1,249 @@
...
+import cx from 'classnames';
+import { afterUpdate, createEventDispatcher } from 'svelte';
+import { Icon } from '$lib';
+
+const dispatch = createEventDispatcher();
+
+// Toggle element on or off
+export let disabled = false;
+// Elements in the left list
+export let left = '';
```
mcous, line 33: These props should definitely be updated to take arrays, not string

packages/core/src/lib/list-box.svelte
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  options[source] = newOptions[source];
+  options[target] = [...options[target], ...newOptions[target]];
+  options = { ...options };
+
+  // copy object to avoid any potential pass-by-reference issues
+  dispatch('move', {
+    options: JSON.parse(JSON.stringify(options)) as Options,
+  });
```
mcous, line 144: All this looks... a little sus. `options` is internal to this component, what's all this spreading / JSON (de)serialization doing?

packages/core/src/lib/list-box/__tests__/list-box.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  height: '200px',
+  suffix: false,
+};
+
+describe('ListBox', () => {
+  it('Renders left options and can move them to the right', async () => {
+    const user = userEvent.setup();
+    render(ListBoxSlot, { ...defaultProps });
+
+    await user.click(screen.getByText('leftText1'));
```
mcous, line 22: When we're clicking elements in tests, we should always use `getByRole('button', ...)` (or whatever similar role). `getByText` should be your third or fourth choice when you're trying to pick an element query in a test

packages/core/src/lib/list-box/__tests__/list-box.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    await user.click(screen.getByText('leftText2'));
+    await user.click(screen.getByText('leftText3'));
+
+    const moveRightButton = screen.getByTestId('move-right');
+    await user.click(moveRightButton);
+
+    expect(screen.getByText('leftText1')).toBeVisible();
```
mcous, line 29: When possible, put elements in a variable so you're only calling `getByWhatever` once. The only time to re-call `getBy` for the same element is if it's getting mounted / unmounted during the course of the test

mcous (commented):
@Kschappacher I see this is already merged but I don't think it was quite ready. It looks like we preserved the weird string-based props from prime legacy, and we need to not do that. Can we re-open the ticket and take another pass at this?

(Alternatively, it looks like this ticket is "lowest" priority, which probably means there's no immediate need for this component. If so, there's not an actual need to take another pass at it right now as long as the ticket stays open)

Kschappacher:
> @Kschappacher I see this is already merged but I don't think it was quite ready. It looks like we preserved the weird string-based props from prime legacy, and we need to not do that. Can we re-open the ticket and take another pass at this?
> 
> (Alternatively, it looks like this ticket is "lowest" priority, which probably means there's no immediate need for this component. If so, there's not an actual need to take another pass at it right now as long as the ticket stays open)

Ill open up the branch again and take a look at all the new comments!

packages/core/src/lib/list-box/__tests__/list-box.spec.ts
Kschappacher: Im not a 100% sure if this is how it works but screen.get returns an html element. I was un sure if that element is a ref to the value on the screen or if it is just a copy of the html element. So to be safe I called it before the elements get moved around and after to be safe. Do you know which case is true? ref to the element or copy of the element?

packages/core/src/lib/list-box.svelte
Kschappacher: Ill remove and try and rework the tests

packages/core/src/lib/list-box.svelte
Kschappacher: I didnt write the original version of this component but from what I understand json deserialization is a way easy way to make a deep copy of the object.

mcous: My question was more along the lines of "why are we doing this cloning?" Looking at this block:

- L137: mutates the `options` object
- L128: mutates the `options` object again
- L129: shallow-clones the `options` object
- L143: deep-clones the `options object and places it in an event

So the reason I say it's "sus" is that we're inconsistent with mutations and cloning, and it's an internal object. It all feels like a design problem with this component's API (props and events)

packages/core/src/lib/list-box/__tests__/list-box.spec.ts
mcous: In `testing-library`, `getByXyz` returns the actual DOM element. If the implementation of the component creates new elements rather than moving them, then yes, you will need to call it again. In that case, I would still recommend assigning to a variable for readability and to communicate the intent of your tests

```ts
const goodNameForThisElement = screen.getByXyz(...)

expect(goodNameForThisElement).toBeVisible()
```

(Playwright is different, in that its methods return `Locators`, which are not evaluated until you call `expect`. I kind of prefer this approach! But testing-library's approach is certainly more simple)

packages/core/src/lib/list-box.svelte
Kschappacher: Looks like both of the move buttons are almost identical. I tried some stuff to screen get them but was unsuccessful. If you have any ideas im down to try them.

PR #394, opened 2023-10-04
Refactor switch

DTCurrie:
Another one from playing with prime-core. This removes the `label` from the switch markup so it doesn't conflict with the label styles and so it can be used with the label features. It also updates the annotations to be editable with slots, and removes the redundant `readonly` prop. Also removes the `tooltip` prop since a `Tooltip` can be used inside the label.

Putting this up after a first pass to get some opinions on it, but using this felt closer to other prime-core components when I was using them in RC. That was the main reason for the change.


ethanlook (approved):
🚀

packages/core/src/lib/__tests__/switch.spec.ts
```diff
@@ -48,12 +34,22 @@ describe('Switch', () => {
   });
 
   it('Renders annotated switch', async () => {
-    render(Switch, { variant: 'annotated' });
+    render(Switch, { annotated: true });
+
+    expect(screen.getByText('Off')).toBeVisible();
+
+    await fireEvent.click(screen.getByRole('switch'));
```
mcous, line 41: [Nit, improvement opportunity]: can we swap userEvent in here?

packages/core/src/lib/switch.svelte
```diff
@@ -1,132 +1,101 @@
...
+      bind:checked={on}
+    />
+  </div>
 
-    {#if variant === 'annotated'}
-      <p class="text-sm">{on ? 'on' : 'off'}</p>
-    {/if}
-  </button>
-</label>
+  {#if annotated}
```
mcous, line 92: [Nit]: This `annotated` feels unnecessary if the caller explicitly passes `on` or `off`, can we throw a slot check in here?

```suggestion
  {#if annotated || $$slots.on || $$slots.off }
```

mcous (approved):

PR #395, opened 2023-10-04
APP-2839: adjust context menu for floating usage

mcous:
This PR makes some changes to the `<ContextMenu>` and friends that are necessary to get a well-behaved floating menu.

Menu:
- Add an `id` prop to the menu, so the button that opens it can set `aria-controls`
- Switch to `width: max-content` rather than `xs` (let me know if you think this should simply be removed)

Items:
- Remove the `select` event in favor of simply forwarding `on:click`
- Remove the `label` prop in favor of simply taking a `<slot />`

Our keyboard navigation of the context menu is "wrong" according to [WAI](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/), but it's a bit of a can of worms, so this PR does not make any attempt to address it.

Please check out the test and storybook changes and keep an eye out for an accompanying app PR


packages/core/src/lib/context-menu/context-menu.svelte
```diff
@@ -14,15 +14,18 @@ A simple list menu with items, usually appears on click.
...
 let extraClasses: cx.Argument = '';
 export { extraClasses as cx };
 </script>
 
 <div
+  {id}
   role="menu"
   class={cx(
-    'max-w-xs border border-medium py-1 shadow-sm filter-none',
+    'border border-medium bg-white py-1 shadow-sm filter-none',
```
ehhong, line 28: thanks for adding this!

ehhong (approved):
these changes look good to me! i thought we were going to add the floating menu to `prime-core` as well? see discussion here https://viaminc.slack.com/archives/C03LCAHM2TT/p1695915470656869

ethanlook (approved):
🚀

ethanlook (approved):

mcous:
> i thought we were going to add the floating menu to prime-core as well?

Oh hm, I misremembered. I will move the app PR into a new Prime PR, accordingly!

PR #399, opened 2023-10-06
APP-2839: add simple floating context menu component

mcous:
Check out the Playground to see the component in action. I don't necessarily _love_ the API, but I think this is about as far as I can take it without actually seeing it in action.

It also doesn't feel like `floating-ui` is super necessary at this stage, but it was how it was implemented in Vue so 🤷


ehhong:
code looks good to me! from manually testing in the playground:
- for the floating menu example, the context menu should close after clicking "Close menu" (the control slot)
- [q] would there be any way to add minor changes to the floating position with this interface? for example, if i want the menu to float just a bit more below the control slot, can i include additive style changes?

packages/core/src/lib/context-menu/__tests__/floating-menu.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+import { writable } from 'svelte/store';
+
+import Subject from './floating-menu.spec.svelte';
+
+describe('<FloatingMenu> component', () => {
+  let user: ReturnType<typeof userEvent.setup>;
+
+  beforeEach(() => {
+    user = userEvent.setup();
+  });
```
ethanlook, line 13: Nice, we should broadcast this

ethanlook (approved):
Do you mind adding example usage to the storybook as well? Thanks!

ehhong:
changes look great, thanks so much!! last thing -- should the context menu close when an item is clicked? or can/should that be controlled by the parent element?

mcous:
@ehhong good catch, I exposed a quick `closeMenu` slot prop so that the parent can wire up a close action to each item's `on:click` handler. I don't think this is a great API, but it'll allow us to wire it up so we can see it all in action before coming up with something better

ehhong (approved):
lgtm!! thanks again for taking this on 🎊

PR #403, opened 2023-10-10
Update radios to look/behave like radios

DTCurrie:
Updating the radios to match the designs here: https://www.figma.com/file/ERHSYDGTXiEQOVGY8ORg6I/PRIME?type=design&node-id=2602-5221&mode=design&t=NDUZKAwepkYxgxVH-0 and just a lot of cleanup and simplification.

Important callouts are now instead of using a row of buttons, we are now using a `fieldset` with radio `input`s, and leveraging the svelte `bind:group` to manage value selection. 

**UPDATE:** I have refactored the original radios into the `ToggleButtons` and cleaned those up a bit. I am open to suggestions on the component name.

Screenshot 2023-10-10 at 12 05 56 PM

Screenshot 2023-10-10 at 1 33 34 PM


packages/core/src/lib/icon/icons.ts
```diff
@@ -76,3 +79,10 @@ export const paths = {
  * The possible icon names that can be rendered. This is good for typing props.
  */
 export type IconName = keyof typeof paths;
+
+/** 
+ * A map of icon names, useful for dynamically adding icons with type safety.
+ */
+export const iconNames = Object.fromEntries<IconName>(
+  Array<IconName>.from(Object.keys(paths)).map((name) => [name, name as IconName])
+) as Record<IconName, IconName>
```
DTCurrie, line 88: Open to suggestions on a better way of easily using icon names dynamically, this came up in the radio and I know we do `icon` selection elsewhere. Usually that looks something like:

```svelte
let icon: IconName:
$: {
  if (thing) {
    icon = 'some-magic-string';
  } elseif {
    icon = 'another-magic-string';
  } else {
    icon = 'yet-another-magic-string';
  }
}
```

Don't love having all the magic strings, and while that works for singular icons it gets messier with determining icons for multiple elements like in the radio. This at least gives us a quick map of names we can refer to, and should make refactoring icon names slightly less painful in the future. If we like this (or something else that accomplishes the same thing), I can refactor other places where we declare icon names to use it as well for consistency.

micheal-parks:
Are we getting rid of the current `<Radio>`? Seems like it would be a good idea to just move it to `<Options>` or `<ButtonGroup>` (I like [Tailwind UI's naming](https://tailwindui.com/components/application-ui/elements/button-groups))

DTCurrie:
> This is a great change, but are we getting rid of the current `<Radio>`? Seems like it would be a good idea to just move it to `<Options>` or `<ButtonGroup>` (I like [Tailwind UI's naming](https://tailwindui.com/components/application-ui/elements/button-groups) but not sure if the current component is really a button group)

I was actually writing up a comment about this, but yeah that is a shorter version of it. I think our radios should be radios, but I also could see the old component still being used. `RadioButtons` could work, I was also thinking `SwitchButtons` but am open to suggestions!

packages/core/src/lib/icon/icons.ts
```diff
@@ -76,3 +79,12 @@ export const paths = {
...
 export type IconName = keyof typeof paths;
+
+const ICON_NAMES = Object.keys(paths) as IconName[];
+
+/**
+ * A map of icon names, useful for dynamically adding icons with type safety.
+ */
+export const iconNames = Object.fromEntries<IconName>(
+  ICON_NAMES.map((name) => [name, name])
+) as Record<IconName, IconName>;
```
DTCurrie, line 90: Open to suggestions on a better way of easily using icon names dynamically, this came up in the radio and I know we do icon selection elsewhere. Usually that looks something like:

let icon: IconName:
$: {
  if (thing) {
    icon = 'some-magic-string';
  } elseif {
    icon = 'another-magic-string';
  } else {
    icon = 'yet-another-magic-string';
  }
}
Don't love having all the magic strings, and while that works for singular icons it gets messier with determining icons for multiple elements like in the radio. This at least gives us a quick map of names we can refer to, and should make refactoring icon names slightly less painful in the future. If we like this (or something else that accomplishes the same thing), I can refactor other places where we declare icon names to use it as well for consistency.

packages/core/src/routes/+page.svelte
```diff
@@ -38,6 +38,7 @@ import {
   useNotify,
   Modal,
 } from '$lib';
+import ToggleButtons from '$lib/toggle-buttons.svelte';
```
ethanlook, line 41: nit: add to above import

ethanlook (approved):
Reminder to bump version of core and blocks!

packages/core/src/lib/icon/icons.ts
mcous: > If we like this (or something else that accomplishes the same thing)

I'm unclear on what exactly this export is accomplishing, based on usage in this PR. Would you mind explaining a little more?

```ts
$: getIcon = (option: string): IconName =>
  isSelected(option)
    ? iconNames['radiobox-marked']
    : iconNames['radiobox-blank'];
```

doesn't seem very different, magic-strings-wise, than:

```ts
$: getIcon = (option: string): IconName =>
  isSelected(option)
    ? 'radiobox-marked'
    : 'radiobox-blank';
```

packages/core/src/lib/icon/icons.ts
```diff
@@ -76,3 +79,12 @@ export const paths = {
  * The possible icon names that can be rendered. This is good for typing props.
  */
 export type IconName = keyof typeof paths;
+
+const ICON_NAMES = Object.keys(paths) as IconName[];
```
mcous, line 83: Why is this casting needed? Is `paths` missing an `as const`?

packages/core/src/lib/__tests__/radio.spec.ts
```diff
@@ -1,142 +1,108 @@
...
   });
 
   it('Marks the selected option correctly', () => {
-    render(Radio, { options: ['Option1', 'Option2'], selected: 'Option1' });
-    expect(screen.getByText('Option1')).toHaveClass(
-      'bg-light border-gray-6 text-default font-semibold'
-    );
+    render(Radio, common);
+    expect(
+      screen.getByLabelText('Option 1').parentElement?.parentElement
```
mcous, line 23: This `parentElement.parentElement` feels a little brittle / coupled with implementation details. Could this be refactored so it's a little more semantically structured?

packages/core/src/lib/__tests__/radio.spec.ts
```diff
@@ -1,142 +1,108 @@
...
+    ).toHaveClass('font-semibold text-default');
   });
 
-  it('Marks the selected option as readonly if a readonly attribute of true has been specified', () => {
+  it('Marks the selected option as disabled if a disabled attribute of true has been specified', () => {
```
mcous, line 27: Same comment, we're checking styles but not HTML semantics

packages/core/src/lib/__tests__/radio.spec.ts
```diff
@@ -1,142 +1,108 @@
...
-    render(Radio, { options: ['Option1', 'Option2'], selected: 'Option1' });
-    expect(screen.getByText('Option1')).toBeVisible();
-    expect(screen.getByText('Option2')).toBeVisible();
+    render(Radio, common);
+    expect(screen.getByLabelText('Option 1')).toBeVisible();
+    expect(screen.getByLabelText('Option 2')).toBeVisible();
+    expect(screen.getByLabelText('Option 3')).toBeVisible();
   });
 
   it('Marks the selected option correctly', () => {
```
mcous, line 20: Should this test be checking more HTML semantics, like which element is `checked`?

mcous (commented):
I see a lot of styling checks in the test but not a lot of checking of HTML semantics / aria properties. The latter definitely seems more important to test; could the tests be adjusted accordingly?

packages/core/src/lib/icon/icons.ts
DTCurrie: So I actually think this was a case of vscode getting confused. I originally tried

```
$: getIcon = (option: string): IconName =>
  isSelected(option)
    ? 'radiobox-marked'
    : 'radiobox-blank';
```

And I was getting an error that I could not cast a `string` to an `IconName`. I thought this was an issue where it was not properly interpreting the type from the strings, but having restarted vscode and going to snap a screenshot of the error, it is no longer showing. I think the reason was, I added those two icons to `paths`, but in the `radio` component it was not detecting that I had added them, so it was just interpreting them as a `string`.

I feel like I need to put a sticky-note on my screen that says "Restart eslint/svelte server in vscode before debugging type issues." It comes up more than I would like.

packages/core/src/lib/__tests__/radio.spec.ts
DTCurrie: Sure, I just tried to take the previous tests and convert them over to make sure things still "worked" as expected, but I agree we should focus more on the html then the classes. I will take a pass at these tests.

mcous (approved):
🧪

PR #404, opened 2023-10-10
Expose that geopose

micheal-parks:
This PR exposes a `baseGeopose` property on `<NavigationMap>`. This is required to display the compass heading of a base both in app.viam.com (if a robot is connected via discovery service) or in the RDK. I also did just a little type cleanup and made type exports explicit.


packages/blocks/src/routes/navigation-map.svelte
```diff
@@ -81,6 +81,18 @@ const obstacles = [
     ],
   },
 ];
+
+let elapsed = 0
+let pose = { lng: -73.97, lat: 40.67, rotation: Math.random() * 360 }
+
+requestAnimationFrame(function frame (time) {
```
micheal-parks, line 88: This is just a fun example in the playground, it can get wacky

PR #406, opened 2023-10-11
Expose map `on:create` event and `map` binding on `<NavigationMap>`

micheal-parks:
There are cases in RDK and App when we would like to get the `MapLibre` map instance to do things such as change the map centerpoint. This PR exposes two ways to do this, via the `create` event and by adding a `bind:map`.


packages/blocks/src/lib/maplibre/index.svelte
```diff
@@ -115,9 +115,9 @@ onMount(() => {
 });
 
 onDestroy(() => {
-  map!.off('move', handleMove);
-  map!.off('resize', handleResize);
-  map!.off('style.load', handleCreate);
+  map?.off('move', handleMove);
```
micheal-parks, line 118: Just making this SSR friendly since `onDestroy` is called on the server.

ethanlook (approved):

PR #408, opened 2023-10-13
Fix toggle button legend margins and ignore all things __tests__

DTCurrie:
After seeing it in `app` I noticed the margins for the `legend`s were missing, also simplified our ignores for test files so anything in a `__tests__` directory is ignored.


micheal-parks (approved):
🚀

PR #410, opened 2023-10-16
Add persisted store, use to persist navigation map center and zoom

micheal-parks:
### Overview

Adds a new `prime-core` export, a special `Writable` store that persists to local or session storage:

```ts
import { persisted } from '@viamrobotics/prime-core'

// Adds storage persistence on the client, returns just a `Writable` on the server.
const value = persisted<string>('storageKey', 'initialValue', 'session' | 'local' | undefined)

value.set('hi') // Adds to storage, will be populated on next refresh.
```

Local/sessionstorage persistence is becoming more common in app, rdk, and prime. I've use store wrappers before and since we're svelting everywhere I figured that this is a good abstraction for our needs.

### Other changes

I've added my most immediate need for usage in `blocks`: persisting map center and zoom.

### Notes

An earlier version of this exists in RDK. I'll remove as a follow up if we like this one.

### DIFF
--- a/packages/core/src/lib/persisted.ts
+++ b/packages/core/src/lib/persisted.ts
@@ -6,6 +6,14 @@ const getStorage = (type: 'local' | 'session') => {
   return type === 'local' ? localStorage : sessionStorage;
 };
 
+const parse = <T>(value: string): T | null => {
+  try {
+    return JSON.parse(value) as T;
+  } catch {
+    return null;
+  }
+}
+
 const createPersisted = <T>(
   key: string,
   initialValue: T | null,
@@ -16,11 +24,7 @@ const createPersisted = <T>(
 
     // eslint-disable-next-line unicorn/no-negated-condition
     if (json !== null) {
-      try {
-        set(JSON.parse(json) as T);
-      } catch {
-        set(null);
-      }
+      set(parse<T>(json))
     } else {
       storage.setItem(key, JSON.stringify(initialValue));
     }
@@ -28,11 +32,7 @@ const createPersisted = <T>(
     const handleStorage = (event: StorageEvent) => {
       if (event.key === key) {
         if (event.newValue) {
-          try {
-            set(JSON.parse(event.newValue) as T);
-          } catch {
-            set(null);
-          }
+          set(parse<T>(event.newValue))
         } else {
           set(null);
         }
--- a/packages/core/src/lib/persisted.ts
+++ b/packages/core/src/lib/persisted.ts
@@ -1,0 +1,88 @@
+import { writable, type Writable } from 'svelte/store';
+
+const stores: Record<string, Writable<unknown> | undefined> = {};
+
+const getStorage = (type: 'local' | 'session') => {
+  return type === 'local' ? localStorage : sessionStorage;
+};
+
+const createPersisted = <T>(
+  key: string,
+  initialValue: T | null,
+  storage: ReturnType<typeof getStorage>
+) => {
+  const store = writable<T | null>(initialValue, (set) => {
+    const json = storage.getItem(key);
+
+    if (json) {
+      try {
+        set(JSON.parse(json) as T);
+      } catch {
+        set(null);
+      }
+    } else {
+      storage.setItem(key, JSON.stringify(initialValue));
+    }
+
+    const handleStorage = (event: StorageEvent) => {
+      if (event.key === key) {
+        if (event.newValue) {
+          try {
+            set(JSON.parse(event.newValue) as T);
+          } catch {
+            set(null);
+          }
+        } else {
+          set(null);
+        }
+      }
+    };
+
+    window.addEventListener('storage', handleStorage);
+
+    return () => window.removeEventListener('storage', handleStorage);
+  });
+
+  const { subscribe, set } = store;
+
+  return {
+    set(value: T): void {
+      storage.setItem(key, JSON.stringify(value));
+      set(value);
+    },
+    update(callback: (value: T | null) => T | null): void {
+      store.update((last) => {
+        const value = callback(last);
+        storage.setItem(key, JSON.stringify(value));
+        return value;
+      });
+    },
+    subscribe,
+  } satisfies Writable<T | null>;
+};
+
+/**
+ * Creates a writable store that persists in localStorage
+ *
+ * @param key The storage key
+ * @param initialValue The initial value to put in storage if no value exists.
+ * @param storageType 'local' or 'session'
+ */
+export const persisted = <T>(
+  key: string,
+  initialValue: T | null = null,
+  storageType: 'local' | 'session' = 'local'
+): Writable<T | null> => {
+  const browser =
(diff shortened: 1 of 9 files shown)


packages/core/src/lib/persisted.ts
```diff
@@ -0,0 +1,88 @@
+import { writable, type Writable } from 'svelte/store';
+
+const stores: Record<string, Writable<unknown> | undefined> = {};
```
mcous, line 3: A module level singleton isn't going to be workable for the SvelteKit app. Why is this record of all stores needed? I'm comfortable leaving it to the caller to avoid key collisions

packages/core/src/lib/persisted.ts
```diff
@@ -0,0 +1,88 @@
...
+};
+
+/**
+ * Creates a writable store that persists in localStorage
+ *
+ * @param key The storage key
+ * @param initialValue The initial value to put in storage if no value exists.
+ * @param storageType 'local' or 'session'
+ */
+export const persisted = <T>(
```
mcous, line 71: Should this `T` be marked as only JSONifiable values? `type-fest` has a [Jsonifiable](https://github.com/sindresorhus/type-fest/blob/main/source/jsonifiable.d.ts) type

packages/core/src/lib/persisted.ts
```diff
@@ -0,0 +1,88 @@
...
+) => {
+  const store = writable<T | null>(initialValue, (set) => {
+    const json = storage.getItem(key);
+
+    if (json) {
+      try {
+        set(JSON.parse(json) as T);
+      } catch {
+        set(null);
+      }
```
mcous, line 22: This safe parse is repeated in a few places, can it be moved to a function?

```ts
const parse = <T>(value: unknown): T => {
  try {
    return JSON.parse(value) as T;
  } catch {
    return null
  }
}
```

packages/core/src/lib/persisted.ts
```diff
@@ -0,0 +1,88 @@
...
+    window.addEventListener('storage', handleStorage);
+
+    return () => window.removeEventListener('storage', handleStorage);
+  });
+
+  const { subscribe, set } = store;
+
+  return {
+    set(value: T): void {
+      storage.setItem(key, JSON.stringify(value));
```
mcous, line 50: This will raise if the value is not JSON-ifiable. I don't object to the raising, but perhaps we want to raise a more actionable error?

packages/core/src/lib/persisted.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const createPersisted = <T>(
+  key: string,
+  initialValue: T | null,
+  storage: ReturnType<typeof getStorage>
+) => {
+  const store = writable<T | null>(initialValue, (set) => {
+    const json = storage.getItem(key);
+
+    if (json) {
```
mcous, line 17: `json` may be `0` or `''`

packages/core/src/lib/persisted.ts
```diff
@@ -0,0 +1,88 @@
...
+      } catch {
+        set(null);
+      }
+    } else {
+      storage.setItem(key, JSON.stringify(initialValue));
+    }
+
+    const handleStorage = (event: StorageEvent) => {
+      if (event.key === key) {
+        if (event.newValue) {
```
mcous, line 29: `event.newValue` may be `0` or `''`

mcous (changes requested):

packages/core/src/lib/persisted.ts
micheal-parks: Done!

packages/core/src/lib/persisted.ts
micheal-parks: Fixed! (the naughty way)

packages/core/src/lib/persisted.ts
micheal-parks: Oh I like that a lot. I'm down for that, will add.

micheal-parks: Done!

packages/core/src/lib/persisted.ts
micheal-parks: Fixed with jsonifiable ✨

micheal-parks: Oh, also do you mean throwing as well, or do we think TS is enough here?

packages/core/src/lib/persisted.ts
mcous: Oh yeah you're right about the window guard. That downgrades my concern to general "be careful about singletons." I think it risks not scaling in the future.

If we're relying on "this singleton is only used in the browser," I think we should make that more obvious in the code / comments to avoid someone shifting it around in the future

packages/core/src/lib/persisted.ts
mcous: I think TS is enough!

packages/core/src/lib/persisted.ts
```diff
@@ -0,0 +1,91 @@
...
+  return type === 'local' ? localStorage : sessionStorage;
+};
+
+const parse = <T>(value: string): T | null => {
+  try {
+    return JSON.parse(value) as T;
+  } catch {
+    return null;
+  }
+};
```
mcous, line 20:
```suggestion
const parse = <T>(value: string | null): T | null => {
  if (value === null) {
    return null
  }  

  try {
    return JSON.parse(value) as T;
  } catch {
    return null;
  }
};
```

packages/core/src/lib/persisted.ts
micheal-parks: Yep, agreed. Added a comment for now

packages/core/src/lib/persisted.ts
```diff
@@ -0,0 +1,91 @@
...
+        } else {
+          set(null);
+        }
+      }
+    };
+
+    window.addEventListener('storage', handleStorage);
+
+    return () => window.removeEventListener('storage', handleStorage);
+  });
```
mcous, line 51: Combined with my comment on `parse` above, we could simplify the body of our store to 

```suggestion
  initialValue = parse(storage.getItem(key)) ?? initialValue
  
  const store = writable<T | null>(initialValue, (set) => {
    storage.setItem(key, JSON.stringify(initialValue));

    const handleStorage = (event: StorageEvent) => {
      if (event.key === key) {
        set(parse<T>(event.newValue));
      }
    };

    window.addEventListener('storage', handleStorage);

    return () => window.removeEventListener('storage', handleStorage);
  });
```

packages/core/src/lib/persisted.ts
mcous: Are we worried about stores staying in scope even when they're not being used?

mcous (approved):

packages/core/src/lib/persisted.ts
micheal-parks: Hmm good question, I don't think so unless we're using a crazy amount of local / session storage once and then never again in any given client app lifecycle?

PR #411, opened 2023-10-16
Add code-snippet

DTCurrie:
Adds the `code-snippet` to prime-core.

Screenshot 2023-10-16 at 3 59 00 PM


packages/core/src/app.d.ts
```diff
@@ -11,6 +11,10 @@ declare global {
      * interface Platform {}
      */
   }
+
+  interface Window {
+    Prism: import('@types/prismjs');
+  }
```
mcous, line 17: Why is Prism on the window?

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,260 @@
+<!--
+@component
+  
+For rendering readonly blocks of code. Uses `prism` for syntax highlighting:
+
+https://prismjs.com
+
+```svelte
+<Breadcrumbs crumbs={['Dessert', 'Cookie']} />
+```
```
mcous, line 10: Needs to be updated

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,260 @@
...
+
+$: copyButtonLabel = 'Copy';
+$: copyButtonIcon = 'content-copy' as IconName;
+
+const dispatch = createEventDispatcher<{
+  copy: { succeeded: boolean; message: string };
+}>();
+
+const getCdnUrl = (src: string) =>
+  `https://cdnjs.cloudflare.com/ajax/libs/prism/${PRISM_VERSION}/${src}`;
```
mcous, line 80: Why aren't we using npm and importing the file(s) from `node_modules`? Relying on a CDN feels hacky

mcous (commented):
Why aren't we specifying prism as an npm dependency and using the `node_modules` files for imports? Downloading from a CDN and messing with `window` seems iffy, especially in a completely new component

packages/core/src/lib/code-snippet.svelte
DTCurrie: This is how it was previously handled, I believe to allow loading the library asynchronously. @micheal-parks and @mrloureed  may know better.

packages/core/src/app.d.ts
DTCurrie: It's added to the `window` by the CDN.

packages/core/src/lib/code-snippet.svelte
micheal-parks: That was me back in the good ol' days. If I recall correctly I did this only to lazy load this library. Doing a dynamic import from node_modules would be much better and worth doing during this port.

DTCurrie: Cool, should clean things up then. What about the CDN for the CSS?

micheal-parks: Beware though, I might have done it this way because of prism not playing well with vite or something. I'd usually go for a node_modules / dynamic import solution.

micheal-parks: You can dynamically import css as well, you can probably get the same files from node modules: https://colinchjs.github.io/2023-10-13/09-59-05-131860-can-you-dynamically-import-css-files-using-javascripts-dynamic-imports/

DTCurrie: @mcous @micheal-parks I threw together an implementation using dynamic imports and leveraging the `autoloader` plugin for `prism`.

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,224 @@
...
+     */
+    await import(
+      // @ts-expect-error no type declaration for this JS file
+      'prismjs/plugins/autoloader/prism-autoloader'
+    );
+
+    // Make sure the autoloader knows where to find our languages
+    (Prism.plugins.autoloader as { languages_path: string }).languages_path =
+      // TODO: do we want a better way to get the version?
+      'https://cdnjs.cloudflare.com/ajax/libs/prism/1.29.0/components/';
```
DTCurrie, line 165: So I am using the CDN here because otherwise we would have to include all of the potential languages we want to support along with our bundle, which does not seem ideal. One downside here is we are including the version number for prismjs in the CDN link, which means bumping our dev dependency on it could cause issues. Interested in suggestions for a better way to manage that.

micheal-parks: In the old version I'd import the version from package.json:

```ts
import pkg from '../../package.json';

const version = pkg.devDependencies.prismjs.replace('^', '');
```

That way we still have our package.json as a versioning source of truth

DTCurrie: So I tried it out and maybe I am missing something, but it works fine for the playground dev environment but fails in the storybook with this error:

```shell
Failed to resolve import "../../package.json" from "../core/dist/code-snippet.svelte". Does the file exist?
3:08:54 PM [vite] Internal server error: Failed to resolve import "../../package.json" from "../core/dist/code-snippet.svelte". Does the file exist?
  Plugin: vite:import-analysis
  File: /Users/devintcurrie/Projects/viam-prime/packages/core/dist/code-snippet.svelte:18:16
  29 |  const { Object: Object_1, console: console_1 } = globals;
  30 |  import "svelte/internal/disclose-version";
  31 |  import pkg from "../../package.json";
  ```

File: packages/core/src/lib/code-snippet.svelte (hunk 3 above)
Comment (micheal-parks) [reply packages/core/src/lib/code-snippet.svelte]: This looks nice!

File: packages/core/src/lib/code-snippet.svelte [hunk 5]
```diff
@@ -0,0 +1,230 @@
...
+    /* eslint-disable-next-line no-console */
+    console.error('Error initializing prismjs', error);
+  }
+});
+</script>
+
+<figure class={cx('flex flex-col gap-2', extraClasses)}>
+  {#if $$slots.default}
+    <figcaption><slot /></figcaption>
+  {/if}
```
DTCurrie, line 183: Added this to allow an easy way to accessibly caption the code-snippets. After talking to Nat, we would generally provide copy contextualizing the snippet above the code block, so this could be a useful way to associate that contextual copy with the `figure` itself. Open to not include this but seems like an easy feature to bake-in that could be handy.

packages/core/src/lib/code-snippet.svelte
DTCurrie: @micheal-parks So I actually stole the solution we use for getting the monaco version for the code editor:

```ts
import PrismPackage from 'prismjs/package.json';

...

const PRISM_VERSION = PrismPackage.version;
```

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,233 @@
...
+    padding: 0;
+    padding-left: 0.5rem;
+    padding-top: 0.5rem;
+  }
+
+  figure pre[class*='language-'] > code[class*='language-'] {
+    font-family: 'Roboto Mono Variable', 'Roboto Mono', ui-monospace, monospace;
+  }
+  </style>
+{/if}
```
mcous, line 233: Why can't these be added to the `class` normally? They all look like regular tailwind utilities

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,233 @@
...
+     */
+    await import(
+      // @ts-expect-error no type declaration for this JS file
+      'prismjs/plugins/autoloader/prism-autoloader'
+    );
+
+    // Make sure the autoloader knows where to find our languages
+    (Prism.plugins.autoloader as { languages_path: string }).languages_path =
+      // TODO: do we want a better way to get the version?
+      `https://cdnjs.cloudflare.com/ajax/libs/prism/${PRISM_VERSION}/components/`;
```
mcous, line 170: Hate to say it, but highlight.js looks to have a much better ESM / dynamic [import story](https://highlightjs.org/#usage) that Prism, and looks to be more actively maintained. The third-party CDN approach still leaves a bad taste in my mouth

That being said, the best way to get the Prism version would be to read it from `package.json` in `vite.config.ts` and inject it into the build via the [define](https://vitejs.dev/config/shared-options.html#define) option

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,230 @@
...
+</CodeSnippet>
+```
+-->
+<svelte:options immutable />
+
+<script
+  context="module"
+  lang="ts"
+>
+import Prism from 'prismjs';
```
mcous, line 20: Is this statically importing all of Prism? That seems concerning

packages/core/src/lib/code-snippet.svelte
mcous: Been playing with highlight.js for a minute, and it's not exactly going well 😢

packages/core/src/lib/code-snippet.svelte
DTCurrie: I actually think I can just import the function we need according to a doc I was looking at but forgot to try it out. Will give it a go tomorrow.

packages/core/src/lib/code-snippet.svelte
DTCurrie: I think @mrloureed had a similar issue with it.

packages/core/src/lib/code-snippet.svelte
DTCurrie: They get overridden by the prism styles. We were using important modifiers on every tailwind class but forcing important felt worse than using more specific selectors. These are essentially a light “reset” of the prism css to not include backgrounds and borders, and use the correct padding and fonts.

If we feel it’s better to use tailing utilities with important modifiers then I can revert it.

packages/core/src/lib/code-snippet.svelte
DTCurrie: Also I tried using define with vite but it looks like vite wasn’t properly replacing it in the build, which meant I had to re-declare it in storybook which we definitely don’t want users needing to do. Glad to re-examine that, but is that better than just grabbing the library version from the library itself? The nice thing there is I didn’t have to do any string replacements.

packages/core/src/lib/code-snippet.svelte
DTCurrie: Appears to be working, we just need to import the `highlightElements` function and the `plugins` Record.

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * We use the prism autoloader to handle loading in language grammar files. The
+ * default path for the grammar files is a CDN link so we don't have to include
+ * grammar files in our bundle. If you prefer to point to your own grammars and
+ * bypass the CDN, define the path to those files.
+ *
+ * See: https://prismjs.com/plugins/autoloader/
+ */
+export let grammarsPath = `https://cdnjs.cloudflare.com/ajax/libs/prism/${PRISM_VERSION}/components/`;
```
DTCurrie, line 75: After the conversation around whether to use the CDN or not, I added this as a potential short-circuit. This will allow users to define their own path to grammar files, if they don't want to lazy-load them from the CDN and grab them from a local static path for example. This may not be necessary, but I could see us including the grammars for SDK languages we support so we can fetch those locally in `app`.

micheal-parks (approved):

packages/core/package.json
```diff
@@ -72,6 +73,8 @@
     "prettier": "^3.0.3",
     "prettier-plugin-svelte": "^3.0.3",
     "prettier-plugin-tailwindcss": "^0.5.4",
+    "prism-themes": "^1.9.0",
+    "prismjs": "^1.29.0",
```
mcous, line 77: Can someone remind me the difference between adding these to `dependencies` vs `devDependencies` in `svelte-package` world? My instinct would be that they need to be in `dependencies`, like other libs we import directly in components

packages/core/src/lib/__tests__/code-snippet.spec.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+import CodeSnippet from '$lib/code-snippet.svelte';
+</script>
+
+<CodeSnippet
+  language="json"
```
mcous, line 6: Something clever I saw in Svelte highlight package was to pass the language file itself, rather than a string, as the `language` prop. No need to deal with it in this PR, but I liked the idea in general

```ts
<script lang="ts">
import CodeSnippet from '$lib/code-snippet.svelte';
import json from 'prism/components/prism-json.min.js';
</script>

<CodeSnippet language={json} />
```

packages/core/src/lib/__tests__/code-snippet.spec.svelte
```diff
@@ -0,0 +1,10 @@
...
+  code={`{ myThing: true }`}
+>
+  This is the caption text.
```
mcous, line 9: Can this be in a named slot? I find it a little confusing that the main content (the code) goes in via a prop and the ancillary content (the caption) goes in via the default slot; it feels backwards in terms of intent

packages/core/src/lib/__tests__/code-snippet.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  it('Renders without the copy button', () => {
+    render(CodeSnippet, { ...common, showCopyButton: false });
+
+    expect(screen.queryByLabelText('Copy')).toBeNull();
+  });
+
+  it('Renders with the passed dependencies', () => {
+    render(CodeSnippet, { ...common, dependencies: ['dep1', 'dep2'] });
+
+    const code = document.querySelector('code.language-json');
```
mcous, line 42: Is `getByRole` code a thing? Also, should we be checking that there's a `<pre>` involved, too?

packages/core/src/lib/__tests__/code-snippet.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+      cx: cxTestArguments,
+    });
+
+    expect(screen.getByRole('figure')).toHaveClass(cxTestResults);
+  });
+
+  it('Renders with the copy button', () => {
+    render(CodeSnippet, common);
+
+    expect(screen.getByLabelText('Copy')).toBeInTheDocument();
```
mcous, line 30: Can this:

- Check for a button with a name rather than the label
- Actually test the copy functionality?

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,227 @@
...
+    dispatch('copy', {
+      succeeded: false,
+      message: 'Failed to copy snippet to the clipboard',
+    });
+  }
+
+  window.setTimeout(() => {
+    copyButtonLabel = 'Copy';
+    copyButtonIcon = 'content-copy';
+  }, 2000);
```
mcous, line 114: Do we need cleanup for this?

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,227 @@
...
+export let grammarsPath = `https://cdnjs.cloudflare.com/ajax/libs/prism/${PRISM_VERSION}/components/`;
+
+/** Additional CSS classes to pass to the aside. */
+let extraClasses: cx.Argument = '';
+export { extraClasses as cx };
+
+let element: HTMLElement | undefined;
+
+$: copyButtonLabel = 'Copy';
+$: copyButtonIcon = 'content-copy' as IconName;
```
mcous, line 84: Should this be a single state that is derived into the proper label and icon per state? e.g. `idle`, `success`, `error`. Similar functionality is present in `RestrictedTextInput`

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,227 @@
...
+    highlight();
+  }
+}
+
+onMount(async () => {
+  try {
+    /**
+     * Load the themes, for vite to properly discover them we cannot use string templates,
+     * so we use a switch statement for our supported themes.
+     */
```
mcous, line 139: We may be able to use [`import.meta.glob`](https://vitejs.dev/guide/features.html#glob-import) to help

mrloureed (approved):
LGTM

mcous (approved):

packages/core/package.json
DTCurrie: Oh good catch. We had them as `devDependencies` because we were dynamically loading everything from the CDN. Now that we don't do that, we should probably move `prismjs` and `prism-themes` to dependencies now.

packages/core/src/lib/__tests__/code-snippet.spec.svelte
DTCurrie: Sure

packages/core/src/lib/code-snippet.svelte
DTCurrie: I believe I tried that originally which is where I discovered the issue with vite but I can try again.

packages/core/src/lib/__tests__/code-snippet.spec.ts
DTCurrie: Not that I am aware of, there is a `figure` role which will get the root element. I check if there is a `pre` I guess just via the selector statement. I can also just get the `figure` and query against that if we prefer.

packages/core/src/lib/__tests__/code-snippet.spec.ts
DTCurrie: On the first point sure, on the second point I guess but IIRC doesn't jsdom not really support the `navigator.clipboard` API. I feel like I recall having to put in some hacky workarounds/mocks, but not sure if something has changed there. I can put something in there if we want though.

packages/core/src/lib/code-snippet.svelte
DTCurrie: Yeah in that doc it says:

> You should also be aware that all the arguments in the import.meta.glob must be passed as literals. You can NOT use variables or expressions in them.

If we tried to use the theme as a variable I am pretty sure it will still yell at us.

packages/core/src/lib/__tests__/code-snippet.spec.svelte
DTCurrie: Interesting, but right now all of the `prismjs` stuff is typed as `any`, and I don't know if I love having a `export let language: any;` prop vs just accepting a string.

packages/core/src/lib/code-snippet.svelte
mcous: import.meta.glob returns an object, keyed by module name. So you'd use the variable when accessing the module via the object, not when constructing the glob. That being said, I don't remember if it plays nicely with node_modules, any probably isn't worth the effort

DTCurrie: It actually might, I was trying to query it by the file name but I think it is the full module path that is the key. I am testing it out now to see.

packages/core/src/lib/code-snippet.svelte
```diff
@@ -0,0 +1,220 @@
...
+  }
+};
+
+$: if (code) {
+  highlight();
+}
+
+onMount(async () => {
+  // This throws a warning but it is a lie: https://github.com/vitejs/vite/issues/12001
+  const glob = import.meta.glob('/node_modules/prism-themes/themes/*.css');
```
mcous, line 141: Might want to limit this glob to the files we actually use

```suggestion
  const glob = import.meta.glob('/node_modules/prism-themes/themes/prism-{vs,vsc-dark-plus}.css');
```

DTCurrie: Ah true, though it does raise the question of if we want to support more themes. Might play around with that and see if we want to open that up or not.

DTCurrie: Oh so I am glad we tested this, fun fact if you change the theme on one code snippet you change the theme on _all_ of them with this new setup. So ... that may not be great. I guess we need to figure out how to scope those CSS files to the component.

DTCurrie: I looked through the code base and we never use anything but the `vs` theme, and I just checked the `vsc-dark-plus` theme does not work well with JSON and the background color we are applying, number values are almost unreadable:

Screenshot 2023-10-24 at 5 50 22 PM

I added JSON later and didn't catch that, all the other languages looks ok with the snippets I provided but they don't test the syntax exhaustively. I may come up with better snippets that illustrate the full syntax of each language as best as possible to help catch these kind of issues.

That said, I think for now the best argument is to just keep things simple and not include the `theme` prop at all. Thoughts @mcous @micheal-parks @mrloureed ?

PR #415, opened 2023-10-23
APP-2991: add selection-ellipse icon

mcous:
Adds the `selection-ellipse` icon. While working on this, I noticed...

- The storybook icon name tooltip was broken, and nobody noticed because
- The storybook stories weren't getting type checked

...so I fixed those, too.

I also learned that I don't know how to spell "ellipse", so please keep an eye out for typos

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.51",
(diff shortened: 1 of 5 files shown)


DTCurrie:
@mrloureed noticed the icon story issue earlier, not sure why it didn’t get fixed. Glad we are running checks now though!

DTCurrie (approved):
✅

PR #416, opened 2023-10-24
Remove code-snippet theme prop

DTCurrie:
Per this conversation: https://github.com/viamrobotics/prime/pull/411#discussion_r1370854790

image

The dark theme does not work well for JSON, which intend to use this code-snippet for frequently. If we want to introduce a dark mode, we should probably think about that more comprehensively.

Also, after we introduced the reactivity statement to rehighligh on code changes, we need to start escaping the `code` passed to this component when setting it as the `innerHtml` of the `element`.

- [x] Add tests for `use-timeout`

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.54",
+  "version": "0.0.55",
   "publishConfig": {
     "access": "public"
   },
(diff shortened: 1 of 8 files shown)


DTCurrie:
Any suggestions for how to test this: https://github.com/viamrobotics/prime/blob/main/packages/core/src/lib/use-timeout.ts

Because it uses lifecycle hooks, it looks like it needs to rendered as a part of a component which makes sense. I tried setting that up but it appears like the tests all time out. Wanted to see if there is any guidelines with testing timeouts in testing-library/svelte before I go digging.

micheal-parks (approved):

PR #418, opened 2023-10-26
APP-2992: Add icons for configure sidebar menu

anjinai:
**APP-2992: Add icons for configure sidebar menu**

- Adds the tools, json, and frame icons to prime core. 
- Adds the icons to legacy too just in case.


mrloureed (approved):
LGTM

PR #420, opened 2023-11-01
APP-3028: add opaque grey variants

mcous:
## Overview

After talking with Nat, this PR adds two colors to our scale that are common in our codebase and have been popping up in new code: `rgba(0, 0, 0, 0.04)` and `rgba(0, 0, 0, 0.08)` - aka the hover and active variants for the ghost button.

I'm open to whatever names, but I went with `ghost-light` and `ghost-medium`.

## Change log

- Add `ghost-light` and `ghost-medium` color variants
- Add `.vscode/settings` to git tracking for:
    - Use `typescript` from `node_modules`
    - Use `prettier` for formatting Svelte
    - Use `eslint` for checking Svelte
    - Add `cx` prop and `cx()` function to Tailwind intellisense

### DIFF
--- a/packages/core/src/lib/button/button.svelte
+++ b/packages/core/src/lib/button/button.svelte
@@ -64,7 +64,7 @@ $: handleDisabled = preventHandler(disabled);
         variant === 'primary' && !disabled,
       'border-gray-9 bg-gray-9 text-white hover:border-black hover:bg-black active:bg-[#000]':
         variant === 'dark' && !disabled,
-      'border-transparent text-default hover:bg-[rgba(0,0,0,0.04)] active:border-[rgba(0,0,0,0.08)] active:bg-[rgba(0,0,0,0.08)]':
+      'border-transparent text-default hover:bg-ghost-light active:border-ghost-medium active:bg-ghost-medium':
         variant === 'ghost' && !disabled,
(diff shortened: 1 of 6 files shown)


packages/core/src/lib/button/button.svelte
```diff
@@ -64,7 +64,7 @@ $: handleDisabled = preventHandler(disabled);
         variant === 'primary' && !disabled,
       'border-gray-9 bg-gray-9 text-white hover:border-black hover:bg-black active:bg-[#000]':
         variant === 'dark' && !disabled,
-      'active-border-[rgba(0,0,0,0.08)] border-transparent text-default hover:bg-[rgba(0,0,0,0.04)] active:bg-[rgba(0,0,0,0.08)]':
```
mcous, line 67: Old bug in prime that was carried into core: active state has not been applied on ghost buttons because of an `active-` vs `active:` typo

DTCurrie (approved):
👻

packages/core/src/lib/button/button.svelte
```diff
@@ -64,7 +64,7 @@ $: handleDisabled = preventHandler(disabled);
...
+      'border-transparent text-default hover:bg-ghost-light active:border-ghost-medium active:bg-ghost-medium':
```
mcous, line 67: Old bug in prime that was carried into core: active state has not been applied on ghost buttons because of an `active-` vs `active:` typo

PR #422, opened 2023-11-08
Replace Modal.isOpen with controlled boolean

mcous:
## Overview

When writing the Modal component for prime-core, I suggested we ditch events in favor of the two-way data binding of an `isOpen` store. After some usage in the app, we've discovered this is a pretty awkward API.

This PR switches the modal back to a plain, controlled `isOpen: boolean` prop and a `close` event, which should be more predictable. As a result, this means the modal is no longer capable of closing itself; it must emit the `close` event and the parent must chose to close it.

I think there are continuing improvements to be made here, especially given the awkwardness of combining the `close` event with any buttons placed in the `primary` and `secondary` slots, but I think this is a step in the right direction.

**This is a breaking change to prime-core.** See the companion PR in the app: https://github.com/viamrobotics/app/pull/2873

## Review requests

- Read through unit test updates
- Test out the playground
- Check storybook

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.58",
+  "version": "0.0.59",
   "publishConfig": {
     "access": "public"
   },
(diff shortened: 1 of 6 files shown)


anjinai (approved):
lgtm too - Followed review requests and have the same comment as emily re: the modal not being visible in storybook.

PR #423, opened 2023-11-13
RSDK-4836 Add control to navigation map to follow the base position

ethanlook:
This follows the base when the user toggles the new button in the bottom right.

Note: this does not stop following the base when the user interacts with the map. MapLibre fires move events during `map.setCenter` and blocks drag events during move - I'm not sure there's a way to get around this. The user will need to toggle the button to stop following and then move the map. Zoom still works while following.

## Change log

- Moved 2D/3D buttons to top row
- Added `navigation-variant` and `navigation-variant-outline` icons
    - Sorted icon names alphabetically
- Added button to toggle "follow base" mode in bottom right
- Added `rAF` loop to follow the base position

## Review requests

Please pull this down and give it a try! (We need a test environment!)

### DIFF
--- a/packages/blocks/src/lib/navigation-map/components/map.svelte
+++ b/packages/blocks/src/lib/navigation-map/components/map.svelte
@@ -36,6 +36,23 @@ const toggleTileset = () => {
 let didHoverTooltip = Boolean(
   localStorage.getItem('navigation-service-card-tooltip-hovered')
 );
+
+let isFollowingBase = false;
+const followBase = () => {
+  if (baseGeoPose && isFollowingBase) {
+    map?.setCenter(baseGeoPose);
+    requestAnimationFrame(followBase);
+  }
+};
+
+const startFollowingBase = () => {
+  isFollowingBase = true;
+  requestAnimationFrame(followBase);
+};
+
+const stopFollowingBase = () => {
+  isFollowingBase = false;
+};
 </script>
 
 <div class="relative h-full w-full items-stretch sm:flex">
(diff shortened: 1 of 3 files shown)


packages/blocks/src/lib/navigation-map/components/map.svelte
```diff
@@ -36,6 +36,23 @@ const toggleTileset = () => {
...
+};
+
+const startFollowingBase = () => {
+  isFollowingBase = true;
+  requestAnimationFrame(followBase);
+};
+
+const stopFollowingBase = () => {
+  isFollowingBase = false;
+};
```
ethanlook, line 55: This could definitely be consolidated to `toggleFollowingBase` but if I can figure it out we want to `stopFollowingBase` when the user drags the map.

packages/blocks/src/lib/navigation-map/components/map.svelte
```diff
@@ -103,15 +120,31 @@ let didHoverTooltip = Boolean(
...
+    <div class="absolute bottom-10 right-2 z-10">
+      <Button
+        disabled={!baseGeoPose}
+        on:click={() => {
+          if (isFollowingBase) {
+            stopFollowingBase();
+          } else {
+            startFollowingBase();
+          }
+        }}
```
mcous, line 140: We can do a little less here

```suggestion
        on:click={isFollowingBase ? stopFollowingBase : startFollowingBase}
```

packages/blocks/src/lib/navigation-map/components/map.svelte
mcous: Could also do a little reactive helper:

```ts
$: toggleFollowingBase = isFollowingBase ? stopFollowingBase : startFollowingBase
```

mcous (commented):
Did not test, but the code LGTM except for the lack of callback cleanup

packages/blocks/src/lib/navigation-map/components/map.svelte
```diff
@@ -36,6 +36,23 @@ const toggleTileset = () => {
...
+  }
+};
+
+const startFollowingBase = () => {
+  isFollowingBase = true;
+  requestAnimationFrame(followBase);
+};
+
+const stopFollowingBase = () => {
+  isFollowingBase = false;
```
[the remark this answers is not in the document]
ethanlook: Great catch - done

packages/blocks/src/lib/navigation-map/components/map.svelte
ethanlook: 🤦🏼  thank you

mcous (approved):
LGTM

PR #424, opened 2023-11-13
APP-3122: Add custom icons for builder tree nav R2D2

anjinai:
## Changes
- Adds svg paths of custom icons (sourced from Figma) to Prime for usage in the robot details page redesign.
- Also adds the apple-keyboard-command MDI icon.

## Screenshots
- Figma
Screen Shot 2023-11-14 at 3 06 41 PM


packages/core/src/lib/icon/icon.svelte
```diff
@@ -64,6 +64,8 @@ const hasNameProperty = Object.hasOwn(paths, name);
 >
   <path
     d={paths[name]}
+    fill-rule="evenodd"
+    clip-rule="evenodd"
```
mrloureed, line 68: Can you confirm that the clip-rule is necessary? After reading docs on MDN I think it is not.

packages/core/src/lib/icon/icons.ts
```diff
@@ -77,6 +77,28 @@ export const paths = {
   'robot-outline': MDI.mdiRobotOutline,
   domain: MDI.mdiDomain,
   'database-outline': MDI.mdiDatabaseOutline,
+  'part-awaiting-setup':
```
mrloureed, line 80: Should we distinguish the custom icon names by prepending them with viam or custom?

mrloureed (changes requested):

packages/core/src/lib/icon/icon.svelte
anjinai: Thanks Jason, removed!

packages/core/src/lib/icon/icons.ts
anjinai: good call, updated to prepend custom icons with "viam"

mrloureed (approved):
LGTM!

Oabraham1 (approved):
LGTM, makes perfect sense!!!

PR #426, opened 2023-11-16
fix mobile layout for notifications

mrloureed:
Layout fix for notifications (mobile now is 100% screen width)
image

Bug fix when text wraps:
image

Current bug when text wraps:
image


mcous (approved):
Makes sense to me!

PR #429, opened 2023-11-17
RSDK-4825 Add plan rendering to navigation map

ethanlook:
## Change log

- Rename "obstacle layer" to "scene layer"
- Add type, props, and stores for plans
- Add plan component with sphere vertices and line segment edges
- Update render loop to allow for multiple `LngLat`s for geometries with buffer attributes

image


packages/blocks/src/lib/navigation-map/index.svelte
```diff
@@ -18,10 +18,11 @@
 import { environment as envStore } from './stores';
 import type { Map as MapType } from 'maplibre-gl';
 import Map from './components/map.svelte';
-import { type Obstacle, NavigationTab, type NavigationTabType } from './types';
+import { type Obstacle, NavigationTab, type NavigationTabType, type Plans } from './types';
```
mcous, line 21: Needs format

packages/blocks/src/lib/navigation-map/index.svelte
```diff
@@ -55,6 +59,7 @@ $: $tabStore = tab;
 $: $tabsStore = tabs;
 $: $waypointsStore = waypoints;
 $: $obstaclesStore = obstacles;
+$: $plansStore = plans;
```
mcous, line 62: Not new to this PR, but this sort of looks like this component should be providing a context rather than using singleton stores. It looks like multiple `NavigationMap` components might not be able to co-exist

packages/blocks/src/lib/navigation-map/plugins/render.ts
mcous, line 1: Not new to this PR, but much of the logic in this module looks like a good candidate for unit testing, which could also provide useful design pressure to decouple from the APIs of our third-party dependencies

packages/blocks/src/lib/navigation-map/stores.ts
```diff
@@ -1,13 +1,14 @@
 import { writable } from 'svelte/store';
 import { currentWritable } from '@threlte/core';
 import type { Waypoint } from '$lib';
-import { type Obstacle, NavigationTab, type NavigationTabType } from './types';
+import { type Obstacle, NavigationTab, type NavigationTabType, type Plans } from './types';
```
mcous, line 4: Needs format

mcous (approved):
Needs a `pnpm format` run, but otherwise LGTM. Added some thoughts about the existing module structure (and testing) for future conversations

packages/blocks/src/lib/navigation-map/components/plan.svelte
```diff
@@ -0,0 +1,26 @@
+<script lang="ts">
+import type { LngLat } from '$lib/maplibre/types';
+import { T } from '@threlte/core';
+import { Color, type ColorRepresentation } from 'three';
+
+export let plan: LngLat[];
+export let color: ColorRepresentation;
+
+// The typings for color are incorrect - they expect a number but need a THREE.Color.
+$: threeColor = new Color(color) as unknown as number;
```
micheal-parks, line 10: We fixed this upstream in `@types/three`. This should be resolved up updating everything. Up for me submitting a dependency update PR today so that we don't have to do this?

packages/blocks/src/lib/navigation-map/index.svelte
micheal-parks: Yep, there's a bug ticket for this, with solution being move to context :)

packages/blocks/src/lib/navigation-map/components/plan.svelte
```diff
@@ -0,0 +1,26 @@
...
+    <T.SphereGeometry />
+    <T.MeshPhongMaterial color={threeColor} />
+  </T.Mesh>
+{/each}
+
+<T.Line userData.lngLat={plan}>
+  <T.BufferGeometry />
+  <T.LineDashedMaterial
+    color={threeColor}
+    args={[{ dashSize: 0.5, gapSize: 0.5 }]}
```
micheal-parks, line 24:
```suggestion
    dashSize={0.5}
    gapSize={0.5}
    args={[{ dashSize: 0.5, gapSize: 0.5 }]}
```
In general I'd say props are preferrable to the args prop unless there's a good reason just cause it's more readable

packages/blocks/src/lib/navigation-map/plugins/render.ts
```diff
@@ -56,19 +58,32 @@ export const renderPlugin = () => {
...
-        const ox = mercatorOffset.x / mercatorScale;
-        const oy = mercatorOffset.y / mercatorScale;
-        object.position.set(cx - ox, 0, cy - oy);
+        if (Array.isArray(lngLat)) {
+          (object as THREE.Mesh).geometry.setFromPoints(
+            lngLat.map((ll) => new THREE.Vector3(...lngLatToPosition(ll)))
+          );
+          if (object instanceof THREE.Line) {
+            object.computeLineDistances();
+          }
```
micheal-parks, line 83:
```suggestion
          if (object instanceof THREE.Line) {
            object.geometry.setFromPoints(
              lngLat.map((ll) => new THREE.Vector3(...lngLatToPosition(ll)))
            );
            object.computeLineDistances();
          }
```

packages/blocks/src/lib/navigation-map/components/plan.svelte
ethanlook: That would be awesome, thanks!

packages/blocks/src/lib/navigation-map/plugins/render.ts
ethanlook: Agreed!

packages/blocks/src/lib/navigation-map/components/plan.svelte
micheal-parks: Merged! You should be able to clean this up now ✨

ethanlook: Thank you!!

ethanlook: Done!

packages/blocks/src/lib/navigation-map/plugins/render.ts
```diff
@@ -56,19 +58,32 @@ export const renderPlugin = () => {
         .fromArray(viewProjectionMatrix)
         .multiply(cameraTransform);
 
+      const lngLatToPosition = (lngLat: LngLat): [number, number, number] => {
```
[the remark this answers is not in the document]
ethanlook: Fixed this up - should be simpler now!

micheal-parks (approved):

PR #430, opened 2023-11-20
Design maintenance nov 17 2023

mrloureed:
New resource pill component added to PRIME core. This came out of design maintenance for new features in APP.

image


packages/core/src/lib/__tests__/resource-pill.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+describe('ResourcePill', () => {
+  it('Renders text within the resource pill if a value attribute is specified', () => {
+    render(ResourcePill, { value: 'test', icon: 'viam-service' });
+    expect(screen.getByText('test')).toBeVisible();
+  });
+
+  it('Renders a normal resource pill that is removable if removable attribute has been specified', () => {
+    render(ResourcePill, { icon: 'viam-service', removable: true });
+    expect(screen.getByRole('button')).toBeInTheDocument();
```
micheal-parks, line 14: Does this button have a label for accessibility that we can select instead?

mrloureed: I will check!

ethanlook (commented):
What's the difference between a pill and a resource pill? It's not clear to me why we aren't just updating the pill component and how we would choose between the two.

mrloureed:
> What's the difference between a pill and a resource pill? It's not clear to me why we aren't just updating the pill component and how we would choose between the two.

The biggest distinguisher is the design. They are similar but have different functionality and style. They are not removable by default, where normal pills are. I'm open to recommendations.

packages/core/src/lib/resource-pill.svelte
```diff
@@ -0,0 +1,69 @@
...
+<svelte:options immutable />
+
+<script lang="ts">
+import { Icon, type IconName } from '$lib';
+import cx from 'classnames';
+
+import { createEventDispatcher } from 'svelte';
+
+/** The text in the resource pill. */
+export let value = '';
```
mcous, line 19: I would _really_ prefer this to be a default `<slot />` rather than a `value` prop

packages/core/src/lib/resource-pill.svelte
```diff
@@ -0,0 +1,69 @@
...
+/** Additional CSS classes to pass to the resource pill. */
+let extraClasses: cx.Argument = '';
+export { extraClasses as cx };
+
+/** The icon shown in the button. */
+export let icon: IconName;
+
+const dispatch = createEventDispatcher<{
+  /** When the resource pill is removed. */
+  remove: { value: string };
```
mcous, line 33: If we switch to `<slot />`, I think we should remove the `value` from the event. The parent can distinguish via how it attached `on:remove` listeners, and I think it complicates the component to include it

mcous (commented):
I think I agree with Ethan's take: I don't quite get it. I also dislike how this encodes an idea like a "resource" into PRIME.

Could we consider something a little simpler and more composable?

- A `<Pill>` component that accepts any children via `<slot />` and applies styles
- A `<RemovablePill>` component that passes a remove button into `<Pill>` and adds a `remove` event

Then, if the **app** - the place where "resource" means something - needs a resource pill, it can also define its own resource pill in terms of `<Pill>` by passing an `<Icon>` in

packages/core/src/lib/resource-pill.svelte
mrloureed: Being that I've moved everything into `Pill` I think we'll have to do that as a follow up as it would require updates in APP and RDK.

packages/core/src/lib/pill.svelte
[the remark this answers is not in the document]
mrloureed: I think that makes sense, however I think it should be a follow up maybe along with slots.

packages/core/src/lib/pill.svelte
```diff
@@ -44,20 +53,30 @@ const handleRemove = () => {
...
+      'gap-1 bg-medium px-2 py-0.5': variant === 'primary',
+      'h-6 gap-1.5 border border-medium bg-light pl-1.5 pr-2':
+        variant === 'resource',
     },
     extraClasses
   )}
   aria-disabled={disabled ? true : undefined}
   aria-readonly={readonly ? true : undefined}
 >
+  {#if icon && variant === 'resource'}
```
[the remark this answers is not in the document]
mrloureed: No. We do not want icons for the primary vafriant.

mcous (approved):

packages/core/src/lib/resource-pill.svelte
mrloureed: Follow up ticket: https://viam.atlassian.net/browse/APP-3275

PR #435, opened 2023-11-27
Make obstacle coloring declarative

ethanlook:
This is prep to render transient obstacles differently. By making coloring declarative, `blocks` does not need to determine what is a transient obstacle.

## Change log

- Add color and label to `Obstacle` type
- Update colors used in dev environment - cyberpunk for static and hologram for transient
    - Note: app and rdk will need to declare these colors as well
- Obstacles become deselected if they are already clicked
- Render a color chip for each obstacle in the list of obstacles, including a tooltip label
- (cleanup) Export `Plans` type

Screenshot 2023-11-27 at 1 56 50 PM

### DIFF
--- a/packages/blocks/src/lib/navigation-map/components/draw-tool.svelte
+++ b/packages/blocks/src/lib/navigation-map/components/draw-tool.svelte
@@ -15,6 +15,7 @@ import {
 } from 'maplibre-gl';
 import { view } from '../stores';
 import * as math from '../lib/math';
+import { theme } from '@viamrobotics/prime-core/theme';
 
 interface $$Events extends Record<string, unknown> {
   /** Fires when a rectangle is drawn. */
@@ -128,6 +129,6 @@ $: if (drawing) {
         on:create={handleGeometryCreate}
       />
     {/if}
-    <T.MeshPhongMaterial color="red" />
+    <T.MeshPhongMaterial color={theme.extend.colors.cyberpunk} />
   </T.Mesh>
 </T.Group>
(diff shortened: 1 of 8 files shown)


packages/blocks/src/lib/navigation-map/lib/create-obstacle.ts
```diff
@@ -10,5 +11,7 @@ export const createObstacle = (
     name,
     location: { lng: lngLat.lng, lat: lngLat.lat },
     geometries: [createGeometry(type)],
+    label: 'static',
+    color: theme.extend.colors.cyberpunk,
```
ethanlook, line 15: This is not declarative - when creating a new obstacle, we hard-code that it is static and cyberpunk color.

I think the correct solution would be to not update the list of obstacles [here](https://github.com/viamrobotics/prime/blob/main/packages/blocks/src/lib/navigation-map/components/scene.svelte#L51-L54) and propagate a `add-obstacle` event. Then we could let the obstacles be uncontrolled and purely declarative.

packages/blocks/src/lib/navigation-map/components/draw-tool.svelte
```diff
@@ -128,6 +129,6 @@ $: if (drawing) {
         on:create={handleGeometryCreate}
       />
     {/if}
-    <T.MeshPhongMaterial color="red" />
+    <T.MeshPhongMaterial color={theme.extend.colors.cyberpunk} />
```
micheal-parks, line 132: 🎸👩‍🎤🔥

micheal-parks (approved):
🚀

PR #436, opened 2023-11-29
APP-3122: make floating menu controlled, add aria props

mcous:
This PR makes two changes to the `<FloatingMenu>` component required by work on APP-3122:

- We need to be able to add accessible attributes to the control element to add a tooltip
- We need to control the open state of the menu to set styles properly

### DIFF
--- a/packages/core/src/lib/context-menu/context-menu.svelte
+++ b/packages/core/src/lib/context-menu/context-menu.svelte
@@ -17,6 +17,12 @@ import cx from 'classnames';
 /** ID attribute of the menu element. */
 export let id: string;
 
+/** Accessible label of the menu. */
+export let label: string | undefined = undefined;
+
+/** ID of the element ID that labels the menu. */
+export let labeledBy: string | undefined = undefined;
+
 /** Additional CSS classes to pass to the menu. */
 let extraClasses: cx.Argument = '';
 export { extraClasses as cx };
@@ -24,6 +30,8 @@ export { extraClasses as cx };
 
 <div
   {id}
+  aria-label={label}
+  aria-labelledby={labeledBy}
   role="menu"
   class={cx(
     'border border-medium bg-white py-1 shadow-sm filter-none',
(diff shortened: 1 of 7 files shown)


Kschappacher (approved):
LGTM

packages/core/src/lib/context-menu/floating-menu.svelte
```diff
@@ -5,21 +5,24 @@ import { clickOutside } from '$lib/click-outside';
...
 import ContextMenu from './context-menu.svelte';
 
+export let isOpen: boolean;
+export let label: string | undefined = undefined;
+export let describedBy: string | undefined = undefined;
 export let placement: FloatingMenuPlacement = 'bottom-start';
 export let offset = 0;
 export let buttonCX: cx.Argument = '';
 export let menuCX: cx.Argument = '';
+export let onChange: (isOpen: boolean) => unknown;
```
ehhong, line 15: [q] what is the thought behind this versus dispatching an event like the modal?

packages/core/src/lib/context-menu/context-menu.svelte
```diff
@@ -17,13 +17,21 @@ import cx from 'classnames';
 /** ID attribute of the menu element. */
 export let id: string;
 
+/** Accessible label of the menu. */
+export let label: string | undefined = undefined;
+
+/** ID of the element ID that labels the menu. */
+export let labeledBy: string | undefined = undefined;
```
ehhong, line 24: [minor] labelledBy vs labeledBy? might be good to keep consistent with the aria class

mcous: I cannot spell 🤦

packages/core/src/lib/context-menu/floating-menu.svelte
mcous: It seems like Svelte is moving away from `on:...` handlers and [towards function props](https://youtu.be/pTgIx-ucMsY?t=14095), so that was my only thought. It also feels slightly invalid omit an event handler now that `isOpen` is fully controlled

ehhong: sounds good to me! thanks for the reference

packages/core/src/lib/context-menu/context-menu.svelte
ehhong: have faith in your spelling abilities!

ehhong (approved):
tested manually, all lgtm!🕴️🛟

PR #438, opened 2023-12-01
Design maintenance nov 30 2023

mrloureed:
PRIME update

- deprecates `message` slot in lieu of `subtitle` slot.
- new `actionEmphasize` slot for use with Buttons.
- margin adjustments for `action` vs `actionEmphasize` slots
- storybook docs improvements

_APP changes required after update. Replace `message` attributes with `subtitle`_

### DIFF
--- a/packages/core/src/lib/banner/__tests__/banner.spec.svelte
+++ b/packages/core/src/lib/banner/__tests__/banner.spec.svelte
@@ -17,6 +17,6 @@ export let extraClasses: cx.Argument = '';
 >
   <svelte:fragment slot="title">This is the title.</svelte:fragment>
   <svelte:fragment slot="subtitle">This is the subtitle.</svelte:fragment>
-  <svelte:fragment slot="message">This is the message.</svelte:fragment>
   <svelte:fragment slot="action">This is the action.</svelte:fragment>
+  <svelte:fragment slot="actionEmphasize">This is the button.</svelte:fragment>
 </Banner>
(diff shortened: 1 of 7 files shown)


packages/core/src/lib/banner/__tests__/banner.spec.svelte
```diff
@@ -17,6 +17,6 @@ export let extraClasses: cx.Argument = '';
 >
   <svelte:fragment slot="title">This is the title.</svelte:fragment>
   <svelte:fragment slot="subtitle">This is the subtitle.</svelte:fragment>
-  <svelte:fragment slot="message">This is the message.</svelte:fragment>
   <svelte:fragment slot="action">This is the action.</svelte:fragment>
+  <svelte:fragment slot="actionEmphasize">This is the button.</svelte:fragment>
```
micheal-parks, line 21: are we doing `actionEmphasize` case or `action-emphasize` case with other slots? just curious if this one is different

packages/core/src/lib/banner/banner.svelte
```diff
@@ -94,7 +94,7 @@ $: {
     />
   </div>
 
-  <div class="relative flex w-full justify-between gap-2 p-3">
+  <div class="relative flex w-full justify-between gap-2 p-3 pr-10">
```
micheal-parks, line 97: `pr-10` is a lot, this is intentional?

micheal-parks (approved):

PR #439, opened 2023-12-04
add db icon

sbal13:
Any idea why this icon got removed in the first place?


anjinai:
> Any idea why this icon got removed in the first place?

I don't remember removing it, but I think I was the last person to mess with prime icons.  Apologies, I may have accidentally fat fingered/deleted it.

anjinai (approved):

PR #441, opened 2023-12-05
Nav map - stop following base on mouse or wheel interaction

ethanlook:
It does the things! I didn't switch from a div to a different element and opted to silence the a11y error instead. Let me know if you agree or if I should switch to a different element.


micheal-parks (approved):
Nice! I think ignoring the warning is fine here 👍

PR #442, opened 2023-12-05
Range input

DTCurrie:
Screenshot 2023-12-05 at 3 32 29 PM


packages/core/package.json
```diff
@@ -24,8 +24,9 @@
       "types": "./dist/index.d.ts",
       "svelte": "./dist/index.js"
     },
-    "./theme": "./theme.ts",
-    "./prime.css": "./prime.css"
+    "./prime.css": "./prime.css",
+    "./plugins": "./plugins.ts",
+    "./theme": "./theme.ts"
```
DTCurrie, line 29: I didn't want to make any major changes as a part of this PR, but I am wondering if we want to combine these exports into a `tailwind.ts` file or something.

ethanlook: There are some places where we import the theme directly to grab a color in JS, for example. It feels natural to grab colors from a file called "theme", but I don't feel terribly strongly. I appreciate keeping this PR more straightforward.

packages/core/src/lib/input/__tests__/range-input.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+   * set the value of range inputs using a change call. Range and numeric inputs do not support
+   * changing values with arrow keys, which would be the ideal solution for testing these. Once
+   * those are supported, we should be able to uncomment these tests.
+   *
+   * See:
+   * https://github.com/testing-library/user-event/issues/1066
+   * https://github.com/testing-library/user-event/issues/1067
+   *
+   *
+   * it('It should slide at increments of step', async () => {
```
ethanlook, line 103: nit: `it.skip` instead of commenting out

packages/core/src/lib/input/range-input.svelte
```diff
@@ -0,0 +1,214 @@
...
+export let max = 100;
+
+/** The amount to add/subtract when moving the slider. */
+export let step = 1;
+
+/** The value of the input, defaults to the value of `min` */
+export let value: number = min;
+
+/** Whether or not the input should be rendered as readonly and be operable. */
+export let readonly = false as boolean | undefined;
```
ethanlook, line 33: Why cast to undefined? Could we not just use the false default? (for `disabled` as well)

packages/core/src/lib/input/range-input.svelte
```diff
@@ -0,0 +1,214 @@
...
+
+  // Adding 10px to center on thumb (width of 20px)
+  const left = ((value - min) / (max - min)) * datalistWidth + 10;
+  indicatorPosition.set(left);
+};
+
+/*
+ * TODO (DTCurrie): The below is valid syntax in svelte to invoke a function on a variable change, we should
+ * update our linters to allow this in svelte files, unless we hate it then we can update `positionIndicator`
+ * to accept a `currentValue` argument and pass `value` to it here. That will trigger whenever `value` changes.
```
ethanlook, line 117: I prefer passing a `currentValue` argument, I think it's easier to understand on first read.

ethanlook (approved):

packages/core/src/lib/input/range-input.svelte
DTCurrie: I was following what we do in the `Input` component. I am guessing there was some linting issues previously, I removed the cast and it is working fine. Going to go ahead and clean that up elsewhere.

packages/core/package.json
DTCurrie: I don't feel super strongly either, but in order to use PRIME you have to import both separately to include them in your config, which kind of feels weird as well. But you should only have to do that once, and you may be importing colors in multiple places, so keeping them separate is fine too. We could also make a `tailwind.ts` file that exports both `plugins` and `themes` and preserve the separate exports as well.

DTCurrie:
Noticed an issue where the numeric input can exceed the max value of the range. Going to update to better constrain the numeric input to the range value.

PR #444, opened 2023-12-06
Upgrade three-related dependencies

micheal-parks:
Doing this to prepare to sync all three dependencies among app, prime, and RDK for the new frame tab.


ethanlook (approved):
🚀

PR #447, opened 2023-12-09
APP-2198: Config page: add component creation success toast

Oabraham1:
[Ticket](https://viam.atlassian.net/browse/APP-2198)

The goal of this PR is two-fold: Implement a `Toast` and display that `Toast` on successful component creation.

## Screenshot
Screenshot 2023-12-15 at 14 48 05


packages/core/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "cSpell.words": ["exitable"]
```
DTCurrie, line 2: Did we intend to commit this?

DTCurrie:
Can you include screenshots of what this component looks like?

packages/core/src/lib/toast/context.ts
```diff
@@ -0,0 +1,141 @@
...
+
+const ToastContextKey = Symbol('toast-context');
+const ToastDuration = 4000;
+
+/** Create the internal toast state and context object. */
+export const createToastContext = (): ToastContext => {
+  const toasts = writable<ToastElement[]>([]);
+  const pageIsVisible = writable(true);
+  const pausedID = writable<string | undefined>();
+  const unsubscribeByID = new Map<string, () => void>();
```
DTCurrie, line 57: Not an issue, but why a `Map` over a `Record`?

packages/core/src/lib/toast/context.ts
```diff
@@ -0,0 +1,141 @@
...
+        resume,
+        dismiss,
+      };
+
+      unsubscribeByID.set(id, unsubscribe);
+      toasts.set([...get(toasts), toast]);
+    };
+
+  const remove = (id: string) => {
+    unsubscribeByID.get(id)?.();
```
DTCurrie, line 93: Why are we doing this `get` first?

ethanlook:
> Can you include screenshots of what this component looks like?

+1 for [APP-2847](https://viam.atlassian.net/browse/APP-2847)

packages/core/src/lib/toast/context.ts
```diff
@@ -0,0 +1,141 @@
...
+  state: ToastState;
+  toast: Toast;
+}
+
+/** Notification sender. */
+export interface Toast {
+  /** Push a neutral toast with a message. */
+  neutral: (message: string, closeable?: boolean) => void;
+  /** Push an upload toast with a message. */
+  upload: (message: string, closeable?: boolean) => void;
```
DTCurrie, line 26: I feel like `upload` is a pretty specific use case for this, should it be something like `info` or `progress`? As a user, I have no idea what an `upload` toast is supposed to look like or signify.

packages/core/src/lib/toast/toast.svelte
```diff
@@ -0,0 +1,118 @@
...
+export { extraClasses as cx };
+
+const dispatch = createEventDispatcher<{
+  /** when the toast close button is pressed*/
+  close: null;
+}>();
+
+const handleClose = () => dispatch('close');
+
+// $: isUploadingFile = variant === ToastVariant.UploadingFile;
```
DTCurrie, line 35: ✂️

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -0,0 +1,119 @@
...
+          variant="ghost"
+          cx="text-gray-7 absolute right-1"
+          aria-label="Dismiss toast"
+          on:click={handleClose}
+        >
+          <Icon name="close" />
+        </Button>
+      {/if}
+
+      {#if action}
```
DTCurrie, line 93: This kind of feels like it should be a `slot`, maybe wrapped in a layout element with the `"absolute right-1"` classes. Especially if all we are doing is having it be an event dispatcher, might as well just let users declare exactly what they need. Will make it more flexible for cases where we may just want to use an icon or something in the future as well.

packages/core/src/lib/toast/context.ts
ethanlook: I pulled down the PR and ran the storybook and it looks like this is intended to be the "process" variant we have in the [figma](https://www.figma.com/file/ERHSYDGTXiEQOVGY8ORg6I/PRIME?type=design&node-id=2650-4838&mode=design&t=hht9SlFQ2x7ltRIa-4).

If this is supposed to be the "process" variant, a few notes:
- Progress bar should go left-to-right
- Process variant should end looking like "success"
- Progress bar should have controlled progress such that it reflects the progress of the long-running action (such as uploading a file)

If this is supposed to be the "info" variant:
- Icon should be the info icon
- There should not be a progress bar

Which [variants](https://www.figma.com/file/ERHSYDGTXiEQOVGY8ORg6I/PRIME?type=design&node-id=2650-4838&mode=design&t=hht9SlFQ2x7ltRIa-4) do we need for R2D2?

packages/core/.vscode/settings.json
DTCurrie: But `exitable` is not a word, and I can't see where we are using it.

packages/core/src/lib/toast/context.ts
DTCurrie: `process` definitely makes sense as a variant name. I am wondering if we even need the process variant right now? The ticket this is attached to is specifically for the success toast, so seems like we could hold off on the `process` variant right now?

packages/core/src/lib/toast/context.ts
mcous: Since we're dynamically adding and removing these handlers, I think a `Map` is better than a plain-object record. We have lint rules that will yell at you for using `delete obj[key]`

packages/core/src/lib/toast/context.ts
mcous: To get and call the unsubscribe function, if needed, before deleting the unsubscribe function

packages/core/src/lib/toast/context.ts
```diff
@@ -0,0 +1,141 @@
...
+/** Internal toast context. */
+export interface ToastContext {
+  state: ToastState;
+  toast: Toast;
+}
+
+/** Notification sender. */
+export interface Toast {
+  /** Push a neutral toast with a message. */
+  neutral: (message: string, closeable?: boolean) => void;
```
mcous, line 24: Same question as below with `upload/process`. Our immediate needs with toasts are all success toasts. Let's not add anything in this PR that we aren't going to use immediately

packages/core/src/lib/toast/context.ts
```diff
@@ -0,0 +1,141 @@
...
+export interface ToastState {
+  toasts: Readable<ToastElement[]>;
+  pageIsVisible: Writable<boolean>;
+}
+
+/** Internal toast item state and actions. */
+export interface ToastElement {
+  id: string;
+  message: string;
+  closeable: boolean | undefined;
```
mcous, line 41: Same comment as elsewhere: all toasts that we care about right now are closable

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -0,0 +1,119 @@
...
+import { ToastVariant, type ToastVariantType } from './variants';
+
+/** The type of toast banner to be displayed to the user */
+export let variant: ToastVariantType;
+
+/** Whether there is a close icon displayed */
+export let closeable: boolean | undefined = false;
+
+/** Whether there is an action button displayed */
+export let action = '';
```
mcous, line 16: Avoid using empty strings as falsey values; use `undefined` to declare that something might not be specified. It's a lot easier to statically analyze

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -0,0 +1,119 @@
...
+const dispatch = createEventDispatcher<{
+  /** when the toast close button is pressed*/
+  close: null;
+  /**  when the action button is pressed*/
+  action: null;
+}>();
+
+const handleClose = () => dispatch('close');
+
+const handleAction = () => dispatch('action');
```
mcous, line 34: I know this was copied over from the the notification code, but can we use function props rather than events? Svelte itself is moving away from events, and in my experience, events are annoying to test for and statically analyze

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -0,0 +1,119 @@
...
+export let variant: ToastVariantType;
+
+/** Whether there is a close icon displayed */
+export let closeable: boolean | undefined = false;
+
+/** Whether there is an action button displayed */
+export let action = '';
+
+/** The scaling applied on the x axis of the page*/
+export let progress = 1;
```
mcous, line 19: Without a `process` variant, do we need this prop?

packages/storybook/src/stories/use-toast.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+```ts
+import { useToast } from '@viamrobotics/prime-core';
+
+const doSomethingCrazy = () => {
+  const toast = useToast();
+
+  toast.neutral('Plain slice please!');
+  toast.upload('Uploading file!');
+  toast.success('Fuhgeddaboudit!');
```
mcous, line 59: I know this comes from the `notify` usage, but I think it's a great time to reimagine this. For me, the level seems better as an options argument rather than having separate methods for each level

packages/core/src/lib/toast/toast-item.svelte
```diff
@@ -0,0 +1,23 @@
...
+export let variant: ToastVariantType;
+export let progress: Readable<number>;
+export let dismiss: () => unknown;
+</script>
+
+<ToastBanner
+  {closeable}
+  cx="pointer-events-auto relative mb-2 w-screen sm:w-[360px]"
+  on:close={dismiss}
+  on:action={dismiss}
```
mcous, line 16: We _do_ need actions to work for R2D2, so this looks like an issue

packages/core/src/lib/toast/context.ts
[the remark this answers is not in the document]
mcous: Agree, I think we need an options object with the following fields:

- `message: string`
- `action: { text: string, handler: () => unknown }`
- `variant: ToastVariant`

As mentioned in other comments, `closable` should not be an option; the closable state of a toast is determined by its variant

packages/core/src/lib/toast/toast-item.svelte
mcous, line 1: Is this component needed? Can it be subsumed into `<ToastContainer>`? It doesn't have any specific tests and it doesn't do anything meaningful other than apply CSS

packages/core/.vscode/settings.json
mcous, line 1: Setting aside the conversation below, this repository is designed to be opened in your editor from the root, so any `.vscode` changes should be committed at `<root>/.vscode` rather than `<root>/packages/[package]/.vscode`. This directory should be (re)moved

mcous (changes requested):

packages/core/src/lib/toast/context.ts
```diff
@@ -0,0 +1,141 @@
...
+
+/** Get access to the toast notifier in a component. */
+export const useToast = (): Toast => {
+  return useToastContext().toast;
+};
+
+/** Get access to the internal toast state. */
+export const useToastState = (): ToastState => {
+  return useToastContext().state;
+};
```
DTCurrie, line 141: Open question for reviewers, do we prefer full syntax here, or just one liners like so?

```suggestion
/** Get access to the toast notifier in a component. */
export const useToast = (): Toast => useToastContext().toast;

/** Get access to the internal toast state. */
export const useToastState = (): ToastState => useToastContext().state;
```

packages/storybook/src/stories/use-toast.mdx
DTCurrie: So update the API to something like this?

```
  toast('Plain slice please!');
  toast('Uploading file!', ToastVariant.Upload);
  toast('Fuhgeddaboudit!', ToastVariant.Success);
```

packages/core/src/lib/toast/context.ts
ethanlook: I think I prefer

```
const { state, toast } = useToast();
```

mcous: One wrinkle: toast state is internal

packages/core/src/lib/toast/toast.svelte
```diff
@@ -0,0 +1,118 @@
...
+          </div>
+        {/if}
+        <span>{message}</span>
+      </div>
+      <Button variant="ghost">Cancel</Button>
+    </div>
+    <div class="h-[calc(100%+2px)] w-full bg-medium">
+      <div
+        style="transform: scale(1, {progress})"
+        class="-mr-py -mb-py h-full w-[3px] origin-bottom-left self-stretch bg-info-dark pr-[20px]"
```
mrloureed, line 91: I don't understand `-mr-py -mb-py`

packages/core/src/lib/toast/context.ts
Oabraham1: Will focus this PR on the success variant.

packages/core/src/lib/toast/toast.svelte
Oabraham1: Removing this file from the PR as it was added in error.

packages/core/src/lib/toast/toast-item.svelte
```diff
@@ -0,0 +1,23 @@
...
+export let message: string;
+export let closeable: boolean | undefined = false;
+export let variant: ToastVariantType;
+export let progress: Readable<number>;
+export let dismiss: () => unknown;
+</script>
+
+<ToastBanner
+  {closeable}
+  cx="pointer-events-auto relative mb-2 w-screen sm:w-[360px]"
```
mrloureed, line 14: You probably don't need the `pointer-events-auto`

Nice use of `sm:` utility

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -0,0 +1,46 @@
...
+
+$: pageIsVisible.set(visibilityState === 'visible');
+</script>
+
+<svelte:document bind:visibilityState />
+
+<div
+  role="alert"
+  aria-label="Toasts"
+  class="pointer-events-none fixed bottom-auto left-1/2 top-auto z-[9999] m-0 -translate-x-1/2 transform p-0"
```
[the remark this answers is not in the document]
Oabraham1: I actually was going to remove both since we do not want the progress of the toast to be paused when a user hovers over it right?

packages/core/src/lib/toast/toast-item.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
+import type { Readable } from 'svelte/store';
+import { ToastBanner, type ToastVariantType } from '$lib/toast';
+
+export let message: string;
+export let closeable: boolean | undefined = false;
```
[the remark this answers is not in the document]
Oabraham1: Yes, I think I was using this as a way whether to determine whether or not a close icon button is displayed, but I'm taking it out

packages/storybook/src/stories/use-toast.mdx
mcous: Yeah, though I'd use an options object, as Ethan suggested in his review. So something like:

```ts
toast({ message: 'Hello world', variant: ToastVariant.SUCCESS })
```

From there, if we wanted, we could layer stuff in like "default to `INFO`" and "allow a plain string to toast an info message"

packages/core/src/lib/toast/toast.svelte
mcous: From what I can tell, they're not tailwind classes. The fact that they're shunted to the front is a warning sign; that's where prettier puts non-tailwind classes

(Fun fact: our linter isn't capable of warning us about non-tailwind classes because that part of eslint-plugin-tailwindcss [isn't fully compatible with eslint-plugin-svelte](https://github.com/francoismassart/eslint-plugin-tailwindcss/issues/214). Yay.)

packages/core/src/lib/toast/toast-item.svelte
Oabraham1: I will modify this. Question: does the success toast we're using use actions in any way? From my understanding, only the close button is displayed on a success toast. I'm asking because I might make the action field optional

mcous: Yes, various success toasts for destructive actions include an "Undo" action

packages/core/src/lib/toast/toast-banner.svelte
Oabraham1: Removing it

packages/core/src/lib/toast/__tests__/toast-banner.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+describe('Toast', () => {
+  it('Renders toast element with appropriate message, and action text', () => {
+    const { container } = render(Toast, {
+      variant: ToastVariant.Success,
+    });
+
+    expect(screen.getByText('This is the message.')).toBeVisible();
+    expect(screen.getByText('This is the action button.')).toBeVisible();
+    expect(container.querySelector('.text-success-dark')).toBeVisible();
+    expect(screen.getByRole('button')).toBeVisible();
```
mcous, line 17: - Use data-testid for testing slots, as described above
- Try to always separate your element queries from you expectations for readability
- Try to always use names when you're querying buttons, especially when they're icon buttons, which need extra a11y care
- Almost never use `container` and almost never use `querySelector`
    - Needing to resort to these is usually a sign that your HTML isn't properly accessible and/or semantic
    - In this case, I would say that the `Icon` needs to be "visible" to the a11y tree as an image with a label of "success," because it conveys meaning for the toast message that isn't available elsewhere

```suggestion
    const message = screen.getByTestId('message')
    const action = screen.getByTestId('action')
    const icon = screen.getByRole('img', { name: /success/iu })
    const dismissButton = screen.getByRole('button', { name: /dismiss/iu })
    
    expect(message).toBeVisible();
    expect(action).toBeVisible();
    expect(icon).toHaveClass('text-success-dark');
    expect(dismissButton).toBeVisible();
```

packages/core/src/lib/toast/__tests__/toast-banner.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it('Calls the dismiss function when the close button is clicked', async () => {
+    const dismissMock = vi.fn();
```
mcous, line 21: [Nit] The "mock" in the name here is a bit redundant; it's defined with `vi.fn()`, we know what it is. `dismiss` is a little easier to read/use

packages/core/src/lib/toast/__tests__/toast-banner.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    render(Toast, {
+      variant: 'success',
+      dismiss: dismissMock,
+    });
+
+    await fireEvent.click(screen.getByLabelText(/Dismiss toast/iu));
```
mcous, line 26: - Separate your queries from user actions
- `getByRole` is almost always preferred if you can use it

```suggestion
    const dismissButton = screen.getByRole('button', { name: /dismiss/iu });
    
    await fireEvent.click(dismissButton);
```

packages/core/src/lib/toast/__tests__/toast-banner.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it('Renders with the passed cx classes', () => {
+    const { container } = render(Toast, {
+      variant: 'success',
+      extraClasses: cxTestArguments,
+    });
+
+    expect(container.querySelector('.bg-medium')).toHaveClass(cxTestResults);
+  });
```
mcous, line 37: This test smells due to the `query-selector`, and what it's testing is a bit of a visual implementation detail. Unless there's a good semantic reason to test a classname (e.g. something representing success has a class to apply our "text-success-dark", something needs to be hidden) I'd recommend avoiding these sorts of tests, _especially_ for internal components like `<ToastBanner />`

packages/core/src/lib/toast/__tests__/toast-banner.spec.ts
mcous, line 1: Some of the test coverage in this file looks redundant with `toast.spec.ts`. Since `toast.spec.ts` tests the actual public interface of the `toast` lib, that's where I'd recommend putting all your tests

These sort of internal unit test _can_ be useful, but constraining yourself to public interfaces creates less coupling between your tests and your internal implementation details, which is almost always preferable if you can manage it

packages/core/src/lib/toast/toast-item.svelte
mcous: Bumping this comment. It looks like splitting apart `ToastBanner` and `ToastItem` is creating a more complicated internal structure for this functionality than is needed

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -0,0 +1,60 @@
...
+let icon: IconName | null = null;
+let iconClasses = '';
+
+$: {
+  // TODO: use a `switch` statement when extending the variants
+  if (variant === ToastVariant.Success) {
+    icon = 'check-circle';
+    iconClasses = 'text-success-dark';
+  }
+}
```
mcous, line 26: This logic looks like it could live in `variants.ts` and be a little easier to read

packages/core/src/lib/toast/toast-item.svelte
```diff
@@ -0,0 +1,28 @@
...
+  {variant}
+  cx="pointer-events-auto relative"
+  on:close={dismiss}
+  {dismiss}
+>
+  <svelte:fragment slot="message">
+    {message}
+  </svelte:fragment>
+  <svelte:fragment slot="action">
+    {#if action}
```
mcous, line 22: This `if` looks too low, this usage of `slot="action"` is going to trigger `if $$slots.action` in `<ToastBanner>`. I think this mostly relates to me comment above about there being one two many components in play in this internal structure

packages/core/src/lib/toast/toast-item.svelte
```diff
@@ -0,0 +1,28 @@
...
+export let message: string;
+export let action: { text: string; handler: () => unknown } | undefined =
+  undefined;
+export let dismiss: () => unknown;
+</script>
+
+<ToastBanner
+  {variant}
+  cx="pointer-events-auto relative"
+  on:close={dismiss}
```
mcous, line 15: Prefer function prop to event emitter

packages/core/src/lib/toast/context.ts
```diff
@@ -0,0 +1,132 @@
...
+
+/** Get access to the toast notifier in a component. */
+
+export const useToast = (): ((params: ToastParams) => void) => {
+  const context = useToastContext();
+
+  return (params: ToastParams) => {
+    context.toast(params);
+  };
+};
```
mcous, line 127:
```suggestion
export const useToast = (): ((params: ToastParams) => void) => {
  return useToastContext().toast;
};
```

packages/core/src/lib/toast/__tests__/toast-banner.spec.ts
Oabraham1: I'll remove the test case :)

packages/core/src/lib/toast/toast-item.svelte
Oabraham1: I can move everything in this file to <ToastBanner>

Oabraham1:
@mcous please let me know if I missed any of your comments, but I think I addressed everything.

packages/core/src/lib/toast/__tests__/toast.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+        },
+      });
+    });
+
+    const status = screen.getByRole('status');
+    const toast = within(status).getByRole('listitem');
+    const actionButton = screen.getByRole('button', {
+      name: /perform action/iu,
+    });
+    await fireEvent.click(actionButton);
```
mcous, line 62: Prefer `@testing-library/user-event` to `fireEvent`

```ts
import userEvent from '@testing-library/user-event'

// sometime before render
const user = userEvent.setup()

// click something
await user.click(element)
```

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -0,0 +1,71 @@
...
+</script>
+
+<div
+  class={cx(
+    'pointer-events-auto relative flex w-auto items-center gap-2 border bg-medium p-3 text-sm',
```
mcous, line 36: [Nit] Is `pointer-events-auto` doing anything here?

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -0,0 +1,47 @@
...
+>
+  <ul class="flex flex-col items-center gap-2">
+    {#each $toasts as { id, pause, resume, ...toast } (id)}
+      <li
+        in:fly={{ y: 256 }}
+        out:fade
+        animate:flip={{ duration: 200 }}
+        on:mouseenter={pause}
+        on:mouseleave={resume}
+        class="inline-block"
```
mcous, line 41: [Nit] I'm not sure `inline-block` does anything when the parent is `flex`

packages/core/src/lib/toast/variants.ts
```diff
@@ -0,0 +1,7 @@
+export const ToastVariant = {
+  Neutral: 'neutral',
+  Upload: 'upload',
```
mcous, line 3: [Nit]: let's remove these while we're not using them

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -0,0 +1,47 @@
...
+
+$: pageIsVisible.set(visibilityState === 'visible');
+</script>
+
+<svelte:document bind:visibilityState />
+
+<div
+  role="status"
+  aria-label="Toasts"
+  class="pointer-events-none fixed bottom-0 left-1/2 top-auto z-[9999] -translate-x-1/2 transform"
```
mcous, line 31: We're missing an `overflow-hidden` on this element; in the storybook, a scrollbar appears while toasts are flying in if you're on desktop with a mouse (on macOS, it's likely worse on Windows)

[Nit] I think `z-[9999]` is equivalent to our `z-max`

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -0,0 +1,47 @@
...
+>
+  <ul class="flex flex-col items-center gap-2">
+    {#each $toasts as { id, pause, resume, ...toast } (id)}
+      <li
+        in:fly={{ y: 256 }}
+        out:fade
```
mcous, line 37: Is it just me, or are the elements not actually fading out? On my machine they don't fade until I remove the `inline-block` class below

mcous (commented):
Almost there! Got a couple small issues with the CSS / animations, but we'll be good to merge once they are resolved

packages/core/src/lib/toast/toast-banner.svelte
Oabraham1: Removed it since the one from `toast-container.svelte` is used to pause the toast progress on hover

packages/core/src/lib/toast/toast-container.svelte
Oabraham1: I removed it.

mcous (approved):

PR #448, opened 2023-12-12
upload icon

sbal13:
Not sure what the procedure is for adding icons, but do we need to update the package version?


ethanlook (approved):
Yup, this is perfect, version bump and all! You even kept it alphabetized 🙏🏼

PR #449, opened 2023-12-13
APP-3124: Add hover delay for tooltips

ethanlook:
Per the [designs for APP-3124](https://www.figma.com/file/GSi2nRrsKHQnP2uYNx5XMK/Make-UI-Not-Shitty-II?type=design&node-id=3100-16026&mode=design&t=MNU0aepvDubaCdJ1-11), some tooltips should only show after 1s of hovering.

## Change log

- Add `hoverDelayMS` prop to `Tooltip` and `TooltipContainer`
- Add `hoverDelayMS` writable to tooltip context, and update `isVisible` derived store to set after timeout
- Add test
- Add example to dev environment
- Add example to storybook

tooltip-hover-delay

### DIFF
--- a/packages/core/src/lib/tooltip/__tests__/tooltip.spec.svelte
+++ b/packages/core/src/lib/tooltip/__tests__/tooltip.spec.svelte
@@ -3,7 +3,7 @@ import { Tooltip, type FloatingPlacement, type TooltipVisibility } from '$lib';
 
 export let location: FloatingPlacement = 'top';
 export let state: TooltipVisibility | undefined = undefined;
-export let hoverDelayMS: number | undefined = undefined;
+export let hoverDelayMS = 0;
 </script>
 
 <Tooltip
--- a/packages/core/src/lib/tooltip/__tests__/tooltip.spec.svelte
+++ b/packages/core/src/lib/tooltip/__tests__/tooltip.spec.svelte
@@ -3,11 +3,13 @@ import { Tooltip, type FloatingPlacement, type TooltipVisibility } from '$lib';
 
 export let location: FloatingPlacement = 'top';
 export let state: TooltipVisibility | undefined = undefined;
+export let hoverDelayMs: number | undefined = undefined;
 </script>
 
 <Tooltip
   {location}
   {state}
+  {hoverDelayMs}
   let:tooltipID
 >
   <button
--- a/packages/core/src/lib/tooltip/__tests__/tooltip.spec.svelte
+++ b/packages/core/src/lib/tooltip/__tests__/tooltip.spec.svelte
@@ -3,13 +3,13 @@ import { Tooltip, type FloatingPlacement, type TooltipVisibility } from '$lib';
 
 export let location: FloatingPlacement = 'top';
 export let state: TooltipVisibility | undefined = undefined;
-export let hoverDelayMs: number | undefined = undefined;
+export let hoverDelayMS: number | undefined = undefined;
 </script>
 
 <Tooltip
   {location}
   {state}
-  {hoverDelayMs}
+  {hoverDelayMS}
   let:tooltipID
 >
   <button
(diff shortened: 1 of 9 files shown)


packages/core/src/lib/tooltip/tooltip-styles.ts
```diff
@@ -73,12 +76,28 @@ const createContext = (): TooltipContext => {
   );
   const style = floatingStyle({ offset: 7, shift: 5, flip: true, auto: true });
 
+  const { set: setHoveredTimeout, clear: clearHoveredTimeout } = useTimeout();
```
mcous, line 79: When I was building this out, I had imagined any delays happening as part of the `isVisible` derived store. `isHovered`, IMO, should track the actual "is this item hovered?" state, while a delay should be incorporated into the logic of "knowing this is hovered, should it be visible?".

If `hoverDelayMS` becomes a store, this also gives us slightly better reactivity / cleanup if the delay value changes, which may be a thing we want to do in the future (e.g. for a group of icons, once the tooltip displays for one of them, we should not use a delay for other icons in the group).

```ts
const hoverDelayMS = writable(0);  
const isVisible = derived(
  [isHovered, visibility, hoverDelayMS],
  ([$isHovered, $visibility, $hoverDelayMS], set) => {
    let cleanup = noop;

    if ($visibility === 'visible') {
      set(true);
    } else if ($visibility === 'invisible' || !$isHovered) {
      set(false);
    } else {
      const timeoutID = setTimeout(() => set(true), $hoverDelayMS);
      cleanup = () => clearTimeout(timeoutID);
    }

    return cleanup;
  },
  false
);
```

packages/core/src/lib/tooltip/__tests__/tooltip.spec.svelte
```diff
@@ -3,11 +3,13 @@ import { Tooltip, type FloatingPlacement, type TooltipVisibility } from '$lib';
 
 export let location: FloatingPlacement = 'top';
 export let state: TooltipVisibility | undefined = undefined;
+export let hoverDelayMs: number | undefined = undefined;
```
mcous, line 6: [Nit] Not sure how much we want to stick to our acronym casing convention, here, but maybe `hoverDelayMS`

packages/core/src/lib/tooltip/tooltip-container.svelte
```diff
@@ -34,7 +34,11 @@
 <script lang="ts">
 import { provideTooltipContext } from './tooltip-styles';
 
-const { id } = provideTooltipContext();
+export let hoverDelayMs: number | undefined = undefined;
```
mcous, line 37: Given that tooltip style calculation is already async, I'm not sure we lose out on anything if we default to `0` rather than `undefined` and accept that we'll sometimes to `setTimeout(..., 0)`.

Alternatively, should we be setting a default non-zero timeout on this? I think that might be the design intent across the board

mcous (commented):
Very cool! Some thoughts that were in my head about this functionality but I didn't write down anywhere

packages/core/src/lib/tooltip/tooltip-container.svelte
ethanlook: Will set the default to `0` rather than `undefined` - I spoke with Nat and a delay is only sometimes used. For the most part, tooltips should be instant.

packages/core/src/lib/tooltip/tooltip-styles.ts
ethanlook: Thanks for the suggestion on this - much cleaner and better reactivity.

packages/core/src/lib/tooltip/tooltip-styles.ts
```diff
@@ -7,6 +7,7 @@ import {
   type FloatingPlacement,
 } from '$lib/floating';
 import { uniqueId } from '$lib/unique-id';
+import { noop } from 'lodash';
```
mcous, line 10: Since we don't use `lodash-es`:

```suggestion
import noop from 'lodash/noop';
```

mcous (approved):

PR #451, opened 2023-12-15
Update core and blocks to SvelteKit v2

mcous:
## Overview

Excluding lockfile:

```
28 files changed, 163 insertions(+), 162 deletions(-)
```

I was working on updating SvelteKit in the downstream app, and immediately hit a couple problems from Prime. I decided to do the update here first to shake everything out.

SvelteKit v2 updates to Vite v5, which is much more strict about CJS vs ESM. We had two imports in Prime that pretended CJS deps were ESM, which Vite v5 refused to bundle:

- PrismJS
- Lodash

We can no longer implicitly destructure an object exported as a CJS default export:

```ts
import { highlightElement, plugins } from 'prismjs';
```

Instead, we must be explicit:

```ts
import Prism from 'prismjs'
const { highlightElement, plugins } = Prism
```

This is good, because it stops hiding imports where we thought we were getting tree-shaking but weren't. For Prism, I did the change above. For lodash, I swapped it out for lodash-es, which was on my todo list anyway.

## Change log

- Update all dependencies, excect:
  - Didn't touch legacy at all
  - Blocks
      - mapbox
  - Storybook
      - sveltekit (SB hasn't caught up to Vite v5 yet)
- Swap lodash for lodash-es
- Make minimal fixes to get builds / checks / dev environments working

## Review requests

Before merging, I'll be `pnpm link`'ing these to app to make sure things keep functioning.

- Try out the dev sites if you can
- Review source code changes with paranoia


DTCurrie (approved):
I was real skeptical about that PrismJs import.

.github/workflows/ci.yml
```diff
@@ -55,6 +55,9 @@ jobs:
       - name: Install 🔧
         uses: viamrobotics/js-config/.github/actions/setup@0006127efb3f62669f662ca31c555a49a034935e
 
+      - name: Build 🏗️
+        run: pnpm run build
+
```
ethanlook, line 60: Should we just combine this with the above?

ethanlook (approved):
Why did tailwind reorder?

.github/workflows/ci.yml
mcous: I'm here for it

mcous:
> Why did tailwind reorder?

The new order is definitely correct, so I'm not sure why it was improperly ordered in the first place. Since the end result is correct I didn't really look that deeply into it

mcous:
Synced with main, updated stragglers, and ran a test with `pnpm link` for my local copy of the app. Did not experience any issues in either Vue or Svelte, so going to merge this as soon as CI is green

PR #454, opened 2023-12-19
APP-2198-FIX: Added padding and refactored function

Oabraham1:
In this PR, I made the `provideToast` function easier to use and added some padding to the toast.

## Screenshot
Screenshot 2023-12-19 at 14 14 48


packages/core/src/lib/toast/__tests__/toast.spec.ts
```diff
@@ -42,7 +41,6 @@ describe('toast', () => {
     await act(() => {
       context.toast({
         message: 'This is a success toast message',
-        variant: ToastVariant.Success,
```
mcous, line 45: I think it's worth keeping this in the test, because otherwise you don't have a test that uses the explicit `variant` option. It will also fix your lint issue

mcous (approved):

PR #457, opened 2023-12-21
APP-2198: Addressing stylistic issues in toast

Oabraham1:
In this PR, I made the following changes to the toast ui element:

- Increased the bottom margin of the toast to `24px`
- Fixed the fade out effect by adding `delay` and `duration`
- Fixed the internal padding of the toast to match the figma designs.

## Visuals

https://github.com/viamrobotics/prime/assets/48206838/4ee2b2f9-8413-4eb9-8399-8946e45416f0

### DIFF
--- a/packages/core/src/lib/toast/toast-banner.svelte
+++ b/packages/core/src/lib/toast/toast-banner.svelte
@@ -2,7 +2,7 @@
 
 <script lang="ts">
 import cx from 'classnames';
-import { Button, Icon } from '$lib';
+import { Icon, IconButton } from '$lib';
 import { iconName, iconClasses } from './variants';
 
 /** The message displayed on the toast */
@@ -10,7 +10,7 @@ export let message: string;
 /** Function called when the close button is clicked */
 export let dismiss: () => void;
 /** The action performed on the button */
-export let action: { text: string; handler: () => void } | undefined =
+export let action: { text: string; handler: () => unknown } | undefined =
   undefined;
 
 /** Additional CSS classes to pass to the banner. */
@@ -37,20 +37,21 @@ export { extraClasses as cx };
     {message}
   </div>
 
-  {#if action}
-    <button
-      on:click={action.handler}
-      aria-label="Perform action"
-    >
-      <span class="text-sm font-medium hover:underline">{action.text}</span>
-    </button>
-  {/if}
-  <Button
-    variant="ghost"
-    cx="text-gray-7"
-    aria-label="Dismiss toast"
-    on:click={dismiss}
-  >
-    <Icon name="close" />
-  </Button>
+  <div class="flex gap-1">
+    {#if action}
+      <button
+        type="button"
+        on:click={action.handler}
+        aria-label="Perform action"
+      >
+        <span class="text-sm font-medium hover:underline">{action.text}</span>
+      </button>
+    {/if}
+    <IconButton
+      cx="text-gray-7"
+      label="Dismiss toast"
+      icon="close"
+      on:click={dismiss}
+    />
+  </div>
 </div>
--- a/packages/core/src/lib/toast/toast-banner.svelte
+++ b/packages/core/src/lib/toast/toast-banner.svelte
@@ -20,7 +20,7 @@ export { extraClasses as cx };
 
 <div
   class={cx(
-    'relative flex h-10 w-auto max-w-[480px] items-center border bg-medium pl-3 pr-1 text-sm',
+    'relative flex h-10 w-max max-w-[480px] items-center border bg-medium pl-3 pr-1 text-sm',
     extraClasses
   )}
 >
@@ -34,7 +34,7 @@ export { extraClasses as cx };
       aria-label="success"
     />
 
-    {message}
+    <p class="truncate">{message}</p>
   </div>
 
   <div class="flex gap-1">
--- a/packages/core/src/lib/toast/toast-banner.svelte
+++ b/packages/core/src/lib/toast/toast-banner.svelte
@@ -20,7 +20,7 @@ export { extraClasses as cx };
 
 <div
   class={cx(
-    'relative flex w-auto items-center gap-2 border bg-medium p-2 text-sm',
+    'relative flex w-auto items-center gap-2 border bg-medium p-1 text-sm',
     extraClasses
   )}
 >
--- a/packages/core/src/lib/toast/toast-banner.svelte
+++ b/packages/core/src/lib/toast/toast-banner.svelte
@@ -10,7 +10,7 @@ export let message: string;
 /** Function called when the close button is clicked */
 export let dismiss: () => void;
 /** The action performed on the button */
-export let action: { text: string; handler: () => unknown } | undefined =
+export let action: { text: string; handler: () => void } | undefined =
   undefined;
 
 /** Additional CSS classes to pass to the banner. */
@@ -20,11 +20,11 @@ export { extraClasses as cx };
 
 <div
   class={cx(
-    'relative flex w-auto items-center gap-2 border bg-medium p-1 text-sm',
+    'relative flex h-10 w-auto max-w-[480px] items-center border bg-medium pl-3 pr-1 text-sm',
     extraClasses
   )}
 >
-  <div class="w-[18px]">
+  <div class="mr-4 flex gap-2">
     <Icon
       size="lg"
       name={iconName}
@@ -33,9 +33,9 @@ export { extraClasses as cx };
       aria-hidden={false}
       aria-label="success"
     />
-  </div>
 
-  <div class="mr-auto">{message}</div>
+    {message}
+  </div>
 
   {#if action}
     <button
--- a/packages/core/src/lib/toast/toast-container.svelte
+++ b/packages/core/src/lib/toast/toast-container.svelte
@@ -11,10 +11,11 @@ It needs access to context from `provideToast`, which should also
 be added to the root layout component.
 -->
 <script lang="ts">
-import { fade, fly } from 'svelte/transition';
+import { fly, fade } from 'svelte/transition';
 
 import { useToastState } from './context';
 import ToastBanner from './toast-banner.svelte';
+import { flip } from 'svelte/animate';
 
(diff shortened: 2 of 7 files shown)


packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -20,7 +20,7 @@ export { extraClasses as cx };
 
 <div
   class={cx(
-    'relative flex w-auto items-center gap-2 border bg-medium p-2 text-sm',
```
mcous, line 23: Neither `gap-2` (8px gap between flex children) nor `p-1` (4px padding on all sides) is correct, according to Figma:

Screenshot 2023-12-26 at 09 05 58

- Left padding: 12px (`pl-3`)
- Right padding: 4px (`pr-1`)
- Top/bottom padding: none, use fixed height of 40px (`h-10`)

We're also missing a max-width of `480px` on this class

Screenshot 2023-12-26 at 09 10 48

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -20,7 +20,7 @@ export { extraClasses as cx };
...
+    'relative flex w-auto items-center gap-2 border bg-medium p-1 text-sm',
     extraClasses
   )}
 >
```
mcous, line 26: Can't comment on a few things below because of GH, but a few problems I see:

- Where is the `w-[18px]` wrapping the `<Icon>`coming from?
    - According to Figma, the icon width is supposed to be 20px, the default for the `lg` variant
    - The `<div>` wrapper with a width override seems unnecessary
- The Figma says the message should have a margin-right of `16px`, so `mr-4` instead of `mr-auto`

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -28,14 +27,13 @@ $: pageIsVisible.set(visibilityState === 'visible');
...
   aria-label="Toasts"
-  class="pointer-events-auto fixed bottom-0 left-1/2 top-auto z-max -translate-x-1/2 transform overflow-hidden p-1.5"
+  class="pointer-events-auto fixed bottom-0 left-1/2 top-auto z-max -translate-x-1/2 transform overflow-hidden pb-6"
 >
   <ul class="flex flex-col items-center gap-2">
     {#each $toasts as { id, pause, resume, ...toast } (id)}
       <li
         in:fly={{ y: 256 }}
-        out:fade
-        animate:flip={{ duration: 200 }}
```
mcous, line 38: Removal of `flip` breaks the gradual shift of old toasts as new ones come in; it causes them to shift instantaneously, which is not what we want

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -28,14 +27,13 @@ $: pageIsVisible.set(visibilityState === 'visible');
 <div
   role="status"
   aria-label="Toasts"
-  class="pointer-events-auto fixed bottom-0 left-1/2 top-auto z-max -translate-x-1/2 transform overflow-hidden p-1.5"
+  class="pointer-events-auto fixed bottom-0 left-1/2 top-auto z-max -translate-x-1/2 transform overflow-hidden pb-6"
 >
   <ul class="flex flex-col items-center gap-2">
```
mcous, line 32: `gap-2` (8px) is wrong here

Screenshot 2023-12-26 at 09 36 32

12px is `gap-3`

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -28,14 +27,13 @@ $: pageIsVisible.set(visibilityState === 'visible');
...
     {#each $toasts as { id, pause, resume, ...toast } (id)}
       <li
         in:fly={{ y: 256 }}
```
mcous, line 35: This value should be `64` to come in from just outside the screen:

- Bottom padding of `24px` +
- Height of element `40px`

The value of `256` comes from notifications, which are coming in on the X axis

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -28,14 +27,13 @@ $: pageIsVisible.set(visibilityState === 'visible');
 <div
   role="status"
   aria-label="Toasts"
-  class="pointer-events-auto fixed bottom-0 left-1/2 top-auto z-max -translate-x-1/2 transform overflow-hidden p-1.5"
```
mcous, line 31: - Remove `pointer-events-auto`, it doesn't do anything
- Remove `overflow-hidden`, it breaks fading when using `animate:flip`

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -28,14 +27,13 @@ $: pageIsVisible.set(visibilityState === 'visible');
...
+  class="pointer-events-auto fixed bottom-0 left-1/2 top-auto z-max -translate-x-1/2 transform overflow-hidden pb-6"
 >
   <ul class="flex flex-col items-center gap-2">
     {#each $toasts as { id, pause, resume, ...toast } (id)}
       <li
         in:fly={{ y: 256 }}
-        out:fade
-        animate:flip={{ duration: 200 }}
+        out:fade={{ duration: 400, delay: 400 }}
```
mcous, line 36: Remove these options, they're not needed

mcous (changes requested):
I think you might have jinxed yourself a little with "FINAL-FIX" in the PR name 🙃. Please go through the Figma with a fine-toothed comb. Hint: hold the "Option" key in Figma to show pixel distances between elements

Other requests:

- The `<ToastBanner>` component is receiving an unknown prop and logging a warning in the console. Do not pass the unknown prop
- Export `<ToastBanner>` from `prime-core` and give it a Storybook entry

packages/core/src/lib/toast/toast-banner.svelte
Oabraham1: The `w-[18px]` was a carry-over from another state that has been totally discarded and I forgot to remove. Removing it in my next push

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -10,7 +10,7 @@ export let message: string;
 /** Function called when the close button is clicked */
 export let dismiss: () => void;
 /** The action performed on the button */
-export let action: { text: string; handler: () => unknown } | undefined =
+export let action: { text: string; handler: () => void } | undefined =
```
mcous, line 13: [Nit] I find `unknown` to be a more useful return type in signatures for interaction handler, was there a reason for this change?

packages/core/src/lib/toast/toast-container.svelte
```diff
@@ -11,11 +11,11 @@ It needs access to context from `provideToast`, which should also
 be added to the root layout component.
 -->
 <script lang="ts">
-import { fade, fly } from 'svelte/transition';
-import { flip } from 'svelte/animate';
+import { fly, fade } from 'svelte/transition';
 
 import { useToastState } from './context';
 import ToastBanner from './toast-banner.svelte';
+import { flip } from 'svelte/animate';
```
mcous, line 18: [Nit] Why was this import statement moved?

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -33,9 +33,9 @@ export { extraClasses as cx };
...
       aria-label="success"
     />
-  </div>
 
-  <div class="mr-auto">{message}</div>
+    {message}
+  </div>
 
   {#if action}
     <button
```
mcous, line 41: Missing `type="button"` on this button, which is important because buttons without an explicit `type` can change out from under you, e.g. by implicitly becoming a `type="submit"` if inside a form

mcous (commented):
- Close button should be a square, but is instead rectangular
    - Use the `<IconButton>` component rather than an `<Icon>` inside a `<Button>`
- There is supposed to be `4px` of space between the action and the button, but there is no space currently

packages/core/src/lib/toast/toast-banner.svelte
Oabraham1: I wanted it to be more consistent with the dismiss function which has a return type of void

Oabraham1: But since it's an interaction handler for actions I can make it `unknown`

packages/core/src/lib/toast/toast-container.svelte
Oabraham1: That was unintentional

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -33,24 +33,25 @@ export { extraClasses as cx };
       aria-hidden={false}
       aria-label="success"
     />
-  </div>
 
-  <div class="mr-auto">{message}</div>
+    {message}
```
mcous, line 37: Nit: wrap this in a `<p class="truncate">` to adhere to the design guidelines that:

> Toasts can be up to 480px wide. **No text wrapping**

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -20,11 +20,11 @@ export { extraClasses as cx };
 
 <div
   class={cx(
-    'relative flex w-auto items-center gap-2 border bg-medium p-2 text-sm',
+    'relative flex h-10 w-auto max-w-[480px] items-center border bg-medium pl-3 pr-1 text-sm',
```
mcous, line 23: Use `w-max` rather than `w-auto` to set `width: max-content` on the items. This ensures they display correctly (sized to their contents) in the storybook

```suggestion
    'relative flex h-10 w-max max-w-[480px] items-center border bg-medium pl-3 pr-1 text-sm',
```

Current:
Screenshot 2024-01-09 at 12 52 08

Expected:
Screenshot 2024-01-09 at 12 52 13

mcous (approved):

mcous:
Looks like the package bump got lost in merge conflict resolution, don't forget to rebump

Oabraham1:
> Looks like the package bump got lost in merge conflict resolution, don't forget to rebump

Yup already re-bumped

PR #458, opened 2024-01-04
APP-3126: clean up various interfaces for searchable select fixes

mcous:
## Overview

While working on `<SearchableSelect>`, I ran into some fiddly bugs and organizational lint that didn't fit neatly into actually working on `<SearchableSelect>`. I threw them all into this separate PR for easier review.

### Bug fixes

- Menu toggle button in `<SelectInput>` now set to `type="button"` to avoid errant form submissions
- Scope `<SelectInput>` group modifiers more strictly to avoid weird hover/focus states when inside nested tailwind `group`s
- Prevent `<SelectInput>` button from taking focus away from the `<input>`
- Fix types of `<Floating>` props

### Features for `<SearchableSelect>`

- Allow `<SelectInput>` focus styles to be disabled so a `combobox` can maintain DOM focus while [visually focusing elsewhere](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-activedescendant)
- Allow `<Floating>` to match its parent's width

### Organizational stuff

- Consolidate exports for `$lib/icon` and `$lib/select`
- Pare back `$lib/select/search` module and tests to only export and test public interfaces
- Add a couple enum-like objects for string union props

## Review requests

There will be a follow-up PR that will be better for smoke testing, so feel free to focus on the code in this one

### DIFF
--- a/packages/core/.eslintrc.cjs
+++ b/packages/core/.eslintrc.cjs
@@ -22,6 +22,8 @@ module.exports = {
   rules: {
     // TODO(mc, 2023-12-15): remove overrides, fix issues
     '@typescript-eslint/no-non-null-assertion': 'warn',
+    // TODO(mc, 2024-01-03): move to base config?
+    'multiline-comment-style': 'off',
   },
   overrides: [
     {
@@ -35,6 +37,13 @@ module.exports = {
       files: ['.eslintrc.cjs', '.prettierrc.cjs'],
       rules: {
         '@typescript-eslint/no-unsafe-assignment': 'off',
+      },
+    },
+    // TODO(mc, 2024-01-03): remove when no-non-null-assertion errors fixed
+    {
+      files: ['**/__tests__/**'],
+      rules: {
+        '@typescript-eslint/no-non-null-assertion': 'off',
       },
     },
   ],
--- a/packages/core/src/lib/select/select-input.svelte
+++ b/packages/core/src/lib/select/select-input.svelte
@@ -1,13 +1,15 @@
 <script lang="ts">
 import cx from 'classnames';
-import type { SelectState } from './select.svelte';
-import Icon from '$lib/icon/icon.svelte';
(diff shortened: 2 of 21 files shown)


packages/core/.eslintrc.cjs
```diff
@@ -22,6 +22,8 @@ module.exports = {
   rules: {
     // TODO(mc, 2023-12-15): remove overrides, fix issues
     '@typescript-eslint/no-non-null-assertion': 'warn',
+    // TODO(mc, 2024-01-03): move to base config?
+    'multiline-comment-style': 'off',
```
mcous, line 26: What do we think about this? This lint adds a lot of friction without any real payoff, in my opinion. Sometimes I just need to write a two-line TODO comment, and this blows it up to four lines. Plus, it'd be nice to reserve `/* */` for JSDoc

packages/core/src/lib/select/select-input.svelte
```diff
@@ -58,18 +85,24 @@ $: errorClasses =
...
+    type="button"
+    class="absolute right-2 top-1/2 -translate-y-1/2 transform"
     tabindex="-1"
     aria-label="Toggle menu"
+    aria-controls={menuId}
+    aria-expanded={isOpen}
     on:click
     on:keydown
+    on:mousedown|preventDefault
+    on:pointerdown|preventDefault
```
mcous, line 101: These prevent focus from happening, so focus remains in the combobox even as the menu is toggled open and closed. `click` events continue to fire

packages/core/.eslintrc.cjs
micheal-parks: Agreed all around with this

mcous: https://github.com/viamrobotics/js-config/pull/34

DTCurrie (approved):
✅

ethanlook (approved):

PR #459, opened 2024-01-04
APP-3126: fix searchable select value and focus issues

mcous:
## Overview

This rebuilds the `<SearchableSelect>` from the ground up with the following goals:

- Fix a11y of the element so it follows [Editable Combobox With List Autocomplete Example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/) from the ARIA APG
    - Elements should be a  `combobox` and `listbox`; currently it is a `combobox` and a `menu`, which turns out, is not correct! (This is also a problem with the resource picker over in app)
    - DOM focus should remain in the `<input role="combobox">` while options are highlighted, using `aria-activedescendant` to manage visual/a11y focus
    - Keyboard handlers should follow WAI recommendations
- Update props of the component to match needs + actual usage
    - Nothing uses the `heading` props, and it wasn't in Figma, so I removed it
    - Add `onFocus` and `onBlur` handlers for validation triggers
    - Replaced events with function props
- Allow arbitrary input values
    - New `exclusive` prop sets the select to allow arbitrary input (default) or exclude it
    - Going off existing usage in the app and the designs, I removed the `button` prop since it didn't quite capture the intent
    - Instead, if `exclusive` is `false` and the search doesn't match an option, a new option at the bottom with a `plus` icon and the current `value` will appear
- Ensure unit tests of `<SearchableSelect>` actually test the a11y
- Fix errant form submissions due to lack of `type="button"` on vanilla `<button>` elements 
- Fix input lockups where the menu can close and you can no longer type into it
- Fix various styling issues (e.g. incorrect menu and option padding)

## Change log

- Rewrite `<SearchableSelect>` and its tests
- Pull in `$lib/keyboard.ts` from the app, where it's been really helpful to declare keyboard interactions

### Notable features

- Allow arbitrary input by default (see above)
- Explicit `errorID` prop for `aria-errormessage`

### Notable removals

- `heading` prop removed (see above)
- `button` prop removed (see above)
- Sync keyboard state with mouse hover state removed
    - The APG example didn't do it as far as I could tell
    - I don't think the extra implementation required is worth the noise, and hover handlers seem to be a source of bugs in the existing `<SearchableSelect>`

## Review requests

Please smoke test the new `<SearchableSelect>` along with the old `<MultiSelect>`, which still uses a lot of infrastructure built for `<SearchableSelect>`. Be on the lookout for styling regressions. If you find any behavior bugs with `<MultiSelect>`, try to figure out if they're regressions of if it's buggy on `main`.

- [ ] Storybook
- [ ] Playground

In the mean time, I'll be prepping an app PR to incorporate these breaking changes

### DIFF
--- a/packages/core/src/lib/keyboard.ts
+++ b/packages/core/src/lib/keyboard.ts
@@ -1,0 +1,26 @@
+export type KeyMap = Partial<Record<string, KeyEventHandler | KeyOptions>>;
+
+export type KeyEventHandler = (event: KeyboardEvent) => unknown;
+
+export interface KeyOptions {
+  handler: KeyEventHandler;
+  preventDefault?: boolean;
+}
+
+export const createHandleKey = (keyMap: KeyMap) => {
+  return (event: KeyboardEvent): void => {
+    const options = keyMap[event.key];
+    const handler = typeof options === 'function' ? options : options?.handler;
+    const preventDefault =
+      typeof options === 'object' ? options.preventDefault : true;
+
+    if (typeof handler === 'function') {
+      handler(event);
+
+      if (preventDefault) {
+        event.preventDefault();
+        event.stopPropagation();
+      }
+    }
+  };
+};
--- a/packages/core/src/lib/select/__tests__/searchable-select.spec.ts
+++ b/packages/core/src/lib/select/__tests__/searchable-select.spec.ts
@@ -275,7 +275,7 @@ describe('combobox list', () => {
     expect(options).toHaveLength(3);
     expect(options[0]).toHaveAccessibleName('hello from');
     expect(options[1]).toHaveAccessibleName('the other side');
-    expect(options[2]).toHaveAccessibleName('other: hello');
+    expect(options[2]).toHaveAccessibleName('hello');
     expect(options[2]).toHaveAttribute('aria-selected', 'false');
   });
 
@@ -287,7 +287,7 @@ describe('combobox list', () => {
     await user.type(search, 'asdf');
     const { options } = getResults();
 
-    expect(options[2]).toHaveAccessibleName('other: asdf');
+    expect(options[2]).toHaveAccessibleName('asdf');
     expect(options[2]).toHaveAttribute('aria-selected', 'true');
     expect(search).toHaveAttribute('aria-activedescendant', options[2]?.id);
   });
@@ -331,6 +331,17 @@ describe('combobox list', () => {
     const { options } = getResults();
 
     expect(options).toHaveLength(3);
+  });
+
+  it('adds a prefix to the "other" option display text', async () => {
+    const user = userEvent.setup();
+    renderSubject({ otherOptionPrefix: 'You said:' });
+
+    const { search } = getResults();
+    await user.type(search, 'hello');
+    const { options } = getResults();
+
+    expect(options[2]).toHaveAccessibleName('You said: hello');
   });
 
   it('empties input value if closed and exclusive', async () => {
--- a/packages/core/src/lib/select/__tests__/searchable-select.spec.ts
+++ b/packages/core/src/lib/select/__tests__/searchable-select.spec.ts
@@ -308,8 +308,6 @@ describe('combobox list', () => {
     const { options } = getResults();
 
     expect(options).toHaveLength(2);
-    expect(options[0]).toHaveAccessibleName('hello from');
-    expect(options[1]).toHaveAccessibleName('the other side');
   });
 
   it('empties input value if closed and exclusive', async () => {
--- a/packages/core/src/lib/select/__tests__/searchable-select.spec.ts
+++ b/packages/core/src/lib/select/__tests__/searchable-select.spec.ts
@@ -196,8 +196,11 @@ describe('combobox list', () => {
     const { search, options } = getResults();
 
     await user.click(search);
-    await user.click(options[0]!);
+    // TODO(mc, 2024-02-03): replace .click with userEvent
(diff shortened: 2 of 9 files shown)


packages/core/src/lib/keyboard.ts
mcous, line 1: I didn't add any explicit tests for this unit, because they'd be fully redundant with component tests. Let me know if this bothers you, though, because I can add more tests

packages/core/src/lib/select/__tests__/searchable-select.spec.ts
```diff
@@ -1,282 +1,459 @@
...
+  it('expands the listbox on focus', async () => {
+    const user = userEvent.setup();
+    renderSubject();
+
+    const { search, button, list } = getResults();
 
-    const select = screen.getByPlaceholderText('Select an option');
+    expect(list).toHaveClass('hidden');
+    expect(search).toHaveAttribute('aria-expanded', 'false');
+    expect(button).toHaveAttribute('aria-expanded', 'false');
```
mcous, line 81: These states are driven by the same logic. Perhaps checking all three is redundant in subsequent tests

packages/core/src/lib/select/__tests__/searchable-select.spec.ts
```diff
@@ -1,282 +1,459 @@
...
+    renderSubject({ exclusive: true });
 
-    expect(select.value).toBe('C.) Option');
+    const { search } = getResults();
+    await user.type(search, 'hello');
+    const { options } = getResults();
+
+    expect(options).toHaveLength(2);
+    expect(options[0]).toHaveAccessibleName('hello from');
+    expect(options[1]).toHaveAccessibleName('the other side');
```
mcous, line 301: Redundant, remove

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -1,241 +1,245 @@
...
-      open={$isOpen}
-      id={menuId}
-      bind:element={menu}
-      on:mouseleave={resetNavigationIndex}
-    >
-      {#if heading}
+  <ul
+    id={LIST_ID}
+    role="listbox"
+    class:hidden={!isExpanded}
```
mcous, line 199: Went with `hidden` over `invisible` here so things wouldn't be clickable; but I'm unsure of if `invisible` was picked for some other important reason. IIRC both should be hidden from screen-readers, so I don't know of any a11y differences

packages/core/src/lib/keyboard.ts
mcous, line 1: I didn't add any explicit tests for this unit, because they'd be fully redundant with component tests. Let me know if this bothers you, though, because I can add more tests

DTCurrie (approved):
✅

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -1,241 +1,261 @@
...
+/** The placeholder of the input. */
+export let placeholder = '';
+
+/**
+ * Whether value must be constrained to `options` on change.
+ * - `false` (Default) - any value may be selected
+ * - `true` - value must be in `options`
+ * - `Function` - value must be in `options` or pass the test function
+ */
+export let exclusive: boolean | ((value: string) => boolean) = false;
```
mcous, line 39: This is a change after using this component in the app; the data UI needs a little more involved filtering than "is the value in the options"

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -1,241 +1,261 @@
...
 
-  if (value === option) {
-    return;
+  setMenuState(CLOSED);
+
+  if (nextValue !== previousValue) {
+    value = nextValue;
+    previousValue = nextValue;
+    onChange?.(nextValue);
   }
```
mcous, line 135: Avoid duplicate `onChange` events

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -1,241 +1,261 @@
...
+    id={LIST_ID}
+    role="listbox"
+    class:hidden={!isExpanded}
+    class="max-h-36 flex-col overflow-y-auto border border-gray-9 bg-white py-1 drop-shadow-md"
+  >
+    {#each allOptions as { option, highlight } (option)}
+      {@const isSelected = activeOption?.option === option}
+      {@const isOther = otherOption?.option === option}
+
+      {#if isOther && allOptions.length > 1}
```
mcous, line 220: Don't display the divider if there's only one option

mcous (commented):
Change after using this PR with app

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -1,241 +1,268 @@
...
+/** Option sorting behavior. */
 export let sort: SortOption = SortOptions.DEFAULT;
 
-/**
- * An optional call-to-action button to render at the bottom of the select menu
- * that will emit the `buttonclick` event when actioned.
- */
-export let button: { text: string; icon: IconName } | undefined = undefined;
+/** Prefix to display with the arbitrary text option, if non-exclusive. */
+export let otherOptionPrefix = '';
```
mcous, line 51: One last addition to this PR: the app needs to be able to display a prefix on the arbitrary input for one of it's usages, e.g. `Create: some arbitrary thing`

Previously this was done by `buttontext`, but with `buttontext`, you had to thread the current value in manually. For now, all we need is a prefix, so I think the simpler, less flexible prop is easier

PR #461, opened 2024-01-08
RSDK-6218 - make postprocessed slam points the viam slam brand color "cyberpunk"

kim-mishra:
This PR makes postprocessed points from the SLAM map  the viam slam brand color "cyberpunk". Regular points only have confidence in the B value, but for postprocessed points we send 100 percent confidence through the red value. Because of this, we can check X and set the color accordingly. 
screenshot_2024-01-10_at_12 24 02_pm_720

### DIFF
--- a/packages/blocks/src/lib/slam-map-2d/color-map.ts
+++ b/packages/blocks/src/lib/slam-map-2d/color-map.ts
@@ -45,13 +45,16 @@ const colorBuckets = (probability: number): THREE.Vector3 => {
 export const mapColorAttributeGrayscale = (colors: THREE.BufferAttribute) => {
   for (let i = 0; i < colors.count; i += 1) {
     /*
-     * Probability is currently assumed to be held in the rgb field of the PCD map, on a scale of 0 to 100.
-     * ticket to look into this further https://viam.atlassian.net/browse/RSDK-2605
+     * Probability is currently assumed to be held in the b part of the rgb field of the PCD map, on a scale of 0 to 100.
+     * ticket to look into this further https://viam.atlassian.net/browse/RSDK-2605. 
+     * 
+     * User generated points that have been added to the pcd are marked as such by putting 100% probability
+     * in the r part of the rgb field. If that has been set, we will draw the point in blue instead of mapping it to its
+     * corresponding shade of gray to signal it is a user generated point.
      */
 
-    // if the point has been postprocessed
     if (colors.getX(i) === 1) {
-      colors.setXYZ(i, 0.8, 0, 0);
+      colors.setXYZ(i, 0, 0, 1);
       continue;
     }
 
--- a/packages/blocks/src/lib/slam-map-2d/color-map.ts
+++ b/packages/blocks/src/lib/slam-map-2d/color-map.ts
@@ -51,10 +51,10 @@ export const mapColorAttributeGrayscale = (colors: THREE.BufferAttribute) => {
 
     // if the point has been postprocessed
     if (colors.getX(i) === 1) {
-      colors.setXYZ(i, .8, 0, 0);
-      continue
+      colors.setXYZ(i, 0.8, 0, 0);
+      continue;
     }
-    
+
     const colorMapPoint = colorBuckets(colors.getZ(i) * 10);
     colors.setXYZ(i, colorMapPoint.x, colorMapPoint.y, colorMapPoint.z);
   }
--- a/packages/blocks/src/lib/slam-map-2d/color-map.ts
+++ b/packages/blocks/src/lib/slam-map-2d/color-map.ts
@@ -1,4 +1,5 @@
 import * as THREE from 'three';
+import { theme } from '@viamrobotics/prime-core/theme';
 
 /*
  * this color map is greyscale. The color map is being used map probability values of a PCD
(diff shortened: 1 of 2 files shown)


packages/blocks/src/lib/slam-map-2d/color-map.ts
```diff
@@ -48,6 +48,13 @@ export const mapColorAttributeGrayscale = (colors: THREE.BufferAttribute) => {
      * Probability is currently assumed to be held in the rgb field of the PCD map, on a scale of 0 to 100.
      * ticket to look into this further https://viam.atlassian.net/browse/RSDK-2605
      */
+
+    // if the point has been postprocessed
+    if (colors.getX(i) === 1) {
+      colors.setXYZ(i, .8, 0, 0);
```
kim-mishra, line 54: For some reason setting this to `colorMapPoint.x` made the line black.

JohnN193 (commented):
hmm the red doesn't actually look good with the axes colors. maybe blue or purple?

prob good to keep esha in the loop since this could be user facing down the line

kim-mishra:
> hmm the red doesn't actually look good with the axes colors. maybe blue or purple?
> 
> prob good to keep esha in the loop since this could be user facing down the line

made it blue for now will run this by esha and make sure its okay for the interim

randhid:
> This PR makes postprocessed points from the SLAM map red. Regular points only have confidence in the B value, but for postprocessed points we send 100 percent confidence through the red value. Because of this, we can check X and set the color accordingly. Screenshot 2024-01-08 at 7 23 49 PM

Is the z axis blue? Also please ad more comments before, imagine someone doesn't know what post-processing is and explain it from no knwledge of the slam map in case someone from fleet has to edit the rc card.

kim-mishra:
> > This PR makes postprocessed points from the SLAM map red. Regular points only have confidence in the B value, but for postprocessed points we send 100 percent confidence through the red value. Because of this, we can check X and set the color accordingly. Screenshot 2024-01-08 at 7 23 49 PM
> 
> Is the z axis blue? Also please ad more comments before, imagine someone doesn't know what post-processing is and explain it from no knwledge of the slam map in case someone from fleet has to edit the rc card.


> > This PR makes postprocessed points from the SLAM map red. Regular points only have confidence in the B value, but for postprocessed points we send 100 percent confidence through the red value. Because of this, we can check X and set the color accordingly. Screenshot 2024-01-08 at 7 23 49 PM
> 
> Is the z axis blue? Also please ad more comments before, imagine someone doesn't know what post-processing is and explain it from no knwledge of the slam map in case someone from fleet has to edit the rc card.

@randhid the prime component is called slam-map-2D so I don't think we have to worry about a z-axis. Adding in comments now!

JohnN193 (approved):
LGTM, just update the name of the pr/commit

packages/blocks/src/lib/slam-map-2d/color-map.ts
```diff
@@ -45,9 +45,19 @@ const colorBuckets = (probability: number): THREE.Vector3 => {
...
+     * Probability is currently assumed to be held in the b part of the rgb field of the PCD map, on a scale of 0 to 100.
+     * ticket to look into this further https://viam.atlassian.net/browse/RSDK-2605.
+     *
+     * User generated points that have been added to the pcd are marked as such by putting 100% probability
+     * in the r part of the rgb field. If that has been set, we will draw the point in blue instead of mapping it to its
+     * corresponding shade of gray to signal it is a user generated point.
      */
+
+    if (colors.getX(i) === 1) {
+      colors.setXYZ(i, 0, 0, 1);
```
ethanlook, line 57: Could we color this something other than blue? (Apologies, I see there was already some discourse about the colors.)

Generally, we have been thinking about our design language in visualization tools and color is often used to indicate meaning. It is quite standard for x, y, and z to be displayed as red, green, and blue. Although this is a 2D map, blue still has meaning in our other visualization tools.

Would you be open to picking a different color? I'm happy to brainstorm further, and we could even grab Nat for a few minutes as well.

Here's a list of our brand colors, which might have a good choice: https://github.com/viamrobotics/prime/blob/main/packages/core/theme.ts#L65-L75

And you can import them like so:

```
import { theme } from '@viamrobotics/prime-core/theme';
```

Then you could make a file-level constant for your chosen color:

```
const USER_GENERATED_POINT_COLOR = new THREE.Color(theme.extend.colors.cyberpunk);
```

And use it to set the attribute:

```
colors.setXYZ(i, USER_GENERATED_POINT_COLOR.r, USER_GENERATED_POINT_COLOR.g, USER_GENERATED_POINT_COLOR.b);
```

kim-mishra: No problem, will update the color. @EshaMaharishi do you have a preference on the color?

EshaMaharishi: 'raspberry' looks good to me, but happy to go with something else if Ethan/Nat prefer!

kim-mishra: going with cyberpunk as per Nat's request

ethanlook (approved):
🚀

ethanlook (approved):
Version bump looks good! After merge, checks will run again and `npm-publish` will be run to publish this version. Once that's finished, from `rdk/web/frontend`, run:

```
npm install -D @viamrobotics/prime-blocks
```

PR #462, opened 2024-01-09
APP-3126: make searchable select autoresize

mcous:
## Overview

There is a bug in the new searchable select if it is rendered in a parent with 0 width (or other changeable size): the floating options list size will be calculated on mount without being updated if the parent's width is later expanded, resulting in the options list being the wrong size in the wrong place.

## Change log

- Turn on auto-resizing for the `SearchableSelect` options list
- Simplify `Floating` and `floatingStyle` to more directly accept `floating-ui/dom` options
    - This will assist some other floating needs in the app

## Review requests

- Code makes sense
- Smoke test the storybook and playground, especially tooltips
- I will be smoke testing against my local app and posting the results here

### DIFF
--- a/packages/core/src/lib/floating/floating-style.ts
+++ b/packages/core/src/lib/floating/floating-style.ts
@@ -17,8 +17,6 @@ import {
   type SizeOptions,
 } from '@floating-ui/dom';
 
-export type {};
-
 import { derived, writable, type Readable } from 'svelte/store';
 import { noop } from 'lodash-es';
 
--- a/packages/core/src/lib/floating/floating-style.ts
+++ b/packages/core/src/lib/floating/floating-style.ts
@@ -11,7 +11,13 @@ import {
   type Side,
   type ComputePositionConfig,
   type ReferenceElement,
+  type OffsetOptions,
+  type FlipOptions,
+  type ShiftOptions,
+  type SizeOptions,
 } from '@floating-ui/dom';
+
+export type {};
 
 import { derived, writable, type Readable } from 'svelte/store';
 import { noop } from 'lodash-es';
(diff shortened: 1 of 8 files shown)


packages/core/src/lib/floating/floating-style.ts
```diff
@@ -11,14 +11,24 @@ import {
   type Side,
   type ComputePositionConfig,
   type ReferenceElement,
+  type OffsetOptions,
+  type FlipOptions,
+  type ShiftOptions,
+  type SizeOptions,
 } from '@floating-ui/dom';
 
+export type {};
```
ethanlook, line 20: 🤔  what do we have here?

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -209,9 +209,10 @@ const handleKeydown = createHandleKey({
   bind:value
 />
 <Floating
-  matchWidth
   offset={4}
   referenceElement={inputElement}
+  size={matchWidth}
+  auto
```
mcous, line 215: TODO: set the `<li>` below to `truncate`

mcous (commented):
Missed one thing

ethanlook (approved):
Code, playground, and storybook all look great.

This PR reminds me that we should have a test environment for the storybook and use the storybook for development. It will make it easier to review code and force us to keep storybook in sync.

DTCurrie (approved):
✅

mrloureed (approved):
LGTM

packages/core/src/lib/floating/floating-style.ts
mcous: Oops!

PR #463, opened 2024-01-10
Add frame editor icons

micheal-parks:
Also compressed size of other custom icons.

Screenshot 2024-01-16 at 13 53 57


packages/core/src/lib/icon/icons.ts
```diff
@@ -92,24 +92,35 @@ export const paths = {
   undo: MDI.mdiUndo,
   upload: MDI.mdiUpload,
   'viam-component':
-    'M12.0001 3.95547L3.95547 12.0001L12.0001 20.0448L20.0448 12.0001L12.0001 3.95547ZM12.5305 1.62201C12.2376 1.32912 11.7627 1.32912 11.4698 1.62201L1.62201 11.4698C1.32912 11.7627 1.32912 12.2376 1.62201 12.5305L11.4698 22.3782C11.7627 22.6711 12.2376 22.6711 12.5305 22.3782L22.3782 12.5305C22.6711 12.2376 22.6711 11.7627 22.3782 11.4698L12.5305 1.62201Z',
+    'm12 4-8 8 8 8 8-8-8-8Zm.5-2.4a.8.8 0 0 0-1 0l-9.9 9.9c-.3.3-.3.7 0 1l9.9 9.9c.3.3.7.3 1 0l9.9-9.9c.3-.3.3-.7 0-1l-9.9-9.9Z',
```
mcous, line 95: Could we punt existing icon compression to a different PR in case we have any issues that would cause a revert?

mcous (approved):

packages/core/src/lib/icon/icons.ts
micheal-parks: For sure

ehhong:
> > what's the reason why these are in app rather than prime?
> 
> Not sure I understand. Could you explain a bit more?

brain fart moment -- thought this was an app PR. please ignore!

PR #464, opened 2024-01-10
Do not add a color class to icon

mcous:
## Overview

#456 added the `text-gray-6` class to the `<Icon>` component. This, in turn, made it impossible to override the color of an icon with any class that comes later in the Tailwind stylesheet, because `text-gray-6` will always be present, so color will be applied according to CSS specificity.

The PR reverts that addition, which will cause all icons to default to the current text color, fixing several regressions in the app. If we want to default icons to a specific color, we will need to come up with a different approach that conditionally adds a color classname

## Change log

- Remove `text-gray-6` from icon

### DIFF
--- a/packages/core/src/lib/icon/icon.svelte
+++ b/packages/core/src/lib/icon/icon.svelte
@@ -41,8 +41,6 @@ export let size: Size = 'base';
 /** Additional CSS classes to pass to the svg. */
 let extraClasses: cx.Argument = '';
 export { extraClasses as cx };
-
-const hasNameProperty = Object.hasOwn(paths, name);
 </script>
 
 <!--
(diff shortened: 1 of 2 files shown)


mrloureed (approved):
LGTM!

packages/core/src/lib/icon/icon.svelte
```diff
@@ -41,23 +41,14 @@ export let size: Size = 'base';
...
 <!--
   Accessibility approach for icon svgs taken from:
   https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label
 -->
 <svg
-  class={cx(
-    'text-gray-6',
-    sizes[size],
-    {
-      hidden: !hasNameProperty,
```
ethanlook, line 57: Why the removal of `hidden` here?

ethanlook (approved):
Gah, good find. This impacted the banner and a few other components as well. Thanks for the fix!

packages/core/src/lib/icon/icon.svelte
mcous: It had been annoying me and it's tripped folks up in the past. If a `name` prop is passed in to the `Icon` that isn't in `iconData`, that is a bug. Adding a `hidden` class to the icon when `name` is invalid feels like pretending that everything is ok

PR #465, opened 2024-01-11
RSDK-6093 - Fix Lat,Lng order is flipped in nav service rc card

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-6093)

Swaps the order of (lng, lat) in the Waypoints tab on the nav RC card.

Before:
Screenshot 2024-01-11 at 2 42 24 PM

After:
Screenshot 2024-01-11 at 2 41 03 PM

### DIFF
--- a/packages/blocks/package.json
+++ b/packages/blocks/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-blocks",
-  "version": "0.0.24",
+  "version": "0.0.25",
   "publishConfig": {
     "access": "public"
   },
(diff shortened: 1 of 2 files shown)


ethanlook (approved):
Great catch, and thank you for the update! Do you mind also bumping the version [here](https://github.com/viamrobotics/prime/blob/main/packages/blocks/package.json#L3) in this PR? Then we'll ultimately need to follow up to use the latest version in RDK and App as well.

nicksanford:
> [Ticket](https://viam.atlassian.net/browse/RSDK-6093)
> 
> Swaps the order of (lng, lat) in the Waypoints tab on the nav RC card.

Done!

ethanlook (approved):

PR #466, opened 2024-01-16
Compress icons

micheal-parks:


ethanlook (approved):
Chatted IRL - would love to push this back upstream to the Figmas. Thank you!

cc @nathartman

PR #470, opened 2024-01-23
RSDK-6434 - Refactor the Prime SlamMap2D component to better path type

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-6434) 

1. Change `SlamMap2d`'s motionPath parameter to be of type `Float32Array | undefined` from `string | undefined`
2. Bump `prime/blocks`npm package version

Follow up from https://github.com/viamrobotics/rdk/pull/3450

I confirmed that the output is still the same in the storybook:
Screenshot 2024-01-23 at 4 32 45 PM

### DIFF
--- a/packages/blocks/src/lib/slam-map-2d/motion-path.svelte
+++ b/packages/blocks/src/lib/slam-map-2d/motion-path.svelte
@@ -34,7 +34,7 @@ const updatePath = (xy?: Float32Array) => {
     }
   }
 
-  lineGeometry.setPositions(new Float32Array(xyz));
+ lineGeometry.setPositions(xyz);
 
   return lineGeometry;
 };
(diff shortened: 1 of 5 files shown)


packages/blocks/src/lib/slam-map-2d/motion-path.svelte
```diff
@@ -15,30 +17,24 @@ import { renderOrder } from './render-order';
...
+    const x = xy[i];
+    const y = xy[i + 1];
+    if (x !== undefined && y !== undefined) {
+      xyz.push(x, y, 0);
     }
   }
 
-  const vertices = new Float32Array(points);
-  lineGeometry.setPositions(vertices);
+  lineGeometry.setPositions(new Float32Array(xyz));
```
micheal-parks, line 37: Ah dang, I completely forgot that `setPositions` requires an `xyz` and not `xy` array.

We can probably find a way to avoid copies altogether in the future but for now we could remove one copy here just by passing a regular JS array:

 ```suggestion
  lineGeometry.setPositions(xyz);
```

micheal-parks (approved):

PR #472, opened 2024-01-29
RSDK-6492 - ViamObject3D should not be in in degrees

nfranczak:
https://viam.atlassian.net/browse/RSDK-6492


micheal-parks (approved):
Thank you! You'll also have to bump the version field in `blocks/package.json` for this to be shipped to NPM.

Once this is shipped, we'll have to upgrade the version of PRIME in RDK and APP. In app we should make sure that there's not another incorrect assumption conversion-wise that would cause a bug with this fix. I can help with that!

nfranczak:
> You'll also have to bump the version field in blocks/package.json for this to be shipped to NPM.

oki, I think I did this.

> Once this is shipped, we'll have to upgrade the version of PRIME in RDK and APP. In app we should make sure that there's not another incorrect assumption conversion-wise that would cause a bug with this fix. I can help with that!

I will definitely need your help with this.
for app do i only need to update `ui/package.json`
for rdk, do i only need to update `web/frontend/package-lock.json`?

To make sure there is not another incorrect assumption conversion-wise, do you recommend just starting a server from app and rdk to see if the bug persists?

PR #476, opened 2024-02-07
Fix <SearchableSelect> shadows

mcous:
## Overview

Various components were using tailwind's [`drop-shadow`](https://tailwindcss.com/docs/drop-shadow) classes to add shadows. As stated by the Tailwind docs:

> This is useful for applying shadows to irregular shapes, like text and SVG elements. For applying shadows to regular elements, you probably want to use [box shadow](https://tailwindcss.com/docs/box-shadow) instead.

This PR swaps in [`shadow`](https://tailwindcss.com/docs/box-shadow), accordingly


ethanlook (approved):
i'm the problem

PR #477, opened 2024-02-07
Design maintenance prime 7 feb 2024

mrloureed:
Fix code snippet copy button color, hover state and position.

https://github.com/viamrobotics/prime/assets/17263/444c6d05-7a56-4cce-8897-f00c0c11a2d1

### DIFF
--- a/packages/core/src/lib/code-snippet.svelte
+++ b/packages/core/src/lib/code-snippet.svelte
@@ -164,13 +164,15 @@ onMount(async () => {
       ></pre>
 
     {#if showCopyButton}
-      <IconButton
-        class="text-black m-2"
-        icon={COPY_STATES[copyState].icon}
-        label={COPY_STATES[copyState].label}
-        on:click={copyToClipboard}
-        on:keyup={copyToClipboard}
-      />
+      <div class="gray-6">
+        <IconButton
+          cx="m-1 text-gray-6"
+          icon={COPY_STATES[copyState].icon}
+          label={COPY_STATES[copyState].label}
+          on:click={copyToClipboard}
+          on:keyup={copyToClipboard}
+        />
+      </div>
     {/if}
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/code-snippet.svelte
```diff
@@ -155,7 +155,7 @@ onMount(async () => {
     <figcaption><slot name="caption" /></figcaption>
   {/if}
 
-  <div class="flex gap-x-4 bg-light p-2">
+  < class="flex gap-x-4 bg-light p-2 items-baseline">
```
micheal-parks, line 158: 😆

micheal-parks (approved):

packages/core/src/lib/code-snippet.svelte
micheal-parks: Give me back my div!!

ethanlook (approved):
LGTM once you give us our `div` back!!

packages/core/src/lib/code-snippet.svelte
mrloureed: https://media.giphy.com/media/wFwfQDzwuLKzC/giphy.gif?cid=ecf05e47u70rx5dzqqto3m0losrhyx9rzehc84ka62w4ss7y&ep=v1_gifs_search&rid=giphy.gif&ct=g

nathartman (changes requested):
i'd request that we lose the Notify confirmation since we're already providing feedback via the checkmark, otherwise looks beautiful 😎

mrloureed:
> > i'd request that we lose the Notify confirmation since we're already providing feedback via the checkmark, otherwise looks beautiful 😎
> 
> This would be handled in App so a follow up task I recommend.

Correct @micheal-parks ?

PR #478, opened 2024-02-08
APP-3302: add `multiple` prop to `<SearchableSelect>`

mcous:
## Overview

Back in #459, we made some minor fixes to the `<SearchableSelect>` element:

- Switch from `menu` to `listbox`, as recommended by WAI
- Keep DOM focus on search box, use `aria-activedescendant` for visual focus
- Minor visual fixes (e.g. padding)
- Switch to `floating-ui` for menu placement

This PR adds a `multiple` prop to allow the same `<SearchableSelect>` to be used to select multiple options.

## Change log

- Add `multiple` prop to `<SearchableSelect>`
- Add `bind`-able `values` prop for use when `multiple={true}`
- Add `onMultiChange` prop for use when `multiple={true}`

## Review requests

- Try out the storybook
- Try out the `core` playground

I did not touch `<Multiselect>` component in this PR. The main difference between the two is that `<Multiselect>` adds pills.

I think we can rework `<Multiselect>` to use `<SearchableSelect>` under the hood, but I'd rather tackle that later, because there's a few data pages that rely on the current component. There's also upcoming designs that display the pills in a slightly different way than before

### DIFF
--- a/packages/core/src/lib/select/__tests__/searchable-select.spec.ts
+++ b/packages/core/src/lib/select/__tests__/searchable-select.spec.ts
@@ -6,6 +6,7 @@ import type { ComponentProps } from 'svelte';
 import { SearchableSelect as Subject, InputStates } from '$lib';
 
 const onChange = vi.fn();
+const onMultiChange = vi.fn();
 const onFocus = vi.fn();
 const onBlur = vi.fn();
 
@@ -13,6 +14,7 @@ const renderSubject = (props: Partial<ComponentProps<Subject>> = {}) => {
   return render(Subject, {
     options: ['hello from', 'the other side'],
     onChange,
+    onMultiChange,
     onFocus,
     onBlur,
     ...props,
@@ -33,8 +35,8 @@ const getResults = (): {
   return { search, button, list, options };
 };
 
-describe('combobox list', () => {
-  it('controls a listbox', () => {
+describe('SearchableSelect', () => {
+  it('is a combobox that controls a listbox', () => {
     renderSubject();
 
     const { search, button, list } = getResults();
@@ -43,6 +45,7 @@ describe('combobox list', () => {
     expect(button).toHaveAttribute('aria-controls', list.id);
     expect(search).toHaveAttribute('aria-controls', list.id);
     expect(search).toHaveAttribute('aria-autocomplete', 'list');
+    expect(search).not.toHaveAttribute('aria-multiselectable');
   });
 
   it('has a placeholder', () => {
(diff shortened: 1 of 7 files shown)


packages/storybook/src/stories/select.stories.svelte
```diff
@@ -57,19 +54,21 @@ import {
   </Label>
 </Story>
 
-<Story name="Multi">
-  <Label cx="h-[200px]">
-    Select multiple objects
-    <Multiselect
+<Story name="Searchable with multiple">
+  <Label cx="h-[200px] max-w-[200px]">
```
DTCurrie, line 58: Not for this PR but I am seeing us use `200px` a lot as an arbitrary value, starting to feel like a `h-50`/`w-50` class might be worth it at some point.

ethanlook: Yeah it's mostly just coming up in the stories. I'm not recalling any other places.

Oabraham1: I never understood this, but why did we settle on a max width of 200px?

DTCurrie (approved):
✅

packages/storybook/src/stories/select.stories.svelte
ethanlook: Design typically wants these inputs to have a max width of something like 200px instead of full width. The exact width is not something we want to prescribe in the component, however. We want the stories to demonstrate that to indicate proper usage.

Oabraham1: Ahh makes sense, thanks for explaining

packages/core/src/lib/select/__tests__/searchable-select.spec.ts
```diff
@@ -471,4 +499,78 @@ describe('combobox list', () => {
...
+      renderSubject({ multiple: true });
+
+      const { search, options } = getResults();
+
+      expect(search).toHaveAttribute('aria-multiselectable', 'true');
+
+      await user.click(search);
+
+      // TODO(mc, 2024-02-03): replace .click with userEvent
+      // https://github.com/testing-library/user-event/issues/1119
```
ethanlook, line 515: oh fun!

packages/core/src/lib/select/searchable-select.svelte
ethanlook, line 1: I wonder if there's a way to type the props such that it requires value + onChange versus values + onMultiChange. Not a blocker, but would be nice to enforce.

ethanlook (approved):
Just was able to give it a code review - LGTM

Oabraham1 (approved):
✅  I think modifying the `MultiSelect` element to use this `SearchableSelect` under the hood vs replacing the `MultiSelect` with `SearchableSelect` will be great especially as the `SearchableSelect` does not have a `clearable` prop that lets us clear all selections at the moment

Looks great to me

packages/core/src/lib/select/searchable-select.svelte
mcous: Will noodle on this for a few minutes and get back to you

mcous: Tried with Generic types, and it did not go well. Maybe a runtime assertion? I think maybe it becomes a moot point once we incorporate this into `<Multiselect>` and it can be our canonical option

PR #486, opened 2024-02-26
APP-3921: Add custom part status icon

anjinai:
## Overview

Adds a part status icon that is two-toned for use with machines that have some parts that are online and some parts that are offline. 

[Figma](https://www.figma.com/file/GSi2nRrsKHQnP2uYNx5XMK/Make-UI-Not-Shitty-II?type=design&node-id=3504-23326&mode=design&t=s69t2tmj9eJzhs5y-4)

Screenshot 2024-02-26 at 5 45 16 PM


packages/core/src/lib/icon/icons.ts
```diff
@@ -116,6 +116,12 @@ export const paths = {
...
   'viam-local-module':
     'M3 3v18h18V3H3Zm5.6 9.9H5V5h8v3.6H8.6V13Zm-3.6 6h14V5h-4v5.6h-4.4V15H5v4Z',
+  'viam-mixed-part-status': {
+    path1:
+      'M8 7.33341C7.26667 7.33341 6.66667 7.93341 6.66667 8.66675C6.66667 9.40008 7.26667 10.0001 8 10.0001C8.73333 10.0001 9.33333 9.40008 9.33333 8.66675C9.33333 7.93341 8.73333 7.33341 8 7.33341ZM12 8.66675C12 6.46008 10.2067 4.66675 8 4.66675C5.79333 4.66675 4 6.46008 4 8.66675C4 10.1467 4.80667 11.4334 6 12.1267L6.66667 10.9667C5.87333 10.5001 5.33333 9.65341 5.33333 8.66675C5.33333 7.19341 6.52667 6.00008 8 6.00008C9.47333 6.00008 10.6667 7.19341 10.6667 8.66675C10.6667 9.65341 10.1267 10.5001 9.33333 10.9667L10 12.1267C11.1933 11.4334 12 10.1467 12 8.66675Z',
+    path2:
+      'M8.00016 2C4.32016 2 1.3335 4.98667 1.3335 8.66667C1.3335 11.1333 2.6735 13.28 4.66016 14.4333L5.32683 13.28C3.74016 12.3533 2.66683 10.64 2.66683 8.66667C2.66683 5.72 5.0535 3.33333 8.00016 3.33333C10.9468 3.33333 13.3335 5.72 13.3335 8.66667C13.3335 10.64 12.2602 12.3533 10.6668 13.28L11.3335 14.4333C13.3268 13.28 14.6668 11.1333 14.6668 8.66667C14.6668 4.98667 11.6802 2 8.00016 2Z',
+  },
```
mcous, line 124: I like this! Let's get a little more generic, since two-tone icons seem like a thing that we'll want more of! Let's make the data look like:

```ts
'viam-mixed-part-status': [
  {
    path: '...'
  },
  {
    path: '...',
    opacity: '0.5'
  }
]
```

(Some extra type definitions may be necessary, that's ok)

Then, instead of checking for `{#if name === 'viam-mixed-part-status'}`, we can do something like:

```svelte
<script lang="ts">
// ...

$: allPaths = typeof paths[name] === 'string'
  ? [{ path: paths[name] }]
  : paths[name]
</script>

<svg
  class={cx(sizes[size], extraClasses)}
  viewBox="0 0 24 24"
  aria-hidden="true"
  focusable="false"
  {...$$restProps}
>
  {#each allPaths as { path: d, opacity }}
    <path
      {d}
      {opacity}
      fill-rule="evenodd"
      fill="currentColor"
    />
  {/each}
</svg>

**Review by mcous — APPROVED** [review]

File: packages/core/src/lib/icon/icons.ts [hunk 2]
```diff
@@ -116,6 +116,12 @@ export const paths = {
     'M17 22v-2h3v-3h2v3.5c0 .4-.2.7-.5 1-.3.3-.6.5-1 .5H17ZM7 22H3.5c-.4 0-.7-.2-1-.5-.3-.3-.5-.6-.5-1V17h2v3h3v2ZM17 2h3.5c.4 0 .7.2 1 .5.3.3.5.6.5 1V7h-2V4h-3V2ZM7 2v2H4v3H2V3.5c0-.4.2-.7.5-1 .3-.3.6-.5 1-.5H7Zm6 15.3 4-2.4v-4.5l-4 2.3v4.6Zm-1-6.4 4-2.3-4-2.3-4 2.3 4 2.3Zm-5 4 4 2.3v-4.5l-4-2.3v4.5Zm11.2-7.3c.5.3.8.7.8 1.3v6.3c0 .6-.3 1-.8 1.4l-5.4 3.1c-.6.4-1 .4-1.6 0l-5.4-3.1c-.5-.4-.8-.8-.8-1.4V9c0-.6.3-1 .8-1.3l5.5-3.2a1.6 1.6 0 0 1 1.4 0l5.5 3.2Z',
   'viam-local-module':
     'M3 3v18h18V3H3Zm5.6 9.9H5V5h8v3.6H8.6V13Zm-3.6 6h14V5h-4v5.6h-4.4V15H5v4Z',
+  'viam-mixed-part-status': {
+    path1:
```
micheal-parks, line 120: One small suggestion: the above paths have been optimized by shaving off decimals to 0.1 precision. Bundle size will be reduced overall if we keep all of our icons to this rule. [This tool allows you to do this optimization](https://svgomg.net/) pretty easily.

anjinai: Ty ty!

packages/core/src/lib/icon/icons.ts
anjinai: Thanks!

packages/core/src/lib/icon/icon.svelte
```diff
@@ -41,22 +41,41 @@ export let size: Size = 'base';
...
 
 <!--
   Accessibility approach for icon svgs taken from:
   https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label
 -->
+
 <svg
   class={cx(sizes[size], extraClasses)}
-  viewBox="0 0 24 24"
+  viewBox={name === 'viam-mixed-part-status' ? '0 0 16 16' : '0 0 24 24'}
```
anjinai, line 68: I still needed to add a ternary here to set the correct viewbox for the mixed part status icon.  Is there a way that I can edit the viewbox of the mixed part status svg myself to match that of the other icons?

mcous: Scaled to 24px!

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none">
  <path fill="currentColor" d="M12 11c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2Zm6 2c0-3.31-2.69-6-6-6s-6 2.69-6 6c0 2.22 1.21 4.15 3 5.19l1-1.74c-1.19-.7-2-1.97-2-3.45 0-2.21 1.79-4 4-4s4 1.79 4 4c0 1.48-.81 2.75-2 3.45l1 1.74c1.79-1.04 3-2.97 3-5.19Z"/>
  <path fill="currentColor" d="M12 3C6.48 3 2 7.48 2 13c0 3.7 2.01 6.92 4.99 8.65l1-1.73C5.61 18.53 4 15.96 4 13c0-4.42 3.58-8 8-8s8 3.58 8 8c0 2.96-1.61 5.53-4 6.92l1 1.73c2.99-1.73 5-4.95 5-8.65 0-5.52-4.48-10-10-10Z" opacity=".5"/>
</svg>
```

Steps:

1. Export SVG from Figma
2. Drop into a new Figma file
3. Set to scale from center
4. Increase size from 16x16 to 24x24
5. Re-export
6. Remove cruft from SVG manually
7. Run through svgomg.com for good measure

packages/core/src/lib/icon/icons.ts
```diff
@@ -1,10 +1,19 @@
 import * as MDI from '@mdi/js';
 
+type Path = string;
+
+export interface CustomIcon {
+  path: string;
+  opacity?: number | undefined;
+}
```
mcous, line 8: Alternatively to re-exporting the icon, you could try something to incorporate different viewboxes into the path data interfaces. I think that makes everything a bit more complicated when we could simply use a standard icon size, so I wouldn't necessarily recommend it

packages/core/src/lib/icon/icon.svelte
anjinai: Thank you MC!

PR #487, opened 2024-02-26
APP-3851: Adjust navigation map spacing

ethanlook:
In R2D2:

image

In legacy config:

image

In control tab:

image

Padding is definitely wonky in the old UI and control tab, but we're about to get rid of the old UI and the control tab already has a lot of jank.


micheal-parks (approved):
👏

PR #488, opened 2024-02-27
add plugin file to build

Kschappacher:
prime core package was missing the plugins.ts file


DTCurrie:
@Kschappacher make sure we use the plugins file in storybook as well.

DTCurrie (approved):
✅

PR #493, opened 2024-03-19
APP-4159 update modal to handle focus differently

Kschappacher:
### Feature
- Change how modal handles focusing when it is opened
- Focus is now handled by the number of slots provided as well as a passed in parameter
- As per the following dialog guides we are removing the heading focusing https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/


packages/core/src/lib/modal.svelte
```diff
@@ -73,15 +75,29 @@ $: if (typeof document !== 'undefined') {
...
+    const slotValues = slotElement?.querySelectorAll('button');
+    if (slotValues?.length !== undefined && slotValues.length > 0) {
+      if (focusPrimaryElement) {
+        slotValues.item(1).focus();
+      } else {
+        slotValues.item(0).focus();
+      }
+    }
+  }
+}
```
mcous, line 92: I'm not sure about adding this logic to the Prime modal; the `querySelector` is a bit of a smell. It looks like our `<Button>` already accepts an `autofocus` prop via `$$restProps`

Can we trim this PR down to remove _all_ focus handling from the `<Modal>` to allow the end-user to make the decision per call-site?

packages/core/src/lib/modal.svelte
```diff
@@ -73,15 +75,29 @@ $: if (typeof document !== 'undefined') {
...
 </script>
 
 <svelte:window on:keydown={isOpen ? handleEscapePress : undefined} />
 
 {#if isOpen}
   <div
     class="fixed left-0 top-0 z-50 flex h-full w-full items-center justify-center bg-black bg-opacity-40"
-    role="dialog"
+    role="alertdialog"
```
mcous, line 100: Should we make this `role` configurable? I think defaulting to `alertdialog` is a bit aggressive; it feels safer to default to `dialog` but allow a called to pass in `role="alertdialog"`

mcous (changes requested):

packages/core/src/lib/modal.svelte
Kschappacher: Yupp good call, added a prop

packages/core/src/lib/modal.svelte
Kschappacher: yeah I think I got a bit too invested in having prime solve all of a users focus cases.

Now that the modal focus logic is gone autofocus totally works in app. 

Removing all the focus stuff

mcous (approved):

PR #494, opened 2024-03-25
APP-3908: ensure all buttons use `<button type="button">`

mcous:
## Overview

Inside a `<form>`, a `<button>` will implicitly become `type="submit"` if not specified. This is never what we want when using Prime components.

## Change log

- Specify `type="button"` for every vanilla `<button>` I could find in core


ehhong (approved):
this one always gets me. thanks for fixing!

anjinai (approved):

PR #496, opened 2024-04-03
Expose on:change event handler on TextInput

ehhong:


packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.99",
+  "version": "0.0.991",
```
ehhong, line 3: how do I upgrade the prime version after `0.0.99`?

mrloureed: `0.0.100` I would think

ehhong: but isn't `0.0.100` < `0.0.99`?

DTCurrie: 0.0.100 is correct.

We will hit 0.1.0 someday.

DTCurrie (approved):
✅ pending version change.

ethanlook (approved):

PR #499, opened 2024-04-08
APP-4319: do not trigger an empty onChange from `SearchableSelect`

mcous:
## Overview

The `<SearchableSelect>` can trigger an empty `onChange` if you blur it while empty. This PR fixes that bug. It looks like this was a previously existing bad behavior rather than a regression from recent work.

## Change log

With `value` separated from `searchValue`, the internal `previousValue` variable is no longer needed. Removing it fixes the bug

## Review requests

- Code changes make sense
- Test additions make sense

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.101",
+  "version": "0.0.102",
   "publishConfig": {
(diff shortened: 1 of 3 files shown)


ethanlook (approved):

zaporter-work (approved):
Was never a big fan of previousValue. Missed this opportunity for cleanup. Thanks!

PR #501, opened 2024-04-09
APP-4460: allow the height of a floating element to be constrained

mcous:
## Overview

We have floating menus that must not overflow the screen. This PR adds various options to our `floating-ui` wrappers to allow the height of a floating element to be constrained.

## Change log

- Add `limitHeight` helper to construct `size` options to limit an element's height to the available space, less some margin
- Allow a floating element to use the `position: fixed` strategy, in case a floating element needs to break out of a clipping element
    - This happens when a floating element is floating off another floating element with a constrained height with `overflow` set
- Add a `<Floating>` component to take extra classnames

## Review requests

I've got this branch `pnpm link`'d up to my local dev environment to verify this works as intended. Code review should suffice here

### DIFF
--- a/packages/core/src/lib/floating/floating.svelte
+++ b/packages/core/src/lib/floating/floating.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-import cx from 'classnames';
+import classnames from 'classnames';
 import { clickOutside } from '$lib/click-outside';
 import {
   floatingStyle,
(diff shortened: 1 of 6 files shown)


packages/core/src/lib/floating/floating.svelte
```diff
@@ -16,12 +17,11 @@ export let offset: number | undefined = undefined;
...
 export let shift: FloatingShiftOptions | undefined = undefined;
 export let size: FloatingSizeOptions | undefined = undefined;
+export let strategy: FloatingStrategy | undefined = undefined;
 export let auto = false;
 export let onClickOutside: ((target: Element) => unknown) | undefined =
   undefined;
-
-let className: cx.Argument = undefined;
-export { className as cx };
+export let cx: classnames.Argument = undefined;
```
ethanlook, line 24: Nice, that's cleaner

ethanlook (approved):

PR #505, opened 2024-04-12
Disable tooltip auto-positioning when not visible

mcous:
## Overview

We don't need to keep tooltip positions up to date if they're not visible. Auto positioning seems relatively expensive (`scroll` listeners, `ResizeObservers`, etc.). This PR turns them off when the tooltip target is not visible

## Review requests

- Auto positioning works when tooltips are visible due to hover
- Auto positioning works when tooltips are visible due to props

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.104",
+  "version": "0.0.105",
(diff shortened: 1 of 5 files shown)


ethanlook (approved):
Awesome. Noticeable improvement? Don't forget to bump the prime version!

anjinai (approved):

PR #513, opened 2024-04-22
[APP-3958] Maplibre component cleanup

micheal-parks:
### Overview

This PR cleans up the `<MapLibre>` component and makes it more flexible for debugging use. I made a quick attempt to unit test but since Maplibre uses WebGL it was becoming more difficult than what it's worth at the moment.

### Changes

* Upgrade MapLibre version.
* Extract map controls to new component for composability.
* Replace events with callback props.
* Remove `!` usage.
* Make `<Marker>` more efficient / easy to use.
* Add example for a map following a marker.

More details in comments.

### DIFF
--- a/packages/blocks/src/lib/maplibre/controls.svelte
+++ b/packages/blocks/src/lib/maplibre/controls.svelte
@@ -1,3 +1,18 @@
+<!-- 
+  @component
+  
+  Adds a set of navigation controls.
+
+  @example
+
+  ```html
+    <MapLibre>
+      <MapLibreControls />
+    </MapLibre>
+  ```
+
+  @see https://maplibre.org/maplibre-gl-js/docs/API/classes/NavigationControl/
+-->
 <script lang="ts">
 import { NavigationControl, type ControlPosition } from 'maplibre-gl';
 import { onMount } from 'svelte';
--- a/packages/blocks/src/lib/maplibre/controls.svelte
+++ b/packages/blocks/src/lib/maplibre/controls.svelte
@@ -1,0 +1,25 @@
+<script lang="ts">
+import { NavigationControl, type ControlPosition } from 'maplibre-gl';
+import { onMount } from 'svelte';
+import { useMapLibre } from './hooks';
+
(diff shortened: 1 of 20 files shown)


packages/blocks/src/lib/maplibre/controls.svelte
```diff
@@ -0,0 +1,40 @@
+<!--
```
micheal-parks, line 1: Extracted out of the `<MapLibre>` component

packages/blocks/src/lib/navigation-map/index.svelte
```diff
@@ -71,7 +71,6 @@ $: $envStore = environment;
 <Map
   bind:map
   {baseGeoPose}
-  on:create
```
micheal-parks, line 74: Not used in app or rdk - so a safe breaking change

packages/blocks/src/routes/following-marker.svelte
```diff
@@ -0,0 +1,39 @@
+<script lang="ts">
```
micheal-parks, line 1: This example is basically how we want to use MapLibre in the movement sensor debug card.

packages/blocks/src/lib/maplibre/index.svelte
```diff
@@ -102,22 +116,20 @@ onMount(() => {
...
   map.on('move', handleMove);
   map.on('resize', handleResize);
   map.on('style.load', handleCreate);
-});
 
-onDestroy(() => {
-  map?.off('move', handleMove);
-  map?.off('resize', handleResize);
-  map?.off('style.load', handleCreate);
+  return () => {
```
micheal-parks, line 128: I've been avoiding `onDestroy` when possible. It runs server side whereas the mount cleanup function does not.

ethanlook (approved):
Awesome! Did you try using this with locally running RDK?

micheal-parks:
Yep! All behavior looks unchanged 👍

zaporter-work (commented):
Bit late to the review, but this is great!

PR #519, opened 2024-05-16
Add progress to icon button

micheal-parks:
Screenshot 2024-05-16 at 12 11 35

### DIFF
--- a/packages/core/src/routes/+page.svelte
+++ b/packages/core/src/routes/+page.svelte
@@ -566,6 +566,13 @@ const onHoverDelayMsInput = (event: Event) => {
     </Button>
 
     <Button progress="indeterminate">Loading</Button>
+    <IconButton
+      label="Pause"
+      variant="secondary"
+      progress="indeterminate"
+      icon="pause"
+      cx="border border-light"
+    />
     <Button
       progress="indeterminate"
       variant="danger">Loading</Button
--- a/packages/storybook/src/stories/icon-button.stories.svelte
+++ b/packages/storybook/src/stories/icon-button.stories.svelte
@@ -35,3 +35,11 @@ import { IconButton } from '@viamrobotics/prime-core';
     disabled
   />
 </Story>
+
+<Story name="Progress">
+  <IconButton
+    icon="refresh"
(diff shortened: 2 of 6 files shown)


packages/storybook/src/stories/icon-button.stories.svelte
```diff
@@ -35,3 +35,11 @@ import { IconButton } from '@viamrobotics/prime-core';
...
   />
 </Story>
+
+<Story name="Progress">
+  <IconButton
+    icon="refresh"
+    progress="indeterminate"
+    label="Restart object"
+  />
+</Story>
```
zaporter-work, line 45: Don't you also have to add a reference to this story to `icon-button.mdx`?

packages/core/src/routes/+page.svelte
```diff
@@ -566,6 +566,13 @@ const onHoverDelayMsInput = (event: Event) => {
     </Button>
 
     <Button progress="indeterminate">Loading</Button>
+    <IconButton
+      label="Pause"
+      variant="secondary"
+      progress="indeterminate"
+      icon="pause"
```
zaporter-work, line 573: Is there a good way to type this so we don't need (or are prevented from) passing an icon if we also pass progress?

This feels surprising -- but not important

zaporter-work (approved):

packages/storybook/src/stories/icon-button.stories.svelte
zaporter-work: (not sure I fully understand storybook)

packages/core/src/routes/+page.svelte
micheal-parks: I get the surprising part, but I don't think that we'd want to make these props mutually exclusive. The most efficient way to use this component is like this because no expensive unmounting / remounting of HTML elements occurs:

```
<script>
  $: progress = loading ? 'indeterminate' : undefined
</script>

<IconButton
  {progress}
  icon='pause'
/>

zaporter-work: Ahh, I hadn't considered that. Neat

PR #520, opened 2024-05-21
App 4656 add ability to truncate breadcrumb items and pills

mrloureed:
- Tailwind's `truncate` class added
- Fix issue where long tooltips get clipped in favor of wrapping

### DIFF
--- a/packages/core/src/lib/breadcrumbs.svelte
+++ b/packages/core/src/lib/breadcrumbs.svelte
@@ -27,7 +27,7 @@ export { extraClasses as cx };
   )}
 >
   {#each crumbs as crumb, index (crumb)}
-    <small class="text-xs">
+    <small class="truncate text-xs">
       {crumb}
     </small>
     {#if index !== crumbs.length - 1}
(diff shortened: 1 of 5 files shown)


packages/core/src/lib/tooltip/tooltip-text.svelte
```diff
@@ -57,7 +57,7 @@ $: style.register({
   />
 
   <div
-    class="flex items-center gap-1 bg-gray-9 px-2 py-1 text-left text-xs text-white"
+    class="flex items-center gap-1 break-all bg-gray-9 px-2 py-1 text-left text-xs text-white"
```
mrloureed, line 60: Prevents clipping when strings are longer than their parent's width and lack spaces or hyphens.

packages/core/src/lib/breadcrumbs.svelte
```diff
@@ -27,7 +27,7 @@ export { extraClasses as cx };
   )}
 >
   {#each crumbs as crumb, index (crumb)}
-    <small class="text-xs">
+    <small class="truncate text-xs">
```
mrloureed, line 30: Add truncate so we can do so where we want to.

anjinai (approved):
lgtm

PR #528, opened 2024-06-06
NumericInput pattern improvements

zaporter-work:
## What

Change pattern for numeric input  & change step to any if not passed. 

## Why

Html pattern is not consistent. Step seems like it is better at marking the field as invalid.

Even with `type="numeric"`, if step==1, the field is marked as invalid if there is a non-1 step size in the input (ex: `1.1`).


packages/core/src/lib/input/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    { val: '1.2-', allowNegative: true, expected: false },
+    { val: '', allowNegative: true, expected: false },
+    { val: ' ', allowNegative: true, expected: false },
+  ])(
+    'validates floating-point numbers ($val)',
+    ({ val, allowNegative, expected }) => {
+      const pattern = getPatternForNumber('number', allowNegative);
+      expect(pattern.test(val)).toBe(expected);
+    }
+  );
```
zaporter-work, line 30: Tests are fairly overkill but easy enough to write & I think they are a good thing to have.

packages/core/src/lib/input/utils.ts
```diff
@@ -12,3 +12,13 @@ export const parseNumericInputValue = (
...
   type === 'number' ? Number.parseFloat(value) : Number.parseInt(value, 10);
+
+export const patternForNumericInput = (
+  type: NumericInputTypes,
+  allowNegative: boolean
+) => {
+  const signPattern = allowNegative ? /[-+]?/u : /\+?/u;
+  return type === 'number'
+    ? new RegExp(`^${signPattern.source}[0-9]+(\\.[0-9]+)?$`, 'u')
+    : new RegExp(`^${signPattern.source}[0-9]+$`, 'u');
```
zaporter-work, line 23: Tried to add comments to these lines, but I cant add comments in the ternary otherwise prettier reformats it and then fails on `line-comment-position`

And I can't break it out into an if-else block because there is also a format lint for prefer-ternary

zaporter-work:
Closing to prioritize other work.

PR #530, opened 2024-06-12
add new link option for pills

mrloureed:
Pill now has an href option

image


peterborges (approved):
LGTM!

packages/core/src/lib/pill.svelte
```diff
@@ -69,7 +72,15 @@ const handleRemove = () => {
       size="sm"
     />
   {/if}
-  <span class="truncate">{value}</span>
+  {#if href}
+    <a
+      {href}
+      target="_blank"
```
micheal-parks, line 78: We should probably make this an export prop too

micheal-parks (approved):

PR #532, opened 2024-06-14
APP-4651: do not render `<SearchableSelect>` list when collapsed

mcous:
## Overview

This PR fixes two performance issues with `<SearchableSelect>` that are having a drastic impact on the app:

- A non-string key is used for the `{#each}` block that renders the options
    - This causes Svelte to very eagerly remove and replace DOM elements during updates
- The menu (and its `{#each}`) are rendered even when the menu is closed - hidden via a `display:hidden`
    - This causes Svelte to run expensive re-render logic when the menu is neither visible nor accessible

This PR fixes these two issues for a noticeable performance improvement in app pages where several `<SearchableSelect>` elements are rendered.

## Review requests

The change in indentation level blew up the diff a little. There are exactly three changes to `<SearchableSelect>`:

- The menu is now wrapped in an `{#if isExapanded}`
- The `<ul>` menu no longer has `class:hidden={!isExpanded}`
- The key of the `{#each}` block has changed from `option` (an object) to `option.value` (a string)

Let me know if you'd like to see a demo of this integrated into the app, but otherwise you can smoke test this in the storybook and playground

### DIFF
--- a/packages/core/src/lib/select/searchable-select.svelte
+++ b/packages/core/src/lib/select/searchable-select.svelte
@@ -28,6 +28,7 @@ import {
   type SortOption,
   type DetailedOption,
 } from './search';
+
 import SelectInput from './select-input.svelte';
 
 /** The options the user should be allowed to search and select from. */
@@ -127,7 +128,7 @@ const resetSearchValue = (
 };
 $: resetSearchValue(value, detailedOptionsMap);
 
-// selectedSeachOption represents the value that was last selected (or the initial value)
+// selectedSearchOption represents the value that was last selected (or the initial value)
 $: selectedSearchOption = detailedOptionsMap[value];
 
 $: searchResults = getSearchResults(detailedOptions, searchValue, sort);
@@ -155,8 +156,8 @@ $: if (menuState === undefined || menuState === FOCUS_SEARCH) {
   );
   // if we don't find any options with a non negative priority,
   // we set the nextAutoSelectIndex to the option that matches
-  // the current seach value. This is used so that when we blur
-  // with a valid seachValue, we autoSelect the correct option
+  // the current search value. This is used so that when we blur
+  // with a valid searchValue, we autoSelect the correct option
   if (nextAutoSelectIndex === -1) {
     nextAutoSelectIndex = allOptions.findIndex(
       ({ option }) => optionDisplayValue(option) === searchValue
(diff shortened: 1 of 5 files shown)


zaporter-work (commented):
Code LGTM, however, when I test this in the playground, clicking to expand the select is scrolling me to the top of the page (on chromium)



https://github.com/viamrobotics/prime/assets/121693134/ab50df0b-76b4-4b49-a1f2-87031e29c783

(firefox refuses to play that video but chromium has no problem)

packages/core/src/lib/use-next-frame.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Add a callback to `requestAnimationFrame`, with cleanup.
+ */
+import { onDestroy } from 'svelte';
+
+export type OnNextFrame = (callback: FrameRequestCallback) => void;
+
+export const useNextFrame = (): OnNextFrame => {
```
zaporter-work, line 8: Cool and interesting pattern.

+1

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -175,9 +178,15 @@ $: activeElement = activeOption
...
   : undefined;
 
-$: if (typeof activeElement?.scrollIntoView === 'function') {
-  activeElement.scrollIntoView({ block: 'nearest' });
-}
+const scrollIntoView = (element: HTMLElement | undefined): void => {
+  if (typeof element?.scrollIntoView === 'function') {
+    onNextFrame(() => {
+      element.scrollIntoView({ block: 'nearest', inline: 'nearest' });
+    });
```
zaporter-work, line 185: [opt] Without reviewing this PR, It would not be intuitive to me why this is required. If you agree that this is not obviously intuitive, a short comment would be good to have.

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -155,8 +158,8 @@ $: if (menuState === undefined || menuState === FOCUS_SEARCH) {
   );
   // if we don't find any options with a non negative priority,
   // we set the nextAutoSelectIndex to the option that matches
-  // the current seach value. This is used so that when we blur
-  // with a valid seachValue, we autoSelect the correct option
+  // the current search value. This is used so that when we blur
+  // with a valid searchValue, we autoSelect the correct option
```
zaporter-work, line 162: I really should add spellcheck to my editor...

zaporter-work (approved):
LGTM Code + manual review.

DTCurrie (approved):
✅

PR #534, opened 2024-06-24
[APP-5332] [APP-4147]: Add 4 more toast styles and update storybook

jgeaso1266:
## Description
Currently there is only a success toast variant. This PR creates 4 more toast variants: warn, danger, info, and neutral in accordance with [Figma](https://www.figma.com/design/ERHSYDGTXiEQOVGY8ORg6I/PRIME?node-id=2650-4380&t=JEKiEAXH14oyuHHr-0). It also makes edits to the existing toast design to be in line with Figma as well. 

**Note:** One test had to be edited to use `getByText` instead of `getByRole` as the aria-label for the action button of the toast banner was removed.

### DIFF
--- a/packages/core/src/lib/toast/__tests__/toast.spec.ts
+++ b/packages/core/src/lib/toast/__tests__/toast.spec.ts
@@ -28,7 +28,7 @@ describe('toast', () => {
 
     const status = screen.getByRole('status');
     const toast = within(status).getByRole('listitem');
-    const icon = screen.getByRole('img', { name: /success/iu });
+    const icon = screen.getByRole('img', { name: /check-circle/iu });
 
     within(toast).getByRole('button', { name: /dismiss/iu });
     expect(toast).toHaveTextContent(/This is a success toast message/iu);
@@ -51,7 +51,9 @@ describe('toast', () => {
 
     const status = screen.getByRole('status');
     const toast = within(status).getByRole('listitem');
-    const actionButton = screen.getByText(/action text/iu);
+    const actionButton = screen.getByRole('button', {
+      name: /action text/iu,
+    });
     await user.click(actionButton);
 
     expect(toast).toHaveTextContent(/action text/iu);
--- a/packages/core/src/lib/toast/__tests__/toast.spec.ts
+++ b/packages/core/src/lib/toast/__tests__/toast.spec.ts
@@ -19,7 +19,7 @@ describe('toast', () => {
     screen.getByRole('status', { name: 'Toasts' });
   });
 
-  it('should display a dismissable toast', async () => {
+  it('should display a dismissible toast', async () => {
     await act(() => {
       context.toast({
         message: 'This is a success toast message',
--- a/packages/core/src/lib/toast/__tests__/toast.spec.ts
+++ b/packages/core/src/lib/toast/__tests__/toast.spec.ts
@@ -51,9 +51,7 @@ describe('toast', () => {
 
     const status = screen.getByRole('status');
     const toast = within(status).getByRole('listitem');
-    const actionButton = screen.getByRole('button', {
-      name: /perform action/iu,
-    });
+    const actionButton = screen.getByText(/action text/iu);
     await user.click(actionButton);
 
     expect(toast).toHaveTextContent(/action text/iu);
--- a/packages/core/src/lib/toast/__tests__/toast.spec.ts
+++ b/packages/core/src/lib/toast/__tests__/toast.spec.ts
@@ -19,7 +19,7 @@ describe('toast', () => {
     screen.getByRole('status', { name: 'Toasts' });
   });
 
-  it('should display a toast with success variant with correct classes and label', async () => {
+  it('should display a dismissable toast', async () => {
     await act(() => {
       context.toast({
         message: 'This is a success toast message',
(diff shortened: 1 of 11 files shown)


packages/core/src/lib/toast/__tests__/toast.spec.ts
```diff
@@ -51,9 +51,7 @@ describe('toast', () => {
 
     const status = screen.getByRole('status');
     const toast = within(status).getByRole('listitem');
-    const actionButton = screen.getByRole('button', {
-      name: /perform action/iu,
-    });
+    const actionButton = screen.getByText(/action text/iu);
```
mcous, line 54: `getByText` should usually be a last resort, or something that you use for finding `<p>` elements or other role-less entities. A `<button>` has a role of `button`, so this should continue to be a `getByRole`

```suggestion
    const actionButton = screen.getByRole('button', {
      name: /action text/iu,
    });
```

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -13,41 +15,80 @@ export let dismiss: () => void;
...
   undefined;
 
+/** The severity of the notification you want to show users*/
+export let variant: ToastVariantType = 'success';
+
 /** Additional CSS classes to pass to the banner. */
 let extraClasses: cx.Argument = '';
 export { extraClasses as cx };
+
+let icon: IconName | null = null;
```
mcous, line 25: We generally avoid `null`; `undefined` is more predictable. See [this article](https://github.com/sindresorhus/meta/discussions/7) for background

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -13,41 +15,80 @@ export let dismiss: () => void;
...
+let iconClasses = '';
+$: {
```
mcous, line 27: Can this functionality of "select icon and text color based on variant" be moved back to `variants.ts`? For bonus points, we could add unit tests to `variants.ts`, too

packages/core/src/lib/toast/toast-banner.svelte
```diff
@@ -13,41 +15,80 @@ export let dismiss: () => void;
...
-      aria-label="success"
-    />
+    {#if icon}
+      <Icon
+        size="lg"
+        name={icon}
+        cx={iconClasses}
+        role="img"
+        aria-hidden={false}
+        aria-label="success"
```
mcous, line 71: This label needs to update according to the variant, too. It makes sense to put this label logic next to the icon and color selection

packages/core/src/lib/toast/__tests__/toast.spec.ts
mcous, line 1: There should be a test that exercises all the variants and checks the accessible names / labels. Alternatively, you could add this a tests for `variants.ts` instead for the same effective coverage

mcous (commented):
Thanks for all the fixes! Looks like we still need to update the accessible text of the banner icon, and we could do with more tests around the variant logic

packages/core/src/lib/toast/__tests__/toast.spec.ts
```diff
@@ -35,6 +36,74 @@ describe('toast', () => {
     expect(icon).toHaveClass('text-success-dark');
   });
 
+  it('should display a toast with info variant with correct classes and label', async () => {
```
mcous, line 39: For future reference, this is a good use case for [`it.each`](https://vitest.dev/api/#test-each). I also think the button and message checking is redundant with other tests. Try to minimize redundant coverage and focus in this test on checking the icon

```ts
it.each([
  {
    variant: ToastVariant.SUCCESS,
    expectedName: /success/iu,
    expectedColor: 'text-success-dark',
  },
  // ...
])('displays correct icon for %variant', ({
  variant,
  expectedName,
  expectedColor,
}) => {
  await act(() => context.toast({ message: 'Hello world' }));

  const status = screen.getByRole('status');
  const toast = within(status).getByRole('listitem');
  const icon = within(toast).getByRole('img', { name: expectedName });

  expect(icon).toHaveClass(expectedColor);
})
```

packages/core/src/lib/toast/variants.ts
```diff
@@ -1,10 +1,29 @@
-import type { IconName } from '$lib/icon/icons';
-
 export const ToastVariant = {
```
mcous, line 1: Avoid combining external details (a toast variant) with internal details (e.g. what tailwind class we apply for a given variant). This should be two objects, on public and one internal:

```ts
export const ToastVariant = {
  SUCCESS: 'success',
  // ...
} as const;

export const DisplayDetailsByVariant = {
  [ToastVariant.SUCCESS]: {
    icon: 'information-outline',
    iconClasses: 'text-info-dark',
    label: 'info',
  },
  // ...
} as const;
```

packages/storybook/src/stories/use-toast.mdx
```diff
@@ -32,19 +32,21 @@ import PageContent from '$lib/layout/page-content.svelte';
...
+To display a toast, first provide access to the Toast context using `provideToast`. Then, you can use the `useToast` function:
 
 ```ts
 import { provideToast } from '@viamrobotics/prime-core';
 
 const doSomethingCrazy = () => {
   const toast = provideToast();
 
-  toast({ message: 'Hello, Brooklyn!', variant: ToastVariant.Success });
+  toast({ message: 'Hello, Brooklyn!', variant: 'success' });
```
mcous, line 43: Can you undo this change? Generally, we want to avoid magic string APIs

packages/core/src/lib/toast/variants.ts
```diff
@@ -1,29 +1,37 @@
...
     iconClasses: 'text-success-dark',
     label: 'success',
   },
-  Neutral: {
+  [ToastVariant.Neutral]: {
     icon: 'domain',
     iconClasses: 'text-gray-7',
     label: 'neutral',
   },
-} as const;
```
mcous, line 5: [Nit] I'd keep this `as const`, if possible

packages/core/src/routes/+page.svelte
```diff
@@ -1708,6 +1708,20 @@ const onHoverDelayMsInput = (event: Event) => {
...
+      action={{
+        text: 'Undo',
+        handler: () => {
+          console.log('Clicked action');
+        },
+      }}
+      dismiss={() => {
+        console.log('Clicked close button');
+      }}
+      cx="max-w-[240px]"
```
mcous, line 1723: [Nit] As discussed in person, this `max-w-[240px]` conflicts with the banners own class of `max-w-[480px]`. This is a little confusing, so we probably shouldn't add this class in the playground

packages/core/src/lib/toast/__tests__/toast.spec.ts
```diff
@@ -128,4 +60,24 @@ describe('toast', () => {
...
     expect(actionHandler).toHaveBeenCalled();
   });
+
+  it.each([
+    {
+      variant: ToastVariant.Success,
+      expectedName: /success/iu,
+      expectedColor: 'text-success-dark',
+    },
+    // ...
```
mcous, line 70: Discussed in Slack: make sure this gets filled out

mcous (approved):
Thanks for taking this on

anjinai (approved):

PR #537, opened 2024-07-03
Bump js-config to get new pnpm/actions-setup

zaporter-work:
See: https://github.com/viamrobotics/js-config/pull/43


.github/actions/setup/action.yml
```diff
@@ -17,7 +17,7 @@ runs:
   using: 'composite'
   steps:
     - name: 'Install pnpm ${{ inputs.pnpm-version }}'
-      uses: pnpm/action-setup@c3b53f6a16e57305370b4ae5a540c2077a1d50dd # v2.2.4
+      uses: pnpm/action-setup@23657c8550e9be4f7ee7e9d7fe8adda6d703bb4e # v4.0.0
```
zaporter-work, line 20: @ethanlook Do you know why prime has an almost-copy of the action from js-config?

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Install 🔧
-        uses: viamrobotics/js-config/.github/actions/setup@0006127efb3f62669f662ca31c555a49a034935e
+        uses: viamrobotics/js-config/.github/actions/setup@7e4e4f1bd14c83aaec2fd877324c38028d8a387b
```
zaporter-work, line 18: [opt] swap this to be unpinned like app.

.github/actions/setup/action.yml
ethanlook: My best guess is that we were copy-pasting the action around, and hadn't gotten to updating prime to use the shared jsconfig.

.github/workflows/ci.yml
ethanlook: My opinion is that we should err on the side of consistency.

ethanlook (approved):

.github/workflows/ci.yml
zaporter-work: :+1: Changed to pin to `@main`

.github/actions/setup/action.yml
ehhong: should we update prime to use the shared jsconfig here?

zaporter-work: Could -- here are the differences:

```diff
7c7
<     default: '18'
---
>     default: '20'
10c10,11
<     description: Registry to configure for publish
---
>     description: 'Registry URL to set'
>     required: false
13,14c14,16
<     description: 'Run `pnpm install`'
<     default: 'true'
---
>     description: 'Run pnpm install to install dependencies'
>     type: boolean
>     default: true
19c21
<     - name: 'Install pnpm ${{ inputs.pnpm-version }}'
---
>     - name: 'Install pnpm'
32c34
<       run: pnpm install --frozen-lockfile
---
>       run: pnpm install
```

Would probably have to do a bit more testing before I drop it in

ehhong: the only significant change seems to be the node version, feel like it shouldn't be too bad?

zaporter-work: @ehhong Swapped in https://github.com/viamrobotics/prime/pull/537/commits/405c7ec8b6b2fd958740ec822df61b39d9ed2528

zaporter-work: This also preserves node 18.

ethanlook: I had typed this up, then Zack told me "make sure to reload before you comment." Using the shared config is the right long-term move, so we can go ahead with this.

> From Ethan past:

My instinct is that we should we should move forward with the change as-is, and then follow-up to use the shared js-config action. It's probably a safe change to start using the shared action, but would be easier if we could roll it back independently and not break PRIME CI.

PR #542, opened 2024-07-23
APP-5548: add toggle icons, fix icon name type-checking

mcous:
- Add `toggle-switch-outline` and `toggle-switch-off-outline` icons
- Ensure icon names are strictly typed, not `string`

### DIFF
--- a/packages/core/src/lib/icon/icons.ts
+++ b/packages/core/src/lib/icon/icons.ts
@@ -13,7 +13,7 @@ export type IconPath = Path | CustomIcon[];
  * Keys should match MDI name
  * e.g. 'account-multiple' for MDI.mdiAccountMultiple
  */
-export const paths: Record<string, IconPath> = {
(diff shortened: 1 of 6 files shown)


packages/core/src/lib/icon/icons.ts
```diff
@@ -13,7 +13,7 @@ export type IconPath = Path | CustomIcon[];
  * Keys should match MDI name
  * e.g. 'account-multiple' for MDI.mdiAccountMultiple
  */
-export const paths: Record<string, IconPath> = {
```
mcous, line 16: This `Record<string, IconPath>` overrode the `as const` below and broke strict icon name type-checking, setting `IconName` to `string` rather than the literal keys of this object

ethanlook: Nice catch!

ethanlook (approved):

PR #545, opened 2024-07-24
[RSDK-7089] Orientation format fix, add LengthCapsuleGeometry

micheal-parks:
### Overview

Funny we mentioned a degrees / radians issue in this week's meeting, cause here I am fixing one I likely caused!

Fixes [RSDK-7089](https://viam.atlassian.net/browse/RSDK-7089).

The solution I've implemented here is pretty much what I've been doing to curtail this problem elsewhere: only use degrees at the surface level - UI layer - and always use radians at the data layer.

I've also done a flyby blocks export addition: `LengthCapsuleGeometry`. This represents Viam's capsule definition, which differs slightly from the Three.js definition. We should be using it everywhere but until now this definition is only in the frame tab. I've applied it to the navigation map, so capsules now have a correct length that matches what the robot thinks.

### Next steps

I think some unit test here are appropriate but I'd like to bundle that work into the unit testing I promised for the other new exports I've recently fabricated.

[RSDK-7089]: https://viam.atlassian.net/browse/RSDK-7089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/blocks/src/lib/navigation-map/components/nav/obstacles.svelte
```diff
@@ -70,10 +70,9 @@ const handleGeometryInput =
   };
 
 const handleOrientationInput =
-  (name: string, geoIndex: number) => (event: CustomEvent<number>) => {
+  (name: string, geoIndex: number) => (value: number) => {
     const index = $obstacles.findIndex((obstacle) => obstacle.name === name);
-    $obstacles[index]!.geometries[geoIndex]!.pose.orientationVector.th =
-      event.detail;
+    $obstacles[index]!.geometries[geoIndex]!.pose.orientationVector.th = value;
```
ethanlook, line 75: nit: Could we avoid using the assertion operator and wrap this in an if statement?

ethanlook (approved):

packages/blocks/src/lib/navigation-map/components/nav/obstacles.svelte
micheal-parks: Yeah, there's a buuuunch of these in blocks. I haven't added any new null assertions, mind if I instead remove them in the new navigation map in app? We'll be able to delete this directory shortly after and just wholesale remove them.

ethanlook: Works for me!

PR #550, opened 2024-08-01
Test: add `createToastContext` fixture

mcous:
## Overview

Bring the "no fixture component needed" context testing model to Prime for toasts

## Review requests

Any additional documentation that you think would be helpful?

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.138",
+  "version": "0.0.139",
   "publishConfig": {
     "access": "public"
   },
@@ -23,6 +23,10 @@
     ".": {
       "types": "./dist/index.d.ts",
       "svelte": "./dist/index.js"
+    },
+    "./__fixtures__": {
+      "types": "./dist/__fixtures__/index.d.ts",
+      "svelte": "./dist/__fixtures__/index.js"
     },
     "./prime.css": "./prime.css",
     "./plugins": "./plugins.ts",
--- a/packages/core/src/lib/__fixtures__/index.ts
+++ b/packages/core/src/lib/__fixtures__/index.ts
@@ -1,0 +1,40 @@
+/** prime-core test fixtures. */
+import { readable, writable } from 'svelte/store';
+import { noop } from 'lodash-es';
+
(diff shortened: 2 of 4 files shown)


Kschappacher (approved):
This is awesome, we should be able to clean up a ton of .spec.svelte files in app

packages/core/package.json
```diff
@@ -24,6 +24,10 @@
       "types": "./dist/index.d.ts",
       "svelte": "./dist/index.js"
     },
+    "./__fixtures__": {
+      "types": "./dist/__fixtures__/index.d.ts",
+      "svelte": "./dist/__fixtures__/index.js"
+    },
```
ethanlook, line 30: I think a little README documentation about this export would be nice if you don't mind!

packages/core/src/lib/__fixtures__/index.ts
```diff
@@ -0,0 +1,40 @@
...
+ * import Subject from '../cool-component.svelte';
+ *
+ * const toast = vi.fn();
+ * const renderSubject = () => {
+ *   const toastContext = createToastContext(toast)
+ *   return render(Subject, { props: {}, context: new Map([toastContext]) });
+ * }
+ * ```
+ */
+export const createToastContext = (toast?: Toast): [unknown, unknown] => [
```
ethanlook, line 34: Your call, but typically we've named this `createNoopToastContext`?

ethanlook (approved):
:shipit:

packages/core/src/lib/__fixtures__/index.ts
```diff
@@ -0,0 +1,40 @@
...
+ * import { render } from '@testing-library/svelte';
+ *
+ * import { createToastContext } from '@viamrobotics/prime-core/__fixtures__';
+ *
+ * import Subject from '../cool-component.svelte';
+ *
+ * const toast = vi.fn();
+ * const renderSubject = () => {
+ *   const toastContext = createToastContext(toast)
+ *   return render(Subject, { props: {}, context: new Map([toastContext]) });
```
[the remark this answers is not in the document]
mcous: Nope, the `toastContext` returned from `createToastContext` is a tuple, so this comes out as:

```ts
context: new Map([
  // toastContext
  [ToastContextKey, /* created context */]
])

```

PR #553, opened 2024-08-12
APP-5855 - Move JSON editors to PRIME

zaporter-work:
Ticket: https://viam.atlassian.net/browse/APP-5855

Creates viamrobotics/prime-editor -> a new package containing JSONDiff and JSONEditor.

Talked with @ethanlook offline and due to codemirror's csr requirements & size, it seems reasonable to move this to its own package.

Still new to this. I fully expect a few iteration cycles on this PR and am happy to go back and forth on naming details (<i>or realize we actually want these in prime-core and close+re-open this PR</i>)

This PR looks big, but 800 lines are in pnpm-lock & most is copy-over. 

<strong>Do we want to wait for Micheal Cousins to get back from vacation? I can just copy-paste some of code into svelte-sdk as a temp solution. </strong>

PR #554, opened 2024-08-12
Dedupe packages 

zaporter-work:
While working on https://github.com/viamrobotics/prime/pull/553, I noticed that our packages have drifted quite a bit and that pnpm-dedupe produces a fairly large diff (6k loc).

Any reason why we shouldn't dedupe? We did this in UI and helped fix a few hard-to-track down bugs (I suspect that is helping cause https://github.com/viamrobotics/prime/actions/runs/10357912623/job/28670954306 (but I suspect that is primarily due to bad dev constraints))

~As a result of this, a few of the legacy stories started producing compilation warnings. Instead of fixing, I just deleted all of them. Please let me know if that is the wrong thing to do.~

### ~TODO(pre-merge) bump versions of all our packages.~ Done in 4eccac5a

(If anyone wants additional reviewers, feel free to add them)


packages/blocks/src/lib/navigation-map/components/nav/obstacles.svelte
```diff
@@ -183,9 +183,7 @@ $: debugMode = $environment === 'debug';
   <li
     class="group sticky bottom-0 z-10 bg-white pt-4"
     on:mouseenter={() => {
-      if (selectedObstacle) {
-        $hovered = selectedObstacle.name;
-      }
+      $hovered = selectedObstacle.name;
```
zaporter-work, line 186: unnecessary-conditional was firing -- this is wrapped in a `{#if selectedObstacle}`

micheal-parks (approved):
Talked offline. This seems safe enough and is blocking so 🚀

packages/blocks/package.json
ethanlook, line 1: We should also clean up the dependencies in `prime-core` and `prime-blocks`:
- Use the same version of each package (e.g. `prime-core.dependencies.svelte === prime-blocks.dependencies.svelte`
- Use proper `peerDependencies` and `devDependencies`
- Then `pnpm dedupe`

Keeping the versions and `devDependencies` vs `dependencies` consistent will result in the best deduplication.

We should probably do this in `prime-legacy` too, but if there's significant drift in dependencies we can ignore it. I feel less strongly about keeping legacy in a good state, especially because blocks and core do not depend on legacy.

We can probably ignore `prime-storybook` because that's going away. @micheal-parks @mrloureed when is that going to happen?

ethanlook (approved):
For potentially doing a rollback - I have a slight preference that we stack the PRs.

1. Remove legacy storybook
2. Dedupe packages

Otherwise LGTM

packages/blocks/package.json
zaporter-work: Planning to leave this for a future PR if that is okay? Messing with core deps resulted in more changed versions which ended up creating linting additional changes.

Trying to get in a good-enough state that editor builds & checks correctly.

Agreed that investing in fixing deps is a good use of our time.

ethanlook: Works for me! Thank you for investigating.

PR #563, opened 2024-08-21
add modal medium variant, add export icon

mrloureed:
image

### DIFF
--- a/packages/core/src/lib/modal.svelte
+++ b/packages/core/src/lib/modal.svelte
@@ -50,7 +50,7 @@ import { clickOutside } from '$lib';
 export let isOpen: boolean;
 
 /** The variant of the modal. */
-export let variant: 'small' | '' = '';
+export let variant: 'small' | 'medium' | '' = '';
 
 export let role = 'dialog';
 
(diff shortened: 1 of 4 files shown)


packages/core/src/lib/modal.svelte
```diff
@@ -85,9 +85,10 @@ $: if (typeof document !== 'undefined') {
     <div
       use:clickOutside={handleCloseModal}
       class={cx(
-        'relative max-w-lg border border-medium bg-white p-6 shadow-sm',
+        'relative border border-medium bg-white p-6 shadow-sm',
         {
-          'max-h-[600px] min-h-[400px] w-[480px]': variant === '',
+          'h-[400px] w-[480px]': variant === '',
+          'h-[600px] w-[800px]': variant === 'medium',
```
micheal-parks, line 91: can we do `w-full h-full max-h-[600px] max-w-[800px]` instead? If that doesn't break things it'd be more responsive.

micheal-parks (approved):

packages/core/src/lib/modal.svelte
mrloureed: Actually these should just be setting width not max width; fixing

PR #570, opened 2024-08-29
Fix: s/overflow-y-scroll/overflow-y-auto/

mcous:
## Overview

Drive-by fix for phantom scrollbars in the navigation service card

Screenshot 2024-08-29 at 14 03 36


micheal-parks (approved):
Thank you!

PR #577, opened 2024-09-17
APP-6256: allow searchable select to flip if needed

mcous:
See APP-6256. Accidentally fixed the issue while investigating


anjinai (approved):

mrloureed (approved):
LGTM

PR #580, opened 2024-09-19
Break words in SearchableSelect value & description

zaporter-work:
I know we are deprecating `SearchableSelect`. However, the `video-path` dropdown for the webcam component builder has caused me enough emotional pain to just come fix this. Thoughts?

Before:
before

After:
20240919_01h54m16s_grim

### DIFF
--- a/packages/core/src/lib/select/searchable-select.svelte
+++ b/packages/core/src/lib/select/searchable-select.svelte
@@ -404,8 +404,11 @@ const handleKeydown = createHandleKey({
                 name={option.icon}
               />
             {/if}
-            <div class="flex flex-col">
-              <p class="text-wrap text-sm">
+            <div
+              class="flex flex-col"
+              style="overflow-wrap: anywhere;"
+            >
+              <p class="text-sm">
                 {#if highlight !== undefined}
                   {@const [prefix, match, suffix] = highlight}
                   {prefix}<span class="bg-yellow-100">{match}</span>{suffix}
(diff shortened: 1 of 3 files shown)


packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -404,8 +404,11 @@ const handleKeydown = createHandleKey({
                 name={option.icon}
               />
             {/if}
-            <div class="flex flex-col">
-              <p class="text-wrap text-sm">
+            <div
+              class="flex flex-col"
+              style="overflow-wrap: anywhere;"
```
zaporter-work, line 409: Tailwind hasn't merged an alternative to `word-break: break-word`.

See discussion: https://github.com/tailwindlabs/tailwindcss/discussions/12127
See PR: https://github.com/tailwindlabs/tailwindcss/pull/12128

packages/core/src/lib/select/searchable-select.svelte
```diff
@@ -404,8 +404,11 @@ const handleKeydown = createHandleKey({
...
+            >
+              <p class="text-sm">
```
zaporter-work, line 411: text-wrap not required.

packages/core/src/lib/select/searchable-select.svelte
mcous: I'm cool with this, but can we do it as a tailwind arbitrary-inside-brackets class instead of an inline style?

zaporter-work: :+1: 53735e13

PR #583, opened 2024-09-27
SearchableSelect LabelledBy Attribute

zaporter-work:
Didn't add a test because I don't know a good way to add phony elements to the screen in vitest+svelte. (without a SearchableSelect.spec.svelte)

Is there a good way to do this?

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.155",
+  "version": "0.0.156",
   "repository": {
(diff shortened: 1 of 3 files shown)


mcous (approved):

mcous:
> Didn't add a test because I don't know a good way to add phony elements to the screen in vitest+svelte. (without a SearchableSelect.spec.svelte)
>
> Is there a good way to do this?

Talked IRL; there is not a good way to do this, unfortunately. Presumably, this behavior will be important for a test of a larger component, so we should get sufficient coverage, just at a distance

PR #585, opened 2024-10-04
APP-6324: Make ToggleButtons purely presentational

ethanlook:
For teleop, `ToggleButtons` get used for page navigation. When the user flips from `Edit -> Monitor`, the workspace gets saved. If there is an error while saving, we do not want to navigate to `Monitor`. Because the `ToggleButtons` set internal state, it appears as if toggling was successful. This makes the `ToggleButtons` purely presentational so that they only update when the parent changes the `selected` prop.

This will certainly be a breaking change somewhere when we bring it into App. I will do this PRIME upgrade myself and be sure to check all instances.

## Change log

- Remove internal `selected` state from `ToggleButtons`
- Update tests
- Update storybook

## TODO

- [ ] @mrloureed where will we update the corresponding design.viam.com page?

### DIFF
--- a/packages/core/src/lib/toggle-buttons.svelte
+++ b/packages/core/src/lib/toggle-buttons.svelte
@@ -51,7 +51,6 @@ $: getButtonClasses = (option: string) => {
 };
 
 const handleClick = (value: string) => {
-  selected = value;
   dispatch('input', value);
 };
 </script>
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.157",
+  "version": "0.0.158",
   "repository": {
     "type": "git",
     "url": "https://github.com/viamrobotics/prime.git",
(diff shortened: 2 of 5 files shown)


micheal-parks (approved):

zaporter-work (approved):
> This will certainly be a breaking change somewhere when we bring it into App. I will do this PRIME upgrade myself and be sure to check all instances.

SGTM. I wonder if tests will also need updates.

packages/core/src/lib/toggle-buttons.svelte
```diff
@@ -51,7 +51,6 @@ $: getButtonClasses = (option: string) => {
 };
 
 const handleClick = (value: string) => {
-  selected = value;
```
zaporter-work, line 54: I wonder if the caller could also put `selected` in a store and then when they handle the input event, if they could just set $selected back to the original value (which IIRC would still trigger an update to this component even though the value is the same).

It might flash, but I think the transition would be quick.

ethanlook: Also an option, but this behavior is weird and I think the component should be purely presentational instead of mixing stateful concerns.

ethanlook:
@mrloureed is OOO - I will sync up with him about design.viam.com when he's back in office and keep this moving!

PR #591, opened 2024-10-24
App 6746 prime core toggle buttons

mrloureed:
`Toggle-Button`

- Don’t have a “legend” – use labels
- Default w-fit instead of w-full


mrloureed:
Bueller? @micheal-parks

micheal-parks (approved):
I live!

PR #593, opened 2024-10-28
Add button link as styles function, introduce tailwind-merge

DTCurrie:
Proposing an idea here. We want to move the button link from `app` to `prime`, but after some recent conversations, I wonder if we instead just want to export a function to generate the styles for a button link. It will allow us to use the tools already built to support a native `<a />` element, including things like autocompleting for `rel` and `target`, which our button link implementation limits. It also creates pretty easy-to-test little units.

Another thing I introduced was [`tailwind-merge`](https://github.com/dcastil/tailwind-merge), which works a lot like `classnames` but also merges tailwind classes down to avoid conflicting styles. With the way we use `extraClasses` this allows an easy way to actually override styles without having to use important tags.

Screenshot 2024-10-28 at 3 35 36 PM

Screenshot 2024-10-28 at 3 37 25 PM

Screenshot 2024-10-28 at 3 37 49 PM

### DIFF
--- a/packages/core/prime.css
+++ b/packages/core/prime.css
@@ -15,3 +15,16 @@ button {
 :host([hidden]) {
   display: none;
 }
+
+/* Just give me the CSS classes option */
+.button-link {
+  @apply flex h-7.5 items-center gap-1.5 whitespace-nowrap px-3 py-1.5 text-xs;
+}
+
+.button-link-ghost {
+  @apply button-link text-default hover:bg-ghost-light active:bg-ghost-medium;
+}
+
+.button-link-dark {
+  @apply button-link border border-gray-9 bg-gray-9 text-white hover:border-black hover:bg-black active:bg-[#000];
+}
--- a/packages/core/src/routes/+page.svelte
+++ b/packages/core/src/routes/+page.svelte
@@ -49,7 +49,10 @@ import {
   TooltipTarget,
   TooltipText,
   VectorInput,
+  BUTTON_LINK_GHOST,
+  BUTTON_LINK_DARK,
 } from '$lib';
+import { twMerge } from 'tailwind-merge';
 
 provideNotify();
 
@@ -466,13 +469,14 @@ const onHoverDelayMsInput = (event: Event) => {
 <NotificationContainer />
 
 <div class="container mx-auto my-4 flex flex-col gap-8 p-4">
-  <h2 class="h-2">New stuff!</h2>
+  <h2 class="h-2">With a function</h2>
 
   <div class="flex items-center gap-2">
     <p class="text-xs">Check out this cool thing:</p>
     <a
       href="https://design.viam.com"
       target="_blank"
+      rel=""
       class={buttonLinkStyles({})}
     >
       My button link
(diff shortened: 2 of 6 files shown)


packages/core/src/routes/+page.svelte
```diff
@@ -1,54 +1,55 @@
...
-  CodeSnippet,
-  RangeInput,
-  Progress,
+  Tooltip,
+  TooltipContainer,
+  TooltipTarget,
+  TooltipText,
+  VectorInput,
 } from '$lib';
-import { uniqueId } from 'lodash-es';
```
DTCurrie, line 53: Just some import sorting.

packages/core/src/routes/+page.svelte
```diff
@@ -193,7 +194,6 @@ console.log(\`The FizzBuzz sequence for n = 15 is:\`);
 console.log(sequence); // Outputs: ["1", "2", "Fizz", "4", "Buzz", "Fizz", "7", "8", "Fizz", "Buzz", "11", "Fizz", "13", "14", "FizzBuzz"]`.trim();
 
 const goSnippet = `
-import "fmt"
```
DTCurrie, line 196: The linter actually hates this, will fix later.

packages/core/src/routes/+page.svelte
```diff
@@ -465,6 +465,44 @@ const onHoverDelayMsInput = (event: Event) => {
...
+    <a
+      href="https://design.viam.com"
+      target="_blank"
+      class={buttonLinkStyles({ extraClasses: 'text-pink-500' })}
+    >
+      My button link
+    </a>
+  </div>
+</div>
+
```
DTCurrie, line 505: I will create a better example if we decide to go this way.

pnpm-lock.yaml
ethanlookpotts, line 1: Haven't reviewed the rest of this PR yet but I wouldn't expect a 22k line diff for the pnpm-lock. `pnpm dedupe`?

mcous (commented):
I am very much in favor of this change, but I think for now, we should avoid `tailwind-merge`. From [their docs](https://github.com/dcastil/tailwind-merge/blob/v2.5.4/docs/when-and-how-to-use-it.md#reasons-not-to-use-it):

> ## Reasons not to use it
> Generally speaking, there are situations where you could use tailwind-merge but probably shouldn't. Think of tailwind-merge as an escape hatch rather than the primary tool to handle style variants.[1](https://github.com/dcastil/tailwind-merge/blob/v2.5.4/docs/when-and-how-to-use-it.md#user-content-fn-simonswiss-quote-c2ae99762253620e6184de9d3da43aca)

I think for now we should keep Prime on `classnames`, and add `tailwind-merge` to the places where we need to do overrides. I imagine then that we can look at our `twMerge` callsites and use that to refine the APIs of our style functions

pnpm-lock.yaml
DTCurrie: Ran it, still getting those changes. Seems like it is bumping the lockfile version from `6.0` to `9.0` and doing any allowed minor/patch bumps on all dependencies. I freshly cloned the repo and installed so may be related to that.

DTCurrie:
> I am very much in favor of this change, but I think for now, we should avoid `tailwind-merge`. From [their docs](https://github.com/dcastil/tailwind-merge/blob/v2.5.4/docs/when-and-how-to-use-it.md#reasons-not-to-use-it):
> 
> > ## Reasons not to use it
> > Generally speaking, there are situations where you could use tailwind-merge but probably shouldn't. Think of tailwind-merge as an escape hatch rather than the primary tool to handle style variants.[1](https://github.com/dcastil/tailwind-merge/blob/v2.5.4/docs/when-and-how-to-use-it.md#user-content-fn-simonswiss-quote-c2ae99762253620e6184de9d3da43aca)
> 
> I think for now we should keep Prime on `classnames`, and add `tailwind-merge` to the places where we need to do overrides. I imagine then that we can look at our `twMerge` callsites and use that to refine the APIs of our style functions

I am okay with not doing it for now, but buttons and inputs are probably our most commonly overwritten components style-wise. It is hard to find all of the cases where we are using the important operator in use since the `!` is so common.

mcous:
> I am okay with not doing it for now, but buttons and inputs are probably our most commonly overwritten components style-wise. It is hard to find all of the cases where we are using the important operator in use since the `!` is so common.

Yeah definitely feeling the pain, and I think replacing `!` usage with `twMerge` is a great idea, so I'm glad you've brought the library up. My only thought is that I think we could keep the escape hatch confined to our application code rather than our component library, so we can use it as a design signal to improve the component library

e.g. If we see `<button class={twMerge(buttonStyles(), 'h-7.5')} >`, we keep our fast velocity but also see that our `buttonStyles` function needs to grow a height option

DTCurrie:
> > I am okay with not doing it for now, but buttons and inputs are probably our most commonly overwritten components style-wise. It is hard to find all of the cases where we are using the important operator in use since the `!` is so common.
> 
> Yeah definitely feeling the pain, and I think replacing `!` usage with `twMerge` is a great idea, so I'm glad you've brought the library up. My only thought is that I think we could keep the escape hatch confined to our application code rather than our component library, so we can use it as a design signal to improve the component library
> 
> e.g. If we see `<button class={twMerge(buttonStyles(), 'h-7.5')} >`, we keep our fast velocity but also see that our `buttonStyles` function needs to grow a height option

Yeah that's fair.

packages/core/prime.css
```diff
@@ -15,3 +15,16 @@ button {
 :host([hidden]) {
   display: none;
 }
+
+/* Just give me the CSS classes option */
+.button-link {
```
micheal-parks, line 20: it doesn't look like anything here is related to `link`, should we just call this `.button`? My assumption when looking at this name is that it's a special collection for link buttons, when it's just the primary button styles.

PR #594, opened 2024-10-30
added z-index to useNotify banner

joycexjiang:
Screenshot 2024-10-30 at 12 09 31 PM

Banner is now shown above objects instead of below.


micheal-parks (approved):

mrloureed (approved):
LGTM!

PR #597, opened 2024-11-01
Add iconCx to Pill

zaporter-work:
Need to change the styles of the icon within a Pill for the teleop part-status pill [see ticket with figma link](https://viam.atlassian.net/browse/APP-6325)


ethanlookpotts (approved):
We need the ability to have a tooltip on the icon as well - is that possible with the pill today?

zaporter-work:
@ethanlookpotts Nope! Good catch. I think that is a good argument for forking the `<Pill>`. Unfortunately, you cannot conditionally define the tooltip description so the best I could come up with is:

```svelte
  {#if icon}
    {#if iconTooltip}
    <Tooltip let:tooltipID>
      <Icon
        aria-described-by={tooltipID}
        name={icon}
        cx={cx('text-gray-6', extraIconClasses)}
        size="sm"
      />
      <p slot="description">{iconTooltip}</p>
    </Tooltip>
    {:else}
      <Icon
        name={icon}
        cx={cx('text-gray-6', extraIconClasses)}
        size="sm"
      />
    {/if}
  {/if}

```

(or some slot with a default value but that messes with the rest of the component.)

Please let me know if you had something else in mind before I close this and bring a small subset of Pill to app.

ethanlookpotts:
@zaporter-work we should avoid forking the Pill implementation if possible - we don't want to get into a state of needed to manage two sources for the same component.

@mrloureed @micheal-parks thoughts on adding a tooltip to Pill? Do you like [Zack's suggestion](https://github.com/viamrobotics/prime/pull/597#issuecomment-2455162044)?

ethanlookpotts (approved):

zaporter-work:
Thanks @ehhong . Swapping to using the Tooltip components themselves works much better.

PR #599, opened 2024-11-12
Fix: allow pill anchor target to remain unspecified

mcous:
## Overview

- The `<Pill>` component can take an `href`
- If an `href` is provided, the `target` attribute of the anchor defaults to `_blank`
- If a consumer does not want to open the link in a new tab, they add `target="_self"`
- `_self` causes SvelteKit to trigger a full page reload instead of client-side routing
    - Intentional, but undocumented behavior from SK, according to an old GH thread in Sapper
    - See how clicking on parts and machines from the "All Machines Dashboard" reloads the page instead of preloading and rendering in

This PR removes the default `_blank` and sets it to `undefined`, which will be a more reliable default. A couple call-sites will need to be updated, but otherwise most existing call-sites that rely on the `_blank` default already specify it explicitly.

## Review requests

Check that above reasoning makes sense

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/prime-core",
-  "version": "0.0.165",
+  "version": "0.0.166",
   "repository": {
     "type": "git",
     "url": "https://github.com/viamrobotics/prime.git",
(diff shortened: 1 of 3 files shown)


mrloureed (approved):
Makes sense. LGTM.

ethanlookpotts (approved):

PR #600, opened 2024-11-12
Prepare Threlte for Svelte 5 upgrade

micheal-parks:
### Overview

This PR gets rid of Threlte imports that have been removed in v8 - the Svelte 5 compatible version.

This is part of a larger slow effort to move us to be ready to upgrade to Svelte 5.

There's actually only one removed import here, and it has to do with event dispatching. Callback props ftw!

### DIFF
--- a/packages/blocks/package.json
+++ b/packages/blocks/package.json
@@ -81,12 +81,12 @@
     "vitest": "^1.1.0"
   },
   "peerDependencies": {
-    "@threlte/core": ">=7 <8",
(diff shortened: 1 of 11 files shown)


ethanlookpotts (approved):
Love callback props! Thanks for keeping an eye on this! Two thoughts:

1) Do you mind following this PR up with a bump in App just so we don't forget about the changes here and have version drift?
2) Can we think of ways to increase test coverage over these components? There's a lot of critical navigation map infra that is going untested at the moment.

PR #605, opened 2024-12-13
APP-7173: Add Prism code highlighting action

mcous:
@ehhong and I noodling on Prism stuff

---
### Overview
Creating a Svelte action to handle code highlighting with Prism to modularize Prism logic and reuse logic between markdown component (in app) and code snippet element (in Prime).

### DIFF
--- a/packages/core/src/lib/code-snippet.svelte
+++ b/packages/core/src/lib/code-snippet.svelte
@@ -17,12 +17,6 @@ https://prismjs.com
   context="module"
   lang="ts"
 >
-import Prism from 'prismjs';
-import PrismPackage from 'prismjs/package.json';
-import 'prism-themes/themes/prism-vs.min.css';
-
-const PRISM_VERSION = PrismPackage.version;
-
 type CodeSnippetCopyState = 'copy' | 'copied' | 'failed';
 const COPY_STATES: Record<
   CodeSnippetCopyState,
@@ -35,11 +29,11 @@ const COPY_STATES: Record<
 </script>
 
 <script lang="ts">
-import { createEventDispatcher, onMount } from 'svelte';
+import { createEventDispatcher } from 'svelte';
 import cx from 'classnames';
 
 import { IconButton, type IconName, useTimeout } from '$lib';
-import { escape } from 'lodash-es';
+import { highlightCode } from './highlight-code/highlight-code';
 
 /**
  * The language to use for syntax highlighting. Must be a language supported by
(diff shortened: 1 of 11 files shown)


packages/core/src/lib/highlight-code/prism.ts
```diff
@@ -0,0 +1,25 @@
...
+ * default path for the grammar files is a CDN link so we don't have to include
+ * grammar files in our bundle. If you prefer to point to your own grammars and
+ * bypass the CDN, define the path to those files.
+ *
+ * See: https://prismjs.com/plugins/autoloader/
+ */
+import 'prismjs/plugins/autoloader/prism-autoloader';
+const grammarsPath = `https://cdnjs.cloudflare.com/ajax/libs/prism/${PrismPackage.version}/components/`;
+
+export const getPrismModule = (): typeof import('prismjs') => {
```
mcous, line 20: Since we're importing it already, I think this can be `typeof Prism`


```suggestion
export const getPrismModule = (): typeof Prism => {
```

packages/core/src/lib/code-snippet.svelte
```diff
@@ -157,8 +102,9 @@ onMount(async () => {
 
   <div class="flex gap-x-4 bg-light p-4">
     <!-- The formatting here is intentional to preserve the formatting of `code` -->
-    <pre class="flex-1 overflow-x-auto language-{language}"><code
-        bind:this={element}
+    <pre
+      class="flex-1 overflow-x-auto"
+      use:highlightCode><code
         class="language-{language}"
         data-dependencies={dependencies.join(',')}>{code}</code
```
mcous, line 109: `data-dependencies` is both prism-specific, and handled automatically by `prism`'s autoloader. I think we can safely remove it

mcous (commented):
LGTM

packages/core/src/lib/code-snippet.svelte
ehhong: good point. removed. will update usages in app to remove this prop as well

packages/core/src/lib/highlight-code/prism.ts
ehhong: updated!

ehhong (approved):
approving on behalf of @mcous

packages/core/src/lib/modal.svelte
```diff
@@ -74,7 +74,9 @@ $: if (typeof document !== 'undefined') {
 }
 
 onDestroy(() => {
-  document.body.classList.remove('overflow-hidden');
+  if (typeof document !== 'undefined') {
+    document.body.classList.remove('overflow-hidden');
+  }
```
ehhong, line 79: fly by fix -- was breaking the playground

ehhong: callout by @micheal-parks to use the return value of `onMount` instead, because `onDestroy` runs on the server

PR #611, opened 2024-12-17
[APP-4532] - Replace `<SearchableSelect>` with `svelte-select`

micheal-parks:
### Overview

This PR replaces our `<SearchableSelect>` component with the `svelte-select` library. I did plenty of restructuring here to try to ameliorate the UX confusion and architectural problems that have arisen out of this component.
### Changelog

* Replace `<SearchableSelect>` with `svelte-select`
* Rename `<SearchableSelect>` as `<AutocompleteInput>`. The original name is confusing and led to the problematic `exclusive` prop.
* Remove `exclusive` prop. This added unnecessary complexity - areas of app that use this should just use a `<Select>` component.
* Split out `multiple` prop into `<Multiselect>` component. This prop was overloading the behavior of this component.
* ~~replace the `cx` prop in favor of `class`. I would like us to move toward this pattern for all prime components. Adding a `cx` prop introduces an unnecessary dependency considering any part of app can import `cx` and pass the resulting string to a Prime component. This also follows the general pattern of "spread whatever you can into the underlying element whenever possible".~~ proposing this later to move this through and avoid delays based on discussion

### Next steps
I'll immediately bump app and fix all breaking changes. I've scanned through what's needed - it won't be too bad.
* Todos in app about adding a required field can be removed, `svelte-select` supports the `required` prop.


packages/core/src/lib/select/index.ts
```diff
@@ -1,5 +1,5 @@
 export { SortOptions, type SortOption } from './search';
-export { default as SearchableSelect } from './searchable-select.svelte';
+export { default as TypeaheadInput } from './typeahead-input.svelte';
```
mcous, line 2: **quibble (naming):** my brain is not cooperating with me when I try to read `typeahead`. I like ditching the "searchable select" nomenclature, but could we maybe name this `Combobox` instead to match ARIA?

I did a little digging in other Svelte component libraries and found that [skeleton](https://www.skeleton.dev/) ships a separate `Autocomplete` component that you can compose into a combobox. So if `Comboxbox` doesn't tickle you, I also like `AutocompleteInput` as a name

packages/core/tailwind.config.ts
```diff
@@ -1,5 +1,5 @@
 import type { Config } from 'tailwindcss';
-import { theme } from './theme';
+import { theme } from './src/theme';
```
mcous, line 2: **issue:** `theme.ts` is references in `package.json::exports` - if we rename it, we need to fix that field, too

packages/core/src/lib/select/typeahead-input.svelte
```diff
@@ -0,0 +1,149 @@
...
+let className = '';
+export { classNames as class };
+
+$: normalizedOptions = options.map((option) => {
+  if (typeof option === 'string') {
+    return { value: option, label: option };
+  }
+
+  return option;
+});
```
mcous, line 58: **thought (performance):** This might be less of a big deal once Svelte 5 is in play, but this list of objects situation has given us a lot of grief in terms of unnecessary DOM operations from this component. I wonder if there's a path here - maybe in a future PR - where we lean more heavily into slots / snippets for stuff like option metadata rather than try to jam it all into props. Seems a little closer to how `svelte-select` itself works

mcous (commented):
Quick initial thoughts. Very excited for this

packages/core/package.json
```diff
@@ -86,6 +86,8 @@
     "publint": "^0.2.6",
     "svelte": "^4.2.8",
     "svelte-check": "^3.6.2",
+    "svelte-select": "^5.8.3",
+    "svelte-tiny-virtual-list": "^2.1.2",
```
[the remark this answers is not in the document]
micheal-parks: Experiment, ended up being to complex for this initial PR, will remove and re-add later when I get a moment!

packages/core/tailwind.config.ts
micheal-parks: Ah good catch. For reference: did this to avoid any issues involving vite's filesystem access scoping

PR #612, opened 2024-12-19
Remove Storybook

micheal-parks:
### Overview

`prime.viam.com` is dead, and the migration to astro is complete. We are free.


ethanlookpotts (approved):
Woohoo! This is great news!

Not for this PR, just thinking out loud: I'm wondering if we should set up a redirect from prime.viam.com to design.viam.com? Thoughts?

Also, just to double check - did you run `pnpm dedupe` to clean up the lockfile?

micheal-parks:
@ethanlookpotts Forgot, thanks for the reminder!

As far as redirects go I believe @mrloureed requested but Steve didn't want to? Not sure if I got the full story there.

PR #614, opened 2025-01-03
DATA-3601 Add code colors

katiepeters:
Add colors to our theme for syntax highlighting so that we have something to use in situations where we are not using codemirror but want it to be visually consistent.

### DIFF
--- a/packages/core/theme.ts
+++ b/packages/core/theme.ts
@@ -20,6 +20,14 @@ export const theme = {
       'subtle-2': '#7a7c80',
       disabled: '#9c9ca4',
       link: '#0066cc',
+
+      // Code colors
+      'code-boolean': '#201293',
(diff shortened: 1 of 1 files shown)


packages/core/theme.ts
```diff
@@ -20,6 +20,14 @@ export const theme = {
       'subtle-2': '#7a7c80',
       disabled: '#9c9ca4',
       link: '#0066cc',
+
+      // Code colors
+      'code-boolean': '#201293',
+      'code-line-number': '#4b758d',
+      'code-null': '#708',
+      'code-number': '#2f6447',
+      'code-property-keys': '#000000',
```
mrloureed, line 29: complete nit: this line could be
```
      'code-property-keys': '#000',
```

mrloureed (approved):
LGTM!

PR #616, opened 2025-01-13
APP-6535: only fire clickOutside if click starts and ends outside

mcous:
## Overview

In browsers, if a click `mousedown` starts in one element and `mouseup` ends in another element, the `click` event is fired on the lowest common ancestor of the two elements.

> If the button is pressed on one element and the pointer is moved outside the element before the button is released, the event is fired on the most specific ancestor element that contained both elements

[MDN - click event](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event)

In most cases, this ancestor is going to trigger our `clickOutside` logic, resulting in frustrating user experiences, especially when selecting text.

## Change log

Swap `click` listener for `mousedown` and `mouseup` listeners using the same "is target outside?" logic. If and only if both events are outside the target element, the handler is fired

### DIFF
--- a/packages/core/src/lib/__tests__/click-outside.spec.ts
+++ b/packages/core/src/lib/__tests__/click-outside.spec.ts
@@ -8,9 +8,10 @@ describe('use:clickOutside', () => {
   const onClickOutside = vi.fn();
 
   it('should trigger a callback only when clicked outside', async () => {
+    const user = userEvent.setup();
+
     render(Subject, { onClickOutside });
 
-    const user = userEvent.setup();
     const subject = screen.getByTestId('subject');
     const insideButton = screen.getByTestId('inside');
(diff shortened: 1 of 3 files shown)


michaellee1019 (approved):

packages/core/src/lib/__tests__/click-outside.spec.ts
```diff
@@ -39,4 +40,21 @@ describe('use:clickOutside', () => {
 
     expect(onClickOutside).not.toHaveBeenCalled();
   });
+
+  it('should not trigger if click starts inside element and moves out', async () => {
```
mcous, line 44: TODO: add corresponding test - start outside, end inside

mcous (commented):
Missing test

mrloureed (approved):
LGTM!

PR #618, opened 2025-01-14
Feat: export icon path data for direct usage in non-Svelte UIs

mcous:
After chatting in dev jam and reviewing a draft PR, it looks like exposing this path data directly will be useful for markdown-based UIs in the app


michaellee1019 (approved):
TY!

mrloureed (approved):
LGTM!

PR #623, opened 2025-01-28
DATA-3600 Add indeterminate checkbox

katiepeters:
New **indeterminate** checkbox ⭐ 

Manually added path value. mdi package does not have the right design.


mrloureed (commented):
Don't forget to bump the version in package.json since we're still using npm

mrloureed (approved):
LGTM

PR #627, opened 2025-01-30
DATA-3600 Tweak checkbox path

katiepeters:
The indeterminate checkbox was ever so slightly not a square. So I had Nat fix it up


mrloureed (approved):
LGTM

PR #629, opened 2025-02-04
Add modal variant for fitting size to content

zaporter-work:
Otherwise there is no way to specify a custom max size for the content. It will always have a max h and w.


micheal-parks (approved):
Should we have an identical PR into app?

zaporter-work:
Merging for now. Can revisit. @mrloureed is busy right now and this is blocking

PR #632, opened 2025-07-07
Add google maps to MapLibre

DTCurrie:
Screenshot 2025-07-07 at 4 26 20 PM
Screenshot 2025-07-07 at 4 30 35 PM

Backwards compatible and safe, requires an API key to be passed which let's us keep it in `.env` files.

### DIFF
--- a/packages/blocks/src/lib/maplibre/__tests__/style.spec.ts
+++ b/packages/blocks/src/lib/maplibre/__tests__/style.spec.ts
@@ -1,0 +1,36 @@
+import { describe, it, expect } from 'vitest';
+import { getStyleSpecification } from '../style';
+import { MapProviders } from '../types';
+
+describe('MapLibre Provider', () => {
+  describe('getStyle', () => {
+    it('should return OpenStreetMap style for open-streets provider', () => {
+      const style = getStyleSpecification('open-street');
+
+      expect(style.version).toBe(8);
+      expect(style.sources).toHaveProperty('osm');
+      expect(style.sources).toHaveProperty('satellite');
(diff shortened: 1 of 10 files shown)


packages/blocks/src/lib/maplibre/__tests__/style.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { describe, it, expect, vi } from 'vitest';
+
+// Mock maplibre-gl module to avoid URL.createObjectURL issues
```
micheal-parks, line 3: ah, cool

packages/blocks/src/lib/navigation-map/index.svelte
```diff
@@ -60,6 +65,12 @@ export let tabs: NavigationTabType[] = [
 /** The pose (Lng,Lat) and rotation of a base. */
 export let baseGeoPose: GeoPose | undefined = undefined;
 
+/** The map provider to use. */
+export let provider: MapProvider = MapProviders.openStreet;
```
micheal-parks, line 69: should we call this mapProvider? just to avoid confusion with context providers

packages/blocks/src/lib/maplibre/index.svelte
```diff
@@ -6,7 +6,13 @@
   Children will mount once the map is fully loaded.
 
   ```svelte
-    <MapLibre>
+    <MapLibre provider="open-street">
```
DTCurrie, line 9:
```suggestion
    <MapLibre mapProvider="open-street">
```

PR #636, opened 2025-11-19
APP-14390 Add bookmark-outline icon

n0nick:


katiepeters (changes requested):
We don't use this repo anymore. Prime was moved into the app repo

n0nick:
> We don't use this repo anymore. Prime was moved into the app repo

doh

OK, looking into that. But should this repo be archived?... It's public!

